PR #168, opened 2017-08-10
Update DuckDNS plugin to also update IPv6

danrspencer:
This doesn't work properly yet, I was struggling to get `dig` added to the as a package in the Dockerfile. It needs to come from a custom repo which looks like is platform dependent (https://pkgs.alpinelinux.org/contents?file=dig).

Pull requesting now as I don't know when I'll get chance to finish figuring this out and it might be a simple fix for someone a bit more familiar with the Docker.

### DIFF
--- a/duckdns/Dockerfile
+++ b/duckdns/Dockerfile
@@ -4,7 +4,7 @@ FROM %%BASE_IMAGE%%
 ENV LANG C.UTF-8
 
 # Setup base
-RUN apk add --no-cache jq curl
+RUN apk add --no-cache jq curl dig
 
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @danrspencer, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#168).

Thanks!

pvizeli:
You Need install `bind-tools` like you suggest with your link

pvizeli:
ping

PR #173, opened 2017-08-21
Rename all keys & consts from repositorie to repository

thibmaek:
Seemed odd not to name it correctly in EN-US format

### DIFF
--- a/git_pull/config.json
+++ b/git_pull/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Git pull",
-  "version": "0.3",
+  "version": "0.3.1",
   "slug": "git_pull",
   "description": "Simple git pull to update the local configuration",
   "url": "https://home-assistant.io/addons/git_pull/",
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @thibmaek, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#173).

Thanks!

thibmaek:
Aside from this seems this add-on is also totally broken, /config has a .git dir but this add-on tries to clone it instead of pulling:

image

```
starting version 3.2.2
[Info] cleanup config folder and clone from repositorie
[Error] can't clone https://github.com/thibmaek/homeassistant into /config
```

Reinstalling add-on doesn't fix it.

gbrayut:
It doesn't work in hassio because the /config folder isn't mapped into the add-on container. The config.json  file needs the [map key](https://home-assistant.io/developers/hassio/addon_config/) with rw to the config folder:
```  
  "hassio_api": true,
  "map": ["config:rw"],
  "options": {
```

PR #179, opened 2017-09-24
New build system

pvizeli:


cruxlin:
I need to replace "asoundrc" to add more HW settings in order to make my USB speaker work with Snips audio server. I had tried to clone these files, replace asoundrc, modify some descriptions to make it a local addon. Although I can install it successfully, the asoundrc file is still the old one.  Do you know how to do the replacement with the pre-built image?  Or may you consider to open more options for speaker settings like rate and format in asoundrc? 

The following bold texts are what I want to add in the asoundrc:
pcm.speaker {
  type plug
  slave {
    pcm "hw:%%SPEAKER%%"
    **rate 48000  
    format "S16_LE"
    channels 2**
  }
}

PR #181, opened 2017-09-26
DuckDNS: Also generate a certificate with Let's Encrypt (0 config)

balloob:
This is based on this blog post: https://www.splitbrain.org/blog/2017-08/10-homeassistant_duckdns_letsencrypt

I was trying to be fancy and do the DNS challenge via the new DuckDNS component in Home Assistant. However, I realize now that this is not going to work because we won't know if we should connect to Home Assistant over SSL or not.

So then I went for attempt 2: instead of creating a new add-on, why not make it part of the DuckDNS add-on? 2 birds, 1 stone.

Right now it will only generate a certificate for the first domain in the DuckDNS list.

I left my previous attempts in the history of this branch.

What do you think of this approach?

(btw this is a proof of concept, I have not tested this yet)


duckdns/run.sh
```diff
@@ -1,12 +1,42 @@
...
+LE_DOMAINS=$(jq --raw-output '.domains[]' $CONFIG_PATH)
+
+# DuckDNS
 TOKEN=$(jq --raw-output '.token' $CONFIG_PATH)
 DOMAINS=$(jq --raw-output '.domains | join(",")' $CONFIG_PATH)
 WAIT_TIME=$(jq --raw-output '.seconds' $CONFIG_PATH)
 
+# Register/generate certificate if terms accepted
+if [ "$LE_TERMS" == "true" ]; then
+  # Only use first domain for Let's Encrypt
```
balloob: Stale comment

duckdns/run.sh
```diff
@@ -1,12 +1,42 @@
...
+  # Create empty dehydrated config file so that this dir will be used for storage
+  touch "$WORK_DIR/config"
+
+  # Generate new certs
+  if [ ! -d "$CERT_DIR/live" ]; then
+    dehydrated --register --accept-terms --config "$WORK_DIR/config"
+  fi
+
+  dehydrated --cron --hook ./hooks.sh --challenge dns-01 "${DOMAIN_ARGS[@]}" --out "$CERT_DIR"  --config "$WORK_DIR/config"
+  crond
```
balloob: We did not setup a cronjob did we? I wonder if we can built it into the running duckdns part. Just have a test for every 12 hours (it will only renew if only certain number of days left)

PR #185, opened 2017-10-06
Update snips to latest version

andreasjacobsen93:


pvizeli:
Thanks. I had made a small update that was pendent

PR #197, opened 2017-10-27
Use jq as a tool to validate all JSON files

alexbarcelo:
In this PR I have added a check to validate JSON. It seems nice to have it, just in case. The behaviour of `xargs` for doing this test seemed correct in my tests. I ended up using it because `jsonlint-php` doesn't behave well with multiple files as parameters, and it was the easiest package to add given that it is in the repositories.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,11 @@ addons:
     - debian-sid    # Grab shellcheck from the Debian repo (o_O)
     packages:
     - shellcheck
+    - jq
 
 script:
  - bash -c 'shopt -s globstar; shellcheck **/*.sh'
+ - bash -c 'shopt -s globstar ; jq . **/*.json'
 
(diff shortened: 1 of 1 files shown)


alexbarcelo:
Travis failed... during `apt` installation... which I don't understand. Maybe the php-cli requirements. I could switch to [another jsonlint](http://zaach.github.com/jsonlint/) if that's the problem, but I find the error strange.

pvizeli:
Good idea, thanks

alexbarcelo:
Updated the title and push-forced some changes. `npm` was not being installed correctly, so in the end I used the `jq` utility which seem to act correctly as a JSON checker.

Now it seems to work as intended.

pvizeli:
Thanks a lot. Very cool :cookie:

PR #198, opened 2017-11-08
Use reflog to calculate diff and allow 'git://' repo addresses

jangeador:
Some people prefer to use ssh addressing for their repo instead of https addressing. Changing the config restriction from `url` to `str` would allow them to do this.

For the reflog change to calculate changed files please refer to this answer from https://stackoverflow.com/questions/964876/head-and-orig-head-in-git

> HEAD is (direct or indirect, i.e. symbolic) reference to the current commit. It is a commit that you have checked in the working directory (unless you made some changes, or equivalent), and it is a commit on top of which "git commit" would make a new one. Usually HEAD is symbolic reference to some other named branch; this branch is currently checked out branch, or current branch. HEAD can also point directly to a commit; this state is called "detached HEAD", and can be understood as being on unnamed, anonymous branch.
> 
> And @ alone is a shortcut for HEAD, since Git 1.8.5
> 
> ORIG_HEAD is previous state of HEAD, set by commands that have possibly dangerous behavior, to be easy to revert them. It is less useful now that Git has reflog: HEAD@{1} is roughly equivalent to ORIG_HEAD (HEAD@{1} is always last value of HEAD, ORIG_HEAD is last value of HEAD before dangerous operation).
> 
> For more information read git(1) manpage, Git User's Manual, the Git Community Book and Git Glossary

### DIFF
--- a/git_pull/config.json
+++ b/git_pull/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Git pull",
-  "version": "2.0",
+  "version": "2.1",
   "slug": "git_pull",
   "description": "Simple git pull to update the local configuration",
   "url": "https://home-assistant.io/addons/git_pull/",
@@ -21,7 +21,7 @@
   "schema": {
     "deployment_key": ["str"],
     "deployment_key_protocol": "match(rsa|dsa|ecdsa|ed25519|rsa)",
-    "repository": "match((?:git|ssh|https?|git@[-\w.]+):(\/\/)?(.*?)(\.git)(\/?|\#[-\d\w._]+?))",
+    "repository": "match((?:git|ssh|https?|git@[-\\w.]+):(\/\/)?(.*?)(\\.git)(\/?|\\#[-\\d\\w._]+?))",
     "auto_restart": "bool",
     "repeat": {
       "active": "bool",
--- a/git_pull/config.json
+++ b/git_pull/config.json
@@ -21,7 +21,7 @@
   "schema": {
     "deployment_key": ["str"],
     "deployment_key_protocol": "match(rsa|dsa|ecdsa|ed25519|rsa)",
-    "repository": "url",
+    "repository": "str",
     "auto_restart": "bool",
     "repeat": {
       "active": "bool",
--- a/git_pull/config.json
+++ b/git_pull/config.json
@@ -21,7 +21,7 @@
   "schema": {
     "deployment_key": ["str"],
     "deployment_key_protocol": "match(rsa|dsa|ecdsa|ed25519|rsa)",
-    "repository": "str",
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @jangeador, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#198).

Thanks!

git_pull/config.json
```diff
@@ -21,7 +21,7 @@
   "schema": {
     "deployment_key": ["str"],
     "deployment_key_protocol": "match(rsa|dsa|ecdsa|ed25519|rsa)",
-    "repository": "url",
+    "repository": "str",
```
pvizeli: I think `match(^[\w]+://[\w\.\-\/]+$)` is better?

git_pull/run.sh
```diff
@@ -43,7 +43,7 @@ while true; do
 
     # Enable autorestart of homeassistant
     if [ "$AUTO_RESTART" == "true" ]; then
-        changed_files="$(git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD)"
+        changed_files="$(git diff-tree -r --name-only --no-commit-id HEAD@{1} HEAD)"
```
pvizeli: You can fix the lint with `# shellcheck disable=` before your statement

git_pull/config.json
jangeador: If we want to get more specific, this will probably work better: `match((?:git|ssh|https?|git@[-\w.]+):(\/\/)?(.*?)(\.git)(\/?|\#[-\d\w._]+?))`

borrowed from https://github.com/jonschlinkert/is-git-url/blob/master/index.js

git_pull/run.sh
jangeador: I don't understand how to do this. can you show me where exactly the statement goes.

pvizeli: use `'HEAD@{1}'`

denniss17:
Nice work! Cant wait to see this merged. Currently it is impossible to clone a private repo, however there is a workaround

Workaround: setup the git repo in `/config` (using `git init` and `git remote add origin <url>`). Then, do `cp /config/.git/HEAD /config/.git/ORIG_HEAD`. After this, the plugin works, no matter what url you fill in in the config. This is because the `git pull` command uses the already configured origin remote.

jangeador:
Thanks for the fixes and for the merge.

jonkristian:
`"repository": "match((?:git|ssh|https?|git@[-\\w.]+):(\/\/)?(.*?)(\\.git)(\/?|\\#[-\\d\\w._]+?))",`
I think we need to improve the regexp here. Because some of us aren't using github, for instance mine is `gitlab@git.domain.com:group/repo.git`

I'd suggest a regexp which can match any word before @.

pvizeli:
you can make a PR, that will be update without create a new version

jonkristian:
Awesome, thanks!

PR #204, opened 2017-11-21
Disable binlogging by default

brainyron:
The default configuration in mariadb will write binlogs without limit or rotation. The binlogging will fill up the disk on small embedded devices and worse: the constant writes are likely to severely shorten the usable lifespan of a SD card when running this on a raspberry pi.

For this application, you're unlikely to ever need the binlogging, since people running hassio aren't likely doing replay log backups or replication.  The innodb transaction logs are separate and not affected by this conversation so the database is still replay-save in the event of a crash/power-loss event. As such, I suggest binlogs should be disabled by default. If we feel that there may be users who want this, we could add a configuration option in to support it, but I would also suggest that such a use case isn't really the target of this addon.

### DIFF
--- a/mariadb/config.json
+++ b/mariadb/config.json
@@ -1,6 +1,6 @@
 {
   "name": "MariaDB",
-  "version": "0.1",
+  "version": "0.2",
   "slug": "mariadb",
   "description": "An SQL database server",
   "url": "https://home-assistant.io/addons/mariadb/",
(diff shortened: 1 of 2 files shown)


brainyron:
Forgive the newbie question -- is this waiting on me to do something before it is merged?

pvizeli:
No, build run now and after that I can merge it

PR #206, opened 2017-12-04
TellStick - Added stdin, clarified desc. and config

endor-force:
- Added Logo
- Added stdin support for hassio addon_stdin calls to tdtool
- Added clarification in description 
- Added second device in example configuration to better visualise formatting.

- [x] Update add-on documentation ( i will do this.)


endor-force:
I do not have a good solution to this word splitting thing..
The bash has been confirmed to work just fine..

tyldum:
@endor-force Tried  --${input} ?

tyldum:
Ah, it wants you to do 
`if ! msg="$(tdtool \"--${input})\""; then`
In case $input contains spaces

endor-force:
Yay.. should be "$input" to preserve whitespaces..  \" would escape the double qoutes in to the command.

endor-force:
Documentation updated:
https://github.com/home-assistant/home-assistant.github.io/pull/4137

pvizeli:
I build it with new base images (alpine 3.7)

PR #207, opened 2017-12-05
Fix bash variable input

endor-force:
This will cause error in validation however it actually works..

I have tried with different versions to this without success in the addon:
`if ! msg="$(tdtool --"$input")"; then`
`if ! msg="$(tdtool --"${input}")"; then`

It passes the data sent via addon_stdio but something is parsed wrong when it is executing the command and tdtool exits with invalid input error.

The only one i got working in real life is these two:
`if ! msg="$(tdtool --$input)"; then`
`if ! msg="$(tdtool --${input})"; then`

### DIFF
--- a/tellstick/config.json
+++ b/tellstick/config.json
@@ -1,6 +1,6 @@
 {
   "name": "TellStick",
-  "version": "0.2",
+  "version": "0.3",
   "slug": "tellstick",
   "description": "TellStick and TellStick Duo service.",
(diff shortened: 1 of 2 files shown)


pvizeli:
Use this: https://github.com/home-assistant/hassio-addons/blob/master/rpc_shutdown/run.sh#L11
Data from Home-Assistant comes as json formated.

PR #208, opened 2017-12-06
Stdin does not work with quoting on $input

endor-force:
There is either something wrong with the handling of quotes when receiving input from json -> jq and parsing it to a quoted variable to bash.

Proof of concept to replicate below.

Created addon for test/play/verification: https://github.com/endor-force/stdin-addon
Service call: `{"addon":"local_stdintest","input":"l /"}`

```
#!/bin/bash
set -e

# Listen for input to tdtool
echo "[Info] Starting event listener"
while read -r input; do
    # removing JSON stuff
    input="$(echo "$input" | jq --join-output '.')"
    echo "[Info] Read alias: $input"
	
    if msg="$(ls -"$input")"; then
        echo "[Info] TellStick Command success -> $msg"
    fi
done
```

Sending data without spaces works.
But if you send `l /` (to run `ls -l /`) it will fail.

### DIFF
--- a/tellstick/config.json
+++ b/tellstick/config.json
@@ -1,6 +1,6 @@
 {
   "name": "TellStick",
-  "version": "0.3",
+  "version": "0.4",
   "slug": "tellstick",
   "description": "TellStick and TellStick Duo service.",
   "url": "https://home-assistant.io/addons/tellstick/",
--- a/tellstick/run.sh
+++ b/tellstick/run.sh
@@ -72,9 +72,9 @@ while read -r input; do
     input="$(echo "$input" | jq --raw-output '.')"
     echo "[Info] Read alias: $input"
     
-    input=("tdtool --$input")
-    input_arr=()
-    for word in ${input[@]}; do
+    input=("--$input")
+    input_arr=("tdtool")
+    for word in $input; do
         input_arr+=("$word")
     done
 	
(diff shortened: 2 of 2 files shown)


pvizeli:
that will work?

endor-force:
no.. it did not

pvizeli:
now I know what do you want. You need parse it into array by spliting with ' ' and parse the array to args like:

pvizeli:
Maybe it will be easy if you modify the input as json like:
```
{
  "type": "learn",
  "id": "id"
}
```
So you can parse the input data like this schema. Maybe you can use other types for different functions.

endor-force:
Tried this.. Same result.. fail.
```
input_arr=()
for word in $input; do
  input_arr+=("$word")
done
	
echo "[Info] Read array: "${input[@]}""
	
if msg="$(ls -"${input[@]}")"; then
  echo "[Info] Command success -> $msg"
fi
```

endor-force:
Verified theory with tellstick add-on locally and it works.

endor-force:
Forget it.. with * it is back to square one .. it does not work in reality.

endor-force:
Stupid

endor-force:
Thank you! 🥇 👍

endor-force:
Works like a charm. 
Updated documentation: https://github.com/home-assistant/home-assistant.github.io/pull/4156

PR #213, opened 2017-12-10
Added sector alarm.

endor-force:
I'm not sure if this is a suitable core addon for hass?
Maybe it should be a community add-on?


frenck:
I'm going be honest here, why not add it to Home Assistant itself?

Secondly, there is a drawback to this method. If a user uninstalls the add-on, it is not actually removed?

Third concern, the risk of tons of add-ons being developed installing components, instead of being added to the main code base.

endor-force:
Points are perfectly valid and I agree.

I will take it back to screen the code and make a PR for adding it to Home Assistant itself.

PR #237, opened 2018-01-21
MQTT Hue bridge 

lazytesting:
This addon wraps Hue MQTT Bridge https://github.com/dale3h/hue-mqtt-bridge


pvizeli:
I'm fine with it, but I will first implement to read the config from home-assistant to allow more user friendly add-ons and after that we can add this :+1:  I think 1 week

nicolas-fricke:
Any update on this? It would be really great to be able to integrate this service as a simple Hass.io add-on :)

PR #239, opened 2018-01-29
Snips update to 0.51.14

tschmidty69:
This updates the add on to use the latest version of snips and adds a default assistant that can trigger the built in intents provided by conversation and shopping_list.

Example utterances:
 - Add beer to my shopping list
 - I've added beer to your shopping list

- Turn the kitchen light off
- Turned off the kitchen light

With a USB mic and speakers hooked up to a pi 3, TTS works for response and the snips_say services.

This will provide a good example for users to see what snips can do.

If you want to update the official docker images feel free to pull/push the ones I have and update the config.json accordingly.


snips/config.json
```diff
@@ -36,5 +36,5 @@
     "speaker": "str",
     "assistant": "str"
   },
-  "image": "homeassistant/{arch}-addon-snips"
+  "image": "tschmidty/{arch}-addon-snips"
```
balloob: We should always publish all images under our account.

snips/config.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Snips.AI",
-  "version": "0.6.5-p1",
+  "version": "0.51.14",
```
balloob: This is the version number of the add-on. We should just bump it to 0.7

snips/run.sh
```diff
@@ -36,24 +53,30 @@ if [ "$MQTT_BRIDGE" == "true" ]; then
...
 # check if a new assistant file exists
 if [ -f "/share/$ASSISTANT" ]; then
-    echo "[Info] Install/Update snips assistant"
-    unzip -o -u "/share/$ASSISTANT" -d "$SNIPS_CONFIG"
+    echo "[INFO] Install/Update snips assistant"
+    unzip -o -u "/share/$ASSISTANT" -d /usr/share/snips
 fi
-ln -s "$SNIPS_CONFIG/assistant/" /usr/share/snips/assistant
+
+sleep 2
```
balloob: Why do we need to sleep 2?

snips/run.sh
```diff
@@ -36,24 +53,30 @@ if [ "$MQTT_BRIDGE" == "true" ]; then
           echo "password $PASSWORD"
       } >> /etc/mosquitto.conf
     fi
-    
+
     {
-        echo "topic # OUT"
-        echo "topic # IN hermes/ hermes/"
+        echo "topic hermes/intent/# out"
```
balloob: Why do we need to be so specific? Can't we just bridge everything under `hermes/` ?

snips/config.json
tschmidty69: 0.6.5 was also the version number of snips at one point. But that makes sense to keep our own so we can iterate between snips releases if we wanted.

snips/run.sh
tschmidty69: Sleep 2 is needed because on my rasp pi 3, the mosquitto broker wasn't starting fast enough for snips-watch to be able to connect to it. There is no retry method in that app but it is really useful to output the intent recognition to the logs. The logging does not show anything particularly useful otherwise.

snips/run.sh
tschmidty69: Snips uses MQTT to actually transmit the audio PCM frames, so it seems inefficient to me to clutter the hassio mqtt broker with constant binary audio frames which are useless to any other app.

snips/run.sh
tschmidty69: To clarify, without snips-watch running the only thing in the logs is when the asr stops and starts listening, so you can infer a hotword was detected but can't see anything else like what text was parsed or what intent was detected. I vaguely thought about some kind of 'wait until mosquitto starts' method but that would be fairly painful in bash and seemed like overkill

snips/config.json
tschmidty69: I don't have access to that account, but if it allows shared access I'd be happy to publish it. My docker id is 'tschmidty'

balloob: Pascal publishes all the add-ons based on the source code.

snips/config.json
balloob: We should not couple the version number. We should be able to update the add-on without updating Snips.

snips/run.sh
balloob: If it's only needed for snips-watch, let's add the delay only to the background command: https://unix.stackexchange.com/a/102961

snips/config.json
```diff
@@ -14,14 +14,14 @@
   "devices": ["/dev/snd:/dev/snd:rwm"],
```
pvizeli: We should use now `"audio": true`

tschmidty69:
@pvizeli The docker container should be pulled from my docker repo and pushed up rather than rebuilding if possible since I included my custom assistant in the container. 

Would you prefer I use something like "ADD https://github.com/tschmidty69/hassio-snips/releases/download/0.1-pre1/assistant.zip /share/assistant-default.zip" in the Dockerfile which would work also so you could do a clean build if needed?

pvizeli:
Yes, It's okay to add a default assistant with basic features like on/off :+1:

snips/Dockerfile
```diff
@@ -1,11 +1,14 @@
-ARG BUILD_FROM
-FROM $BUILD_FROM
+#ARG BUILD_FROM
+#FROM $BUILD_FROM
+FROM snipsdocker/platform:x86-0.51.14
```
pvizeli: Revert this change

snips/Dockerfile
```diff
@@ -1,11 +1,14 @@
...
 
 # Copy data
 COPY run.sh /
 COPY mosquitto.conf /etc/
 COPY asoundrc /root/.asoundrc
 
+ADD https://github.com/tschmidty69/hassio-snips/releases/download/0.1-pre1/assistant.zip /assistant-default.zip
```
pvizeli: Use a build argument for the version with `build.json`: https://home-assistant.io/developers/hassio/addon_config/#add-on-extended-build

tschmidty69: Got it. Can be changed to point to a release in this repo if you want to upload it.

pvizeli: You can upload the zip into this folder as a file.

pvizeli:
You can upload the assistant.zip into this folder. I have no access. After that it should be okay

tschmidty69:
[assistant.zip](https://github.com/home-assistant/hassio-addons/files/1675012/assistant.zip)

Ah, didn't realize you could do that yet.

pvizeli:
Good work :+1:

PR #277, opened 2018-04-07
Add Lets Encrypt + CloudFlare addon

kiall:
Add a Lets Encrypt + CloudFlare addon.

This addon is heavily based on the DuckDNS addon (It's essentially a copy+paste!), and will issue Lets Encrypt TLS certificates using a DNS-01 challenge setup with CloudFlare's DNS service.

### DIFF
--- a/letsencrypt_cloudflare/run.sh
+++ b/letsencrypt_cloudflare/run.sh
@@ -5,16 +5,16 @@ CERT_DIR=/data/letsencrypt
 WORK_DIR=/data/workdir
 CONFIG_PATH=/data/options.json
 
-# Let's encrypt
+# Let's Encrypt
 LE_TERMS=$(jq --raw-output '.lets_encrypt.accept_terms' $CONFIG_PATH)
 LE_DOMAINS=$(jq --raw-output '.domains[]' $CONFIG_PATH)
 LE_UPDATE="0"
 
 # CloudFlare
-export CF_EMAIL=$(jq --raw-output '.cloudflare.email' $CONFIG_PATH)
-export CF_KEY=$(jq --raw-output '.cloudflare.key' $CONFIG_PATH)
+CF_EMAIL=$(jq --raw-output '.cloudflare.email' $CONFIG_PATH)
+CF_KEY=$(jq --raw-output '.cloudflare.key' $CONFIG_PATH)
 
-DOMAINS=$(jq --raw-output '.domains | join(",")' $CONFIG_PATH)
+# Common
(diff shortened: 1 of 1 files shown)


letsencrypt_cloudflare/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# Add env
+ENV LANG C.UTF-8
+
+# Setup base
+ARG DEHYDRATED_VERSION
+ARG CLOUDFLARE_HOOK_SHA1
+
+# Setup Base System
+RUN apk add --no-cache git jq curl libressl python3 \
+  && ln -s /usr/bin/python3 /usr/bin/python
```
pvizeli: That is needed?

Apocrathia:
@balloob Can we meet in the middle and add the DNS verification support to the existing Let's Encrypt addon? There are a lot of us who are using CloudFlare for DNS.

@pvizeli You're authored the majority of the Let's Encrypt addon, and I see that you've been using certbot for it. Could we add CloudFlare support to it? It looks like the `config.json` just needs a variable for the verification method and the optional fields required by DNS verification. [The CloudFlare plugin is built into certbot now](https://github.com/certbot/certbot/tree/master/certbot-dns-cloudflare).

Apocrathia:
@pvizeli [I have made a couple of minor tweaks to the existing Let's Encrypt addon.](https://github.com/Apocrathia/hassio-addons/tree/certbot-dns-support/letsencrypt) I will need to flesh out the functionality, but if this looks okay, we may be able to kill two birds with one stone by adding DNS-01 challenge support, as well as preventing us from having another addon in the main repository.

balloob:
I just suggested in #312 to move Let's Encrypt add-on out of the official repository. If there is an add-on that would do dynamic DNS + Let's Encrypt and support multiple providers, that would be something I would like for the official repo.

Just because something is not accepted for the official repository, doesn't mean that you can't use it or share it with others. We've made it very easy to share add-ons: https://developers.home-assistant.io/docs/en/hassio_addon_repository.html

Apocrathia:
I would honestly say that dynamic DNS and SSL certificate generation are two separate functions, and having the DuckDNS addon do both is even more confusing. In addition, I would argue that DNS-o-Matic provides far more functionality than just DuckDNS, if dynamic DNS is the end goal. I agree with your comment in #312 that we should have some sort of an "advanced" addon repo.

balloob:
Combining Let's Encrypt with a dynamic DNS provider using the DNS challenge makes generating certificates possible without requiring users to open ports on their router, making the process significant easier and more secure.

PR #278, opened 2018-04-09
Add support for longer domain names in nginx_proxy

### LINKED ISSUE #243: Nginx: Increase server_names_hash_bucket_size
When using "long" dns server name from for example duckdns nginx won't start and prints an error message:

`nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 32`

Setting `server_names_hash_bucket_size 32;` in the http part of nginx.conf should fix the issue.

Setting domain to `*.duckdns.org` works in the meantime.

`{
  "domain": "*.duckdns.org",
  "certfile": "fullchain.pem",
  "keyfile": "privkey.pem"
}`

danielwelch:
Fixes #243 

Currently, `server_names_hash_bucket_size` is set to `32`, which is fine in most cases but prevents longer domain names from working. In my case, I ran into the need to increase the size when creating a  `server_name` like the following: `subdomain.mydomain.duckdns.org`

### DIFF
--- a/nginx_proxy/nginx.conf
+++ b/nginx_proxy/nginx.conf
@@ -10,6 +10,7 @@ http {
     map $http_upgrade $connection_upgrade {
         default upgrade;
         ''      close;
+    server_names_hash_bucket_size 64;
     }
 
     server {
--- a/nginx_proxy/nginx.conf
+++ b/nginx_proxy/nginx.conf
@@ -10,8 +10,9 @@ http {
     map $http_upgrade $connection_upgrade {
         default upgrade;
         ''      close;
-    server_names_hash_bucket_size 64;
     }
+    
+    server_names_hash_bucket_size 64;
 
     server {
         server_name _;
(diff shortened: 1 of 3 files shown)


nginx_proxy/nginx.conf
```diff
@@ -10,6 +10,7 @@ http {
     map $http_upgrade $connection_upgrade {
         default upgrade;
         ''      close;
+    server_names_hash_bucket_size 64;
```
frenck: Indenting is off.

Secondly, why not add this to the `http` level of the configuration?

danielwelch: You're correct, this *should* be in the `http` level of the configuration, not where I've put it. I accidentally put it within the map brackets, hence why the indentation is also off. My fault, tried using the Github online editor since it was such an easy fix and messed up.

pvizeli:
Please update the changelog and pump the version +0.1.

PR #285, opened 2018-05-02
Snips 0.55.2

tschmidty69:
Updated to latest snips, 0.55.2
Added customtts option which uses HA providers
Added german and french assistants


snips/run.sh
```diff
@@ -5,13 +5,40 @@ CONFIG_PATH=/data/options.json
...
+        echo "[INFO] - Using custom tts"
+        echo "provider = \"customtts\"" >> /etc/snips.toml
+        echo "customtts = { command = [\"/usr/bin/customtts.sh\", \"$PLATFORM\", \"%%OUTPUT_FILE%%\", \"$LANG\", \"%%TEXT%%\"] }" >> /etc/snips.toml
+    fi
+else
+    echo "[INFO] - Using default tts (picotts)"
+fi
+
+echo "[INFO] Copying mpg123 for $ARCH"
+mv -v /usr/bin/mpg123-$ARCH /usr/bin/mpg123
```
pvizeli: Do that inside Dockerfile

tschmidty69: Sure. The issue with a package is that the snips platform is actually based on debian so we would need to add a fair number of dependencies. I actually compiled this from source to get a static binary just to make it easier. I could probably set it up in the docker file to install it via apt as part of the build process.

snips/Dockerfile
```diff
@@ -4,8 +4,12 @@ FROM $BUILD_FROM
...
 COPY run.sh /
 COPY mosquitto.conf /etc/
-COPY assistant.zip /assistant-default.zip
+COPY customtts.sh /usr/bin
+COPY assistant_* /
 
 RUN chmod a+x /run.sh
+RUN chmod a+x /usr/bin/customtts.sh
+RUN echo nameserver 1.1.1.1 > /etc/resolv.conf; apt-get update;\
+ apt-get install -y mpg123; apt-get clean all
```
pvizeli: Please group all this RUNs into a single RUN (docker best praxis)

snips/config.json
```diff
@@ -20,7 +21,10 @@
       "user": "",
       "password": ""
     },
-    "assistant": "assistant.zip"
+    "assistant": "assistant.zip",
+    "language": "en",
+    "custom_tts": false,
+    "tts_platform": "amazon_polly"
```
pvizeli: Group this like:
```
custom_tts: {
  active: bool,
  platform: string,
}
```

snips/config.json
```diff
@@ -30,7 +34,10 @@
...
       "password": "str"
     },
-    "assistant": "str"
+    "assistant": "str",
+    "language": "str",
+    "custom_tts": "bool",
+    "tts_platform": "str"
   },
-  "image": "homeassistant/{arch}-addon-snips"
+  "image": "tschmidty/{arch}-addon-snips"
```
pvizeli: don't change this on this PR

snips/customtts.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ "$URL" = "" ]; then
+    exit 1
+fi
+
+rm /tmp/temp.mp3
+curl -s -H "x-ha-access: $API_KEY" "$URL" -o /tmp/temp.mp3
+if [ -f /tmp/temp.mp3 ]; then
+  /usr/bin/mpg123 -w "$FILE" /tmp/temp.mp3
+fi
+rm /tmp/temp.mp3
```
pvizeli: Maybe we can directly play this with need to download the file?

tschmidty69: The snips server only accepts WAV audio. We could play it but then it doesn't go through the snips dialog manager so can get odd behaviour like it hearing itself speak.

tschmidty69: To clarify, google and amazon both support PCM but it would still need to be massaged into WAV files via sox or some other program, so best to just stick with mp3 and convert it here.

pvizeli: I mean, can we play the URL on mpg123 ?

pvizeli:
Thanks for maintain and update it 👍 please set next time the correct version on build.json too. In build it was 0.53.17 and in changelog it was 0.55.1 and in the description you had use 0.55.2

snips/customtts.sh
tschmidty69: No. It wouldn't be done through snips dialog manager, so snips wouldn't know if it was playing so wouldn't know when it was done and could be listening at the same time (ie hearing itself talk) Also if you had multiroom installations of snips it wouldn't play on the correct site.

frenck:
This PR has been clearly merged untested.
Since there is an error in the configuration schema, the add-on won't even be loaded by Hass.io.

I've opened an PR containing the hotfix. #289

tschmidty69:
@TheZoker Pls start an issue and feel free to tag me. Nothing changed in the mosquitto configuration. When you do can you post your config options and the start of the log that shows mosquitto trying to start?

PR #304, opened 2018-05-13
Add server host and port options to nginx

gabrielboliveira:
This will add options to change the default home assistant host and port to the nginx addon.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.3
+- Add option for custom Homeassistant server host and port
+
 ## 1.2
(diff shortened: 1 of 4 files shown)


homeassistant:
Hi @gabrielboliveira, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#304).

Thanks!

gabrielboliveira:
Now that I realized that we are running under the Docker network, so I'm not sure if makes any sense to change the hostname.

Perhaps the port is something we can change, since it's an available configuration: https://www.home-assistant.io/components/http/

gabrielboliveira:
@pvizeli Thanks for the explanation. Let's close this, since it doesn't fit for this case.

PR #305, opened 2018-05-13
Add PR template to show CLA link

gabrielboliveira:
This is a suggestion. Add a PR template with the CLA link, so the user can pre-sign it before opening the PR.

Feel free to close it, if you guys think it's not really necessary.


pvizeli:
We use the same workflow as Home-Assistant

PR #311, opened 2018-05-30
Disable guest access for Samba by default

balloob:
By default, the Samba add-on is too open. Guest access + write access to config and other add-ons means that it's easy to take control of the system. Let's disable guest access by default (users could still enable it) so that we try to steer users to add username/password.

Also wondering if we can limit Samba by default to the local subnets?

### DIFF
--- a/samba/config.json
+++ b/samba/config.json
@@ -11,7 +11,7 @@
   "options": {
     "workgroup": "WORKGROUP",
     "name": "hassio",
-    "guest": true,
+    "guest": false,
     "map": {
       "config": true,
       "addons": true,


pvizeli:
I have no trouble with this. The main idea behind this was that it is simple and work with a click.

balloob:
The tests are failing but does not seem related to this PR.

cogneato:
@balloob @pvizeli  Changing "allow guest" from true to false never happened but the allowed hosts option was added. I submitted on the "allow guest" as a new issue in case the failing tests had anything to do it? Not sure I should have.

balloob:
I like that idea. Can we merge this one first, and then have that idea be added as a new issue that can be worked on ?

PR #330, opened 2018-06-22
Update Configurator to version 0.2.9

danielperna84:
Updating Configurator to version 0.2.9.
This includes the following changes:

- Material Icons and HASS-help now open in new tab instead of modal
- Open file by URL
- Added ENFORCE_BASEPATH option
- Cosmetic fix for scaled viewports
- Added search-function for entities
- Updated Ace Editor to 1.3.3
- Updated jQuery to 3.3.1
- Updated js-yaml to 3.12.0

### DIFF
--- a/configurator/config.json
+++ b/configurator/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Configurator",
-  "version": "0.3",
+  "version": "0.4",
   "slug": "configurator",
   "description": "Browser-based configuration file editor for Home Assistant.",
   "url": "https://home-assistant.io/addons/configurator",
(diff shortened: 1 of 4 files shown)


configurator/config.json
```diff
@@ -18,11 +18,12 @@
...
     "keyfile": "privkey.pem",
     "ssl": false,
-    "allowed_networks": ["192.168.0.0/16"],
+    "allowed_networks": ["192.168.0.0/16", "172.30.0.0/16"],
     "banned_ips": ["8.8.8.8"],
     "banlimit": 0,
     "ignore_pattern": ["__pycache__"],
-    "dirsfirst": false
+    "dirsfirst": false,
+    "enforce_basepath": false
```
poldim: What does this do?  I don't see it mentioned in the docs?

danielperna84: It's not really relevant for the hassio version of the addon because it already is inside a container. It's targeted at standalone installations where in theory the whole filesystem would be accessible.

pvizeli: In this case, we should remove it from user configuration

danielperna84: Ok. Version 0.3.0 is almost done. I'll remove it with the next upgrade.

PR #335, opened 2018-06-28
Proposed: Set guest default to false

cogneato:
Change default for guest access to "false" and inform users about setting it to true in the docs?
There is much discussion of this setting currently so I thought I'd submit.

### DIFF
--- a/samba/config.json
+++ b/samba/config.json
@@ -11,7 +11,7 @@
   "options": {
     "workgroup": "WORKGROUP",
     "name": "hassio",
-    "guest": true,
+    "guest": false,
     "map": {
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @cogneato, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#335).

Thanks!

frenck:
Isn't this a duplicate of #311 @cogneato ?

balloob:
Yeah it's a duplicate. We should get going on merging that one.

Issue #338, opened 2018-07-02
Update Mosquitto add-on


tjorim:
The official Mosquitto MQTT broker is still based on version 1.4.12.
There has been some releases addressing mostly security vulnerabilities (up to 1.4.15).
Also a 1.5 features release which "brings a host of changes to the broker, including performance improvements and [...] numerous bug fixes".

pvizeli:
Please make a PR 👍 / If alpine 3.8 is out for docker, we update it anyway but not before. But if you provide a PR i will merge it.

PR #342, opened 2018-07-03
add websocket listener config (plain and ssl)

ToSa27:
defaults to same config as the original (plain mqtt enabled, all other disabled) but adds ability to configure websockets listeners for both plain and ssl

Fix: #338


homeassistant:
Hi @ToSa27, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#342).

Thanks!

ToSa27:
@pvizeli done - please check again.
I would have celled it 1.1 because there are no breaking changes, just an added feature which should qualify for a minor increase not a major but looking at other addons it seems like you don't make that distinction in this repo, hence I did the requested change to 2.0

pvizeli:
Please update the documentation on website too

pvizeli:
Fix: #338

PR #345, opened 2018-07-04
custom_deps, let PYPI variable accept comma seperated value

mories76:
comma seperated value can be used to install multiple packages where one depend on another to be build.

(my first PR ever, so please let me know if I am doing this the right way)


mories76:
sorry, i've commited the wrong file.

PR #346, opened 2018-07-04
custom_deps, let PYPI variable accept comma seperated 

mories76:
comma seperated value can be used to install multiple packages where one depend on another to be build.

(my second PR ever, already messed up the first one, so please let me know if I am doing this the right way)

### DIFF
--- a/custom_deps/run.sh
+++ b/custom_deps/run.sh
@@ -8,7 +8,7 @@ APK=$(jq --raw-output ".apk[] // empty" $CONFIG_PATH)
 
 # create bash array from comma seperated value using Internal field seperator
 IFS=','
-read -ra packages <<< $PYPI
+read -ra packages <<< "$PYPI"
 
 # Cleanup old deps
 echo "[Info] Remove old deps"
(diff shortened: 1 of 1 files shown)


pvizeli:
What is wrong with the array?

mories76:
I want to use a custom component which uses pyaes, sleekxmpp and nefit-client, when installing them with one pip install command the dependency for nefit-client isn't met.
Installing them one by one solves the issue

pvizeli:
The config is a array, you can but all modules into this array and install it with one call

kunago:
I also need to install multiple pypi packages and even though I am properly setting the JSON array, I always get an error on install.

pvizeli:
What kind of error? We could try to fix that.

mories76:
I've had the same invalid requirement issue when using the array
"pypi": [ "pyaes", "sleekxmpp", "nefit-client" ]

I also tried
"pypi": [ "pyaes sleekxmpp nefit-client" ]

I think it has something to do with the outer quotation on the --no-dependecies parameter.
But even then when trying to install multiple packages in one pip install command from the command line in a debugging ssh session Nefit-client doesn't install because it needs the pyaes to be installed first.
Apparently pip doesn't install the packages one by one, [https://github.com/pypa/pip/issues/1386](url) might be related.
So my idea was to issue multiple pip install command for each value in the array.

Converting the json array to a bash array didn't work me, so O went for the comma seperated options.

@pvizeli The next time i will raise an issue first to investigate any further, before a PR

kunago:
@mories76 : I also tried various syntaxes and everything failed except for a situation when I inserted just a single pip package; then it worked fine although with no dependencies installed.

PR #350, opened 2018-07-05
Modify git_pull to support reset as well as pull

matt-FFFFFF:
I have been using this git pull plugin but it used to complain that the local files had been modified when they hadn't. I could successfully do a git pull using remote tools like vscode but this plugin would fail.  So, I wanted to add an option to support ```git reset --hard``` to support overwriting local changes with the remote.

This has been exposed using an additional option called git_command.  This can be pull (default) or reset.

Other changes:
- I have also added a check that the local git repo is set to the correct remote URL.
- I have added 2 extra options to allow for the branch and remote names being non-standard. Sensible defaults have been specified for each.
- Modified the validate config function to report when there are changes even if auto restart is off

Am happy to update the docs if this is merged.

### DIFF
--- a/git_pull/CHANGELOG.md
+++ b/git_pull/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.8
+- Add option to use git reset instead of git pull
+- Validate git origin URL
+
 ## 4.7
 - Update Hass.io CLI to 1.3.1
 
--- a/git_pull/config.json
+++ b/git_pull/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Git pull",
-  "version": "4.7",
+  "version": "4.8",
   "slug": "git_pull",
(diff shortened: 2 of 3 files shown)


homeassistant:
Hi @matt-FFFFFF, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#350).

Thanks!

matt-FFFFFF:
Re: Travis is errors on double quotes

The double quotes are there, they are just escaped ```\"``` due to already being inside double quotes.

pvizeli:
Thanks

pvizeli:
Can you also update the documentation on website?

matt-FFFFFF:
Sure, will do this in the next day or so.

PR #357, opened 2018-07-15
Update to Configurator 0.3.1

danielperna84:
The changes of the configurator itself can be seen [here](https://github.com/danielperna84/hass-configurator/releases/tag/0.3.1) and [here](https://github.com/danielperna84/hass-configurator/releases/tag/0.3.0) (it's two versions of the configurator in one).

Because it now supports environment variables I got rid of the workaround using the `map.py` helper script. It's now all done in `run.sh`. Thanks to the hassos vmdk I could easily verify everything is working as it should.

I will update the documentation shortly with everything that has changed.

Question: does the line `"image": "homeassistant/{arch}-addon-configurator"` in the config.json mean that there is a compiled image that gets pulled when installing the addon? During testing the live build was pretty fast.

Oh, and openssh is installed to ensure ssh-based git repositories can be contacted.

Regarding the `enforce_basepath` option talked about in the other PR: I'll keep it. To make changes for SSH/Git users need the ability to edit files outside of the /config folder. So turning the option on is the safer setup, but some may want it the other way. As stated above, I'll take care of the documentation along with an explanation what it is there for.


pvizeli:
Please do only allow expose option they a Hass.io user really need:
- `ignore_ssl`
- `verify_hostname`
- `notify_service` ? Why need that be changeable?

danielperna84:
`ignore_ssl` is needed for people with self-signed certificates. The configurator accesses the HASS API to get the entities etc.. This doesn't work if SSL fails. And we can't assume every user will use a valid vertificate.
`verify_hostname` is for people who expose to the web. Requests by IP will be rejected in this case. But people who don't expose may use the IP. So this option needs to be configurable.
`notify_service` will in most cases stay as the persistent notification. But people that use the `SESAME` feature would benefit from another service that pushes messages to their phones. If they get a message `Your sesame has been used...` while they haven't accessed the configurator, they would know someone else did.

So in my opinion all of these settings make sense to be configurable.

pvizeli:
Like you can change the API endpoint :)

danielperna84:
I think your changes require that the `run.sh` needs to be changed too to handle null-values. I'll test it later today.

pvizeli:
That is all done bei Hass.io and that make it simple for user. Instead to make it hyper configurable, we make it simple and don't allow it to configure wrong.

danielperna84:
I'm fine with not having it in the defaults, so nobody has to decide if it should be set or not. But as it is a security feature, I don't think it is good to take it out completely. Having it in the schema at the bottom allows people who want to use it to set a value. And people that don't read the documentation don't even know about it and won't set it.
I fixed the documentation by the way. A correct value would be `example.domain.com`, not `true`. If that is not in the host header, then the request will get no result. That way shodan-scans won't get results if they only include the IP in the host header.

pvizeli:
What is the benefit of `verify_hostname`? It is very simple to spoof this too. A SSL certificate should be enaught that this feature is not needed. And if some one use only http should change it to https. I think this option give only a wring security feeling.

danielperna84:
Yes, it can easily be spoofed. But for that it has to be known. I have my own domain + wildcard certificate. My fqdn looks like `f4e4iuth4kuh4lkjlskf.mydomain.com`. When the configurator is exposed directly, then the hostname is irrelevant and shodan etc. would be able to gather information. But with this option enabled, requests without the correct hostname get a 403 forbidden response. So attackers only know that there's some sort of HTTP server running, but don't get to the point where they could brute-force the basic-auth passwort.

Yes, it adds only a little bit of security, but with this it's possible to prevent shodan from indexing the basic auth realm (_HASS Configurator_).
It's not part of the default config, only optional. So I don't see a reason why it should not be included. People don't have to use it. And if they don't read the documentation they wouldn't even know about that option. Existing installations would keep running like they already do.

To put it simple: this is like having a virtual host in Apache / nginx where you only get the upstream with the correct hostname, otherwise you're being rejected.

danielperna84:
I removed the option. I don't want to delay the release even more.

pvizeli:
The usecase is okay

PR #364, opened 2018-08-03
Disable AppArmor for OCCU

pvizeli:
I have no time to add a AppArmor Profile for OCCU, I will do it later


GOVEN007:
When could we expect rpi_rf working? Thank you.

PR #368, opened 2018-08-08
Upgrade Configurator to 0.3.2

danielperna84:
- Fixed numeric password regression
- Fixed list-options (ALLOWED_NETWORKS, BANNED_IPS, IGNORE_PATTERN)
- Improved handling of UTF-8 encoded files (minor issue)

### DIFF
--- a/configurator/CHANGELOG.md
+++ b/configurator/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## 1.2
+
+### Configurator 0.3.2
 - Fixed numeric password regression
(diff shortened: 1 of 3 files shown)


danielperna84:
Please don't merge yet. There seems to be another issue: https://github.com/danielperna84/hass-configurator/issues/112

danielperna84:
@pvizeli 
Could you read through the issue quickly? To me it doesn't seem to be an error with the configurator. It's calling the API the same way as before. Did something in hassio with calling the API change? Or may add-ons like pihole break access to the API?

danielperna84:
It seems as if updating the supervisor fixed the problem in one case. If this makes sense, then the PR can be merged.

Issue #374, opened 2018-08-16
check_config not checking new config if container not destroyed


Censored3:
Hi

I'm running a manual check_config container on another machine to check my config before pushing it to my HA installation. I don't destroy the container everytime to save time and not let the HA install again every boot.
This, however, causes an issue because the config is copied to another location before checking, and the temp dir is still filled with my previous config. The files are not overwritten with the new config.
While I know that in the proper add-on operation, the container is recreated everytime and this issue is thus not triggered, I do propose a change for a little more optimal execution.

* Add rm -rf /tmp/config before cp -r /config /tmp/config,
* Just check the config in /config instead of copying it over to /tmp first - I don't think there's really a reason not to just check the config in /config?

Thanks!

PR #383, opened 2018-08-20
check_config: do not copy config to /tmp on start - #374

Censored3:
https://github.com/home-assistant/hassio-addons/issues/374

I just made this change for myself, to be able to run this addon standalone on my dev machine.

Please feel free to reject this PR if there is indeed a valid reason to copy the config to /tmp first - in that case I will modify it to first remove the config in /tmp before copying.

Thx

### DIFF
--- a/check_config/config.json
+++ b/check_config/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Check Home Assistant configuration",
-  "version": "0.7",
+  "version": "0.8",
   "slug": "check_config",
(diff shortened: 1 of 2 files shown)


Censored3:
I just made this change for myself, to be able to run this addon standalone on my dev machine.

Please feel free to reject this PR if there is indeed a valid reason to copy the config to /tmp first - in that case I will modify it to first remove the config in /tmp before copying.

Thx

Censored3:
will add new PR with added bugfix

PR #384, opened 2018-08-20
Fix repeat option detection - Closes home-assistant/hassio-addons#375

Censored3:
Fixes https://github.com/home-assistant/hassio-addons/issues/375


Censored3:
sorry faulty request

Issue #387, opened 2018-08-20
check_config: stops working for HA>=0.76.0


Censored3:
```
[Info] Start install HomeAssistant latest
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-y32z_anc/cffi/
[Error] Install HomeAssistant: Collecting homeassistant
  Using cached https://files.pythonhosted.org/packages/7a/9f/5c6951442b94dff3530aef3058cf7429fd7839230cddf481fc76437b357f/homeassistant-0.76.1-py3-none-any.whl
Collecting jinja2>=2.10 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/7f/ff/ae64bacdfc95f27a016a7bed8e8686763ba4d277a78ca76f32659220a731/Jinja2-2.10-py2.py3-none-any.whl
Requirement already satisfied: pip>=8.0.3 in /usr/lib/python3.6/site-packages (from homeassistant) (18.0)
Collecting attrs==18.1.0 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/41/59/cedf87e91ed541be7957c501a92102f9cc6363c623a7666d69d51c78ac5b/attrs-18.1.0-py2.py3-none-any.whl
Collecting astral==1.6.1 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/01/2b/0b60f1cef7d2a7a863cd06bae5fc7e3553803739c4834e921b4996920f76/astral-1.6.1-py2.py3-none-any.whl
Collecting pyyaml<4,>=3.13 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/9e/a3/1d13970c3f36777c583f136c136f804d70f500168edc1edea6daa7200769/PyYAML-3.13.tar.gz
Collecting cryptography==2.3.1 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/22/21/233e38f74188db94e8451ef6385754a98f3cad9b59bedf3a8e8b14988be4/cryptography-2.3.1.tar.gz
Collecting async-timeout==3.0.0 (from homeassistant)
  Using cached https://files.pythonhosted.org/packages/96/0f/e6357458c87fb4ed8f3df215773f3caad40968f10e05552cbd8bd28415e4/async_timeout-3.0.0-py3-none-any.whl
Collecting aiohttp==3.3.2 (from homeassistant)
[...]

PR #388, opened 2018-08-20
check_config: bugfixes

Censored3:
#387: add-on stops working for HA >= 0.76.0 -- added OS pkgs build-base python3-dev libffi-dev openssl-dev

Regarding #374 :
I just made this change for myself, to be able to run this addon standalone on my dev machine.
Please feel free to adjust this PR if there is indeed a valid reason to copy the config to /tmp first - in that case I will modify it to first remove the config in /tmp before copying.

Thx

### DIFF
--- a/check_config/Dockerfile
+++ b/check_config/Dockerfile
@@ -5,7 +5,7 @@ FROM $BUILD_FROM
 ENV LANG C.UTF-8
 
 # Setup base
-RUN apk add --no-cache jq python3 \
+RUN apk add --no-cache jq python3 python3-dev libffi-dev openssl-dev \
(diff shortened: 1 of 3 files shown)


pvizeli:
We copy the config, because `/config` is not rw for this add-on.

Censored3:
Dear @pvizeli 

I feel cp -fr is not an adequate, foolproof solution. It will indeed overwrite the config already in place, however, files deleted in the production config are not deleted in the add-ons destination.
Doing an `rm -rf /tmp/config` before the cp could solve this; which _should_ be safe.
Otherwise, could doing an rsync maybe be a solution. It'll probably need to be added to the container though...

PR #413, opened 2018-09-24
Proxy API for UI development

pvizeli:
Run this add-on and use the Token from log for you local home assistant to develop the UI


frenck:
Ow, this is useful!
I've been fetching tokens like a mad man lately.

PR #437, opened 2018-11-03
New SMB add-on

pvizeli:


samba/smb.conf
```diff
@@ -13,6 +15,47 @@
...
+   path = /share
+
+   valid users = %%USERNAME%%
+   force user = root
+   force group = root
+
+[backup]
+   browseable = yes
+   writeable = yes
+   path = /share
```
corbanmailloux: @pvizeli It looks like the `[share]` and `[backup]` shares have the same path.
Should this be `/backup` instead?

tjorim: I think you are correct, opened PR https://github.com/home-assistant/hassio-addons/pull/439

PR #442, opened 2018-11-04
Update to new mqtt generation

pvizeli:


Diaoul:
*I cannot find any description of this change nor is there any related issue so I post here.*

This PR broke my MQTT instance. May I ask what motivated those changes?

I use clients with SSL certificate verification option:
```
listener 8883
cafile /path/to/ca.crt
certfile /path/to/server.crt
keyfile /path/to/server.key
require_certificate true
```

Now I need to reverse the config generation script and disable SSL so that I can add it back with proper security options. And I still get a warning message.

Also, now this part bothers me a little bit:
> You don’t need do this because you can use Home Assistant users too without any configuration.

Why would I want to allow HA users to log into MQTT and publish messages? How can I turn this off? Previous behaviour didn't allow that.

Generally speaking, I think configuration should be split into `.conf.add` files that will append (or prepend) some configuration to the default. And `.conf` files would completely override the config file.

PR #448, opened 2018-11-06
Update  configurator to 0.3.4

danielperna84:
Changelog:
- Display covered last item in file browser
- Proper displaying of file browser
- Added support for long-lived JWT access tokens
- Spelling fix
- Added env_var tag exception
- Event observer: disconnect on modal close
- Added editor option to hide file / directory stats
- Added hass_ws_api option to correctly prefill the websocket URL
- Open common image formats in new tab
- Update dependencies
- Added allowed_domains option

No new options have been added. This is mostly cosmetic + the JWT-support.

What I don't know about is if the new support for the long-lived access tokens may break the add-on. Currently in `run.sh` we have `HC_HASS_API_PASSWORD=$HASSIO_TOKEN`. I don't know about the internals of this `$HASSIO_TOKEN`. What the configurator does internally:
If the token has exactly 2 dots (JWT-Tokens are `asdf.asdf.asdf`), then it will use the new authorization header for the long-lived access tokens. If there are no 2 dots, then the old header for the `api_password` will be used to keep the configurator backwards-compatible. If this is a problem we somehow have to address this issue.

And as we talked about before, there's some other auth-stuff you want to implement. Let me know what exactly you need or feel free to push relevant changes yourself.


pvizeli:
HASSIO_TOKEN can also be used as long live token. Home Assistant use a JWT token but that need not be a JWT token for Baerer tokens.

danielperna84:
Ok, so the way I am dealing with the token won't break the existing functionality. Then the only this missing for this PR are those other changes you've talked about.

pvizeli:
Yeah, all is fine. Will hook into this after TF

danielperna84:
Any news on this?

danielperna84:
I don't know about possible internal changes of hassio, but with everything you removed people loose a lot of configuration options. Without SSL support they won't be able to embed the configurator as a panel (a lot of users are doing that), and removing the username and password is a security issue. This would only make sense if hassio includes some sort of reverse proxy that's in front of the port 80 configurator, and that proxy makes use of the authentication of Home Assistant. Is that the case?

GaryOkie:
> Yeah, all is fine. Will hook into this after TF

This update will allow  Hassio **SSL** users to easily incorporate the configurator into the left frame, with pass-through authentication - is that right?     If so, wow - great!  Is there any further progress?

danielperna84:
Version 0.3.4 of the configurator has been released. I have commited the changes required for the new version. The new `hass_ws_api` option is used to prefill the input-field for the URL of the websocket api, which is used by the event observer. This can not be auto-detected. Most of the times it should be the same as the base URL in Home Assistant, but we can't rely on that. THat's why I chose to let the users configure it manually if they want to use it.

pvizeli:
I'm sorry about the delay. I close this PR now.

Why? This add-on will be the first Add-on they use the ingress function. I release today a new PR they run on the internal ingress routing.

PR #475, opened 2018-12-05
Support the changes in specific files without restarting home assistant

adamgoodbar:
Support the changes in specific files without restarting home assistant.  Files such as the lovelace config can be picked up without a restart of the service, decreasing impact of changes.

Especially since services like zwave take time to start back up after a reload, any time we can incorporate change without a restart is a bonus for the user experience.

### DIFF
--- a/git_pull/CHANGELOG.md
+++ b/git_pull/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 6
+- Allow to disable Home Assistant restart for specific file changes
+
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @adamgoodbar, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#475).

Thanks!

Issue #492, opened 2018-12-21
Configurator does not start in default config


TKAB:
I freshly installed hassio on a Raspberry Pi, added the `Configurator` addon, but it didn't work. It took me some time to figure out that the addon didn't start and that it only starts if there's a password set. This is a problem because there's no error message in the UI and for someone new to the project it's not obvious that the service is not running.

The default config should already have a password set or the description in the UI should explicitly point out that the default config needs to be changed - since it's strange behavior for a default config to not work.

PR #493, opened 2018-12-21
Add hint to clarify Configurator needs password for running

TKAB:
As explained in #492, the current default config does not allow the addon to be started. This password should allow the addon to run without modification and therefore remove a source of frustration for users.

### DIFF
--- a/configurator/config.json
+++ b/configurator/config.json
@@ -19,7 +19,7 @@
   ],
   "options": {
     "username": "admin",
-    "password": null,
+    "password": "admin",
     "ssl": false,
     "certfile": "fullchain.pem",
     "keyfile": "privkey.pem",
--- a/configurator/config.json
+++ b/configurator/config.json
@@ -2,7 +2,7 @@
   "name": "Configurator",
   "version": "2",
   "slug": "configurator",
-  "description": "Browser-based configuration file editor for Home Assistant.",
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @TKAB, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#493).

Thanks!

frenck:
The default password is not set for security reasons.

TKAB:
I understand that, but it is still a point of frustration right at the start, because the Configurator is what you need to get a first shot at editing `configuration.yaml`. An error message or a clear hint in the UI would also help. How about that?

frenck:
@TKAB That I would welcome as a PR.

pvizeli:
There is a update in progress with support the home-assistant auth backend

TKAB:
@pvizeli So does it make sense to improve something then or will it be a long time until that update is ready?

TKAB:
That's a minimal fix. The drawback is, that the hint is visible in the add-ons store overview.

frenck:
Sorry, @TKAB, but that is just awful.

Let us close this PR, I know it could be better, but as @pvizeli already mentioned, the add-on will move to Home Assistant authentication pretty soon.

Nevertheless, thank you for your effort in helping improve things and opening a PR.

PR #504, opened 2019-01-14
git_pull: don't break repeating function when internet connection lost once

der-optimist:
Problem: if repeat_interval is set, but internet connection is not available during a check, the script exits and the repeating function is broken. This should be solved by the change, and by the if statements it's still ensured that the following step is only executed if the previous exited without an error

### DIFF
--- a/git_pull/run.sh
+++ b/git_pull/run.sh
@@ -215,10 +215,13 @@ function validate-config {
 cd /config || { echo "[Error] Failed to cd into /config"; exit 1; }
 
 while true; do
-    check-ssh-key
-    setup-user-password
-    git-synchronize
-    validate-config
+    if check-ssh-key ; then
+        if setup-user-password ; then
+            if git-synchronize ; then
+                validate-config
+            fi
+        fi
+    fi
      # do we repeat?
     if [ ! "$REPEAT_ACTIVE" == "true" ]; then
         exit 0


homeassistant:
Hi @der-optimist, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#504).

Thanks!

pvizeli:
do think that work because the functions have no return value and don't break on an error?

pvizeli:
I think there is a `set` option they break a function after an error. Maybe with this option it could work

der-optimist:
I don't think that this will work, because the exit commands in the functions terminate the whole script. For some errors that's correct I think (e.g. for "Creation of backup directory failed" => exit the script). But for the ones which are related to internet connection and which can be gone a few minutes later, for those I think we should replace exit by return. What do you think?


{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/home-assistant/hassio-addons","title":"home-assistant/hassio-addons","subtitle":"GitHub repository","main_image_url":"https://github.githubassets.com/images/email/message_cards/header.png","avatar_image_url":"https://github.githubassets.com/images/email/message_cards/avatar.png","action":{"name":"Open in GitHub","url":"https://github.com/home-assistant/hassio-addons"}},"updates":{"snippets":[{"icon":"PERSON","message":"@pvizeli in #504: I think there is a `set` option they break a function after an error. Maybe with this option it could work"}],"action":{"name":"View Pull Request","url":"https://github.com/home-assistant/hassio-addons/pull/504#issuecomment-455548807"}}}
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/home-assistant/hassio-addons/pull/504#issuecomment-455548807",
"url": "https://github.com/home-assistant/hassio-addons/pull/504#issuecomment-455548807",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

PR #511, opened 2019-01-21
continue script when internet connection lost once

der-optimist:
changed some "exit 1" to "return 1" in function git-synchronize. By this, the script should not terminate when there's no internet connection during a repeat check. I left "exit 1" for the initial, critical and config topics => script should terminate here I think.

### DIFF
--- a/git_pull/run.sh
+++ b/git_pull/run.sh
@@ -126,13 +126,13 @@ function git-synchronize {
 
             # Always do a fetch to update repos
             echo "[Info] Start git fetch..."
-            git fetch "$GIT_REMOTE" || { echo "[Error] Git fetch failed"; exit 1; }
+            git fetch "$GIT_REMOTE" || { echo "[Error] Git fetch failed"; return 1; }
 
             # Prune if configured
             if [ "$GIT_PRUNE" == "true" ]
             then
               echo "[Info] Start git prune..."
-              git prune || { echo "[Error] Git prune failed"; exit 1; }
+              git prune || { echo "[Error] Git prune failed"; return 1; }
             fi
 
             # Do we switch branches?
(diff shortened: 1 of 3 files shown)


git_pull/run.sh
```diff
@@ -217,8 +217,9 @@ cd /config || { echo "[Error] Failed to cd into /config"; exit 1; }
 while true; do
     check-ssh-key
     setup-user-password
-    git-synchronize
-    validate-config
+    git-synchronize; if [ $? -ne 1 ] ; then
```
pvizeli: please use style: `if git-synchronize; then`

der-optimist: you are totally right, that's much better. I changed it

git_pull/run.sh
```diff
@@ -126,13 +126,13 @@ function git-synchronize {
 
             # Always do a fetch to update repos
             echo "[Info] Start git fetch..."
-            git fetch "$GIT_REMOTE" || { echo "[Error] Git fetch failed"; exit 1; }
+            git fetch "$GIT_REMOTE" || { echo "[Error] Git fetch failed"; return 1; }
```
pvizeli: `{}` is a function, I think you return only the function they call the echo. Maybe `git xy || echo ""; return 1`?

der-optimist: That's exactly what I thought first too, but then I did a quick test:

function test_ok {
echo "function without error" || { echo "[Error] text"; return 1; }
echo "this is executed"
}
function test_nok {
echo "function with error"
cause-an-error || { echo "[Error] text"; return 1; }
echo "this is not executed"
echo "this is not executed"
}
test_ok ; echo "$?"
test_nok ; echo "$?"

And it works like it should. If I use
`... || echo ""; return 1`
then the `return 1` is executed also if there is no error => this doesn't work. 
I saw that with the function it does what I think we need, but I not exactly unterstand why. 
The exit code of `test_ok` is 0, the exit code of `test_nok` is 1 and `echo "this is not executed"` is really not executed.

der-optimist: OK, thanks to stackoverflow I know now that `{}` is not a function but only a "command grouping". That's why the `git ... || { echo "[Error] text"; return 1; }` stops the function `git-synchronize` as we need it

der-optimist:
Is this what you wanted me to do? I'm not an experienced github user yet :-/

PR #526, opened 2019-02-27
Move custom deps to python 3.7 base

zewelor:
After move home-assistant to python 3.7, I think this also should be moved. When python deps are installed, they ends up in lib/python3.6/site-packages/ , so ha doesn't pick it up, if I'm correct.


pvizeli:
I moved to https://github.com/home-assistant/hassio-addons-development

PR #557, opened 2019-03-23
nginx-proxy: Use of variables in customizing

HeMUC:
Postpone the variable replacement to allow usage of %%FULLCHAIN%%, %%PRIVKEY%% and %%HSTS%% in customizing. %%DOMAIN%% presumably not needed but no harm either.

(As this is my first fork, fix and pull request please foregive any shortcomings...)


pvizeli:
You need to update the script to update all config files too with sed otherwise it will not work.

PR #570, opened 2019-04-16
Deconz update

pvizeli:
- Remove ingress
- Update deconz


TheZoker:
@pvizeli Is there any specific reason, why ingress support was removed again?

frenck:
@TheZoker, please don't use comments on closed/merged PR's. Those are hard to track.
Your answer: The application has issues running behind a reverse proxy in a subfolder (which is basically what Ingress is). We had to remove it in order to keep things working.

Issue #523, opened 2019-02-20
Unable to limit logs


sickidolderivative:
I want to limit the amount of messages in mosquitto addon logs. I have a simple mqtt client that reconnects every second to read and/or post. That results in excessive logs.

I run Home Assistant 0.87.0 on HassOS 2.8.

My addon setup is the following:
{
"logins": [],
"anonymous": false,
"customize": {
"active": true,
"folder": "mosquitto"
},
"certfile": "fullchain.pem",
"keyfile": "privkey.pem"
}

Content of /share/mosquitto/mosquitto.conf file is the following:
log_type none
connection_messages false
log_dest none

Despite the above settings, I get multiple [INFO] messages on every login:

[INFO] found mqtt on Home Assistant
[INFO] found mqtt on Home Assistant
[INFO] found mqtt on Home Assistant
[INFO] found mqtt on Home Assistant

mqtt is the username in HA used for login by mosquitto client.

sickidolderivative:
There's even more messages in the System Log (Hass.io->System):
19-02-21 06:34:26 INFO (MainThread) [hassio.auth] Auth request from core_mosquitto for mqtt
19-02-21 06:34:27 INFO (MainThread) [hassio.auth] Success login from mqtt
19-02-21 06:34:27 INFO (MainThread) [hassio.auth] Auth request from core_mosquitto for mqtt
19-02-21 06:34:29 INFO (MainThread) [hassio.auth] Success login from mqtt
19-02-21 06:34:30 INFO (MainThread) [hassio.auth] Auth request from core_mosquitto for mqtt
19-02-21 06:34:31 INFO (MainThread) [hassio.auth] Success login from mqtt

Is there a way to stop this madness?

sickidolderivative:
With the release of Mosquitto 4.1 the addon logs are flooded with even more messages:

`[INFO] found mqtt on local database
1552314298: |-- getuser(mqtt) AUTHENTICATED=1 by http
1552314298: |-- mosquitto_auth_acl_check(..., client id not available, mqtt, homie/mPowerPRO/port2/voltage, MOSQ_ACL_WRITE)
1552314298: |-- url=http://127.0.0.1:8080/superuser
1552314298: |-- data=username=mqtt&password=&topic=&acc=-1&clientid=
1552314298: |-- aclcheck(mqtt, homie/mPowerPRO/port2/voltage, 2) SUPERUSER=Y by http
[...]

PR #572, opened 2019-04-17
Disable verbose logging for http auth plugin

signaleleven:
Solves (at least in part) #523 

Avoids logging cleartext passwords when using the http auth plugin.

### DIFF
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 4.2
+- Enable quiet logging for http auth plugin
+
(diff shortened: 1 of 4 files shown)


homeassistant:
Hi @signaleleven, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#572).

Thanks!

signaleleven:
I updated the changelog and made the change optional via config (although I chose to enable it by default since dumping passwords in the logs by default is to be avoided, IMHO)

PR #592, opened 2019-06-01
configurator: Add git SSH support

adamgreg:
Add an option to the Configurator add-on to define an SSH key for use by git. Based on the git_pull add-on.

### DIFF
--- a/configurator/config.json
+++ b/configurator/config.json
@@ -33,14 +33,16 @@
     "enforce_basepath": false,
     "ignore_pattern": [
       "__pycache__"
-    ]
+    ],
+    "ssh_keys": []
   },
   "schema": {
     "dirsfirst": "bool",
     "enforce_basepath": "bool",
     "ignore_pattern": [
       "str"
-    ]
+    ],
+    "ssh_keys": ["str"]
   },
   "image": "homeassistant/{arch}-addon-configurator"
 }
--- a/configurator/config.json
+++ b/configurator/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Configurator",
-  "version": "3.4",
+  "version": "3.5",
   "slug": "configurator",
   "description": "Browser-based configuration file editor for Home Assistant",
   "url": "https://home-assistant.io/addons/configurator",
--- a/configurator/data/run.sh
+++ b/configurator/data/run.sh
@@ -6,6 +6,28 @@ ENFORCE_BASEPATH=$(bashio::config 'enforce_basepath')
 IGNORE_PATTERN="$(bashio::jq "/data/options.json" ".ignore_pattern")"
 WAIT_PIDS=()
 
+# If any SSH key files are defined in the configuration options, add them for use by git
+if bashio::config.has_value "ssh_keys"; then
+  # Start the SSH agent
+  bashio::log.info "Starting SSH agent"
+  eval "$(ssh-agent -s)"
+
+  # Add the keys defined in the configuration options
+  while read -r filename; do
+    if bashio::fs.file_exists "$filename"; then
+      bashio::log.info "Adding SSH private key file \"$filename\""
+      ssh-add -q "$filename"
+    else
+      bashio::log.error "SSH key file \"$filename\" not found"
+    fi
+  done <<< "$(bashio::config 'ssh_keys')"
+
+  # Disable strict host key checking
+  mkdir -p ~/.ssh
+  echo "Host *
+    StrictHostKeyChecking no" > ~/.ssh/config
+fi
+
 # Setup and run Frontend
 sed -i "s/%%PORT%%/8080/g" /etc/nginx/nginx-ingress.conf
 sed -i "s/%%PORT_INGRESS%%/8099/g" /etc/nginx/nginx-ingress.conf
--- a/configurator/CHANGELOG.md
+++ b/configurator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.5
+
+- Add support for SSH keys
+
 ## 3.4
 
 - Adds documentation to add-on repository
--- a/configurator/Dockerfile
+++ b/configurator/Dockerfile
@@ -5,7 +5,7 @@ FROM $BUILD_FROM
 ARG CONFIGURATOR_VERSION
 ARG HASSIO_AUTH_VERSION
 RUN apk add --no-cache \
-    git nginx nginx-mod-http-lua lua-resty-http \
+    git nginx nginx-mod-http-lua lua-resty-http openssh-client \
     && git clone --depth 1 -b ${HASSIO_AUTH_VERSION} \
       "https://github.com/home-assistant/hassio-auth" \
     && cp -f hassio-auth/nginx-frontend/ha-auth.lua /etc/nginx/ \
--- a/configurator/README.md
+++ b/configurator/README.md
@@ -57,7 +57,8 @@ Add-on configuration:
   "enforce_basepath": false,
   "ignore_pattern": [
     "__pycache__"
-  ]
+  ],
+  "ssh_keys": []
 }
 ```
 
@@ -75,6 +76,10 @@ If set to `true`, access is limited to files within the `/config` directory.
 
 This option allows you to hide files and folders from the file browser tree.
 By default, it hides the `__pycache__` folders.
+
+### Option: `ssh_keys` (required)
+
+A list of filenames containing SSH private keys. These can be used to allow for access to remote git repositories.
 
 ## Known issues and limitations
 


homeassistant:
Hi @adamgreg, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#592).

Thanks!

configurator/data/run.sh
```diff
@@ -4,8 +4,34 @@ set -e
...
 IGNORE_PATTERN="$(bashio::jq "/data/options.json" ".ignore_pattern")"
+
+DEPLOYMENT_KEY=$(bashio::config 'deployment_key')
+DEPLOYMENT_KEY_PROTOCOL=$(bashio::config 'deployment_key_protocol')
+
 WAIT_PIDS=()
 
+# Adds the SSH deployment key specified in the config, for use by git
+function add-ssh-key {
+    echo "[Info] Start adding SSH key"
```
frenck: Please use Bashio for this.

configurator/data/run.sh
```diff
@@ -4,8 +4,34 @@ set -e
...
+    mkdir -p ~/.ssh
+
+    (
+        echo "Host *"
+        echo "    StrictHostKeyChecking no"
+    ) > ~/.ssh/config
+
+    echo "[Info] Setup deployment_key on id_${DEPLOYMENT_KEY_PROTOCOL}"
```
frenck: Please use Bashio for this.

configurator/data/run.sh
```diff
@@ -4,8 +4,34 @@ set -e
...
+    while read -r line; do
+        echo "$line" >> "${HOME}/.ssh/id_${DEPLOYMENT_KEY_PROTOCOL}"
+    done <<< "$DEPLOYMENT_KEY"
+
+    chmod 600 "${HOME}/.ssh/config"
+    chmod 600 "${HOME}/.ssh/id_${DEPLOYMENT_KEY_PROTOCOL}"
+}
+
+# Add SSH key for git, if defined
+if [[ ! -z ${DEPLOYMENT_KEY} ]]; then add-ssh-key; fi
```
frenck: Use Bashio for this, e.g.:

`if bashio::config.has_value "deployment_key"; then add-ssh-key; fi`

adamgreg: Done

configurator/data/run.sh
adamgreg: Done

configurator/data/run.sh
adamgreg: Done

configurator/config.json
```diff
@@ -33,14 +33,18 @@
...
   },
   "schema": {
     "dirsfirst": "bool",
     "enforce_basepath": "bool",
     "ignore_pattern": [
       "str"
-    ]
+    ],
+    "deployment_key": ["str"],
+    "deployment_key_protocol": "match(rsa|dsa|ecdsa|ed25519|rsa)"
```
frenck: Been thinking about this line... because it made no sense to me why it is even there...

Figured, because this PR is writing the deployment key to disk.

Just a thought, wouldn't it make more sense to start the ssh-agent and just add the key in on the fly?

frenck:
I really think we should use the ssh-agent, instead of writing keys to disk.

In short, it would be something like this:

```bash
eval $(ssh-agent -s)
ssh-add <(echo "${SSH_KEY}")
```

(ish, untested top of mind code, might be incorrect).

configurator/config.json
adamgreg: Sounds good to me - I just copied the approach (and code!) of the git_pull add-on. I'm completely new to Docker and not so hot on bash. I'll try using `ssh-add` instead, as suggested.

frenck:
The configurator gives full blown GIT support, right?
That makes this an SSH key... not to confuse with a deployment key (which is an SSH key that is only allowed to pull), so I would consider renaming that config option as well.

adamgreg:
Ok, I think that covers all the comments so far. Thanks a lot for the prompt review, @frenck !

configurator/data/run.sh
```diff
@@ -6,6 +6,21 @@ ENFORCE_BASEPATH=$(bashio::config 'enforce_basepath')
...
 WAIT_PIDS=()
 
+# If an SSH key is defined in the configuration options, add it for use by git
+if bashio::config.has_value "ssh_key"; then
+  bashio::log.info "Adding SSH key, for use by git"
+
+  # Start the SSH agent and add the key
+  eval $(ssh-agent -s)
+  KEY=$(bashio::config 'ssh_key')
+  ssh-add -q <(echo "$KEY")
```
frenck: According to the schema in `config.json` this a list of keys?

frenck:
@adamgreg Left an additional comment on the new code.
Also, the build fails. Please make sure it passes.

configurator/data/run.sh
adamgreg: This is again a straight copy from the git_pull add-on. The lines of a single key are split up into a JSON array. See the example here: https://www.home-assistant.io/addons/git_pull/ . Multi-line strings are not supported in JSON, so it looks like we're stuck with either an array, or a very long string on a single line.

I've just tried passing a key to `ssh-add` with all the newlines removed (both with & without the header & footer lines), and it said the format was invalid. I think this means that if you wanted to put the whole key in a single JSON string you would have to add "\n" control characters. I think the method used by git_pull might be the best we can do.

In any case, I think it makes sense for the format of the option to match the format used by git_pull, since the two add-ons are likely to be used together if the configuration is managed in git.

rbray89:
@frenck @adamgreg What is the status of this? Was going to work on the same issue until I checked pending pull requests.

adamgreg:
Hi @rbray , I think I've addressed everything. I've been using it locally for the past month with no problems.

configurator/data/run.sh
pvizeli: You need address this before we can merge it. If they is an array, you need loop over it

frenck:
@adamgreg There is still one thing to address. You can take a look at the ssh addons to see how to solve it.

configurator/data/run.sh
adamgreg: I'm sorry, but I'm really not sure what you'd like me to do here - it's not an array of keys, it's a single key split into an array of lines. The approach is exactly the same as used in the git_pull add-on in this same repository.

I've tested the script and it successfully adds the key to the agent. Is something more needed? I'd be happy to make changes, but I'd need more details.

frenck: @adamgreg You'll need to loop over the keys:

Example: <https://github.com/home-assistant/hassio-addons/blob/6af0f83c55bb9db3bdff9524cf3f0ec361148baf/ssh/data/run.sh#L11>

adamgreg: It's only 1 key, and the code works. Nothing to loop over.

The ssh add-on deals with a list of (short) authorised key strings. The code in this PR (as well as the code in this repo's git_pull add-on, which I assume is acceptable) deals with a single (long) private key, split at each newline into an array of strings. Honestly, the script works.

frenck: Code that works does not have to be implemented correctly @adamgreg. This has to be adjusted as pointed out now a couple of time.

Let me lay it out for you:

You have specified in the add-on config, that a list of keys can be provided:

https://github.com/home-assistant/hassio-addons/blob/5796b073284fa477f8bcf88b3dbb90918d3f34c2/configurator/config.json#L45

However, in the code, only 1 item is picked and the code does not take into account the configuration can contain a list of items. A list needs to be iterated.

Does it work if you enter a single key? Sure. Is the list handled? Nope.

I'm currently unsure what the problem is at this point. You only have to modify 3 lines, as I already kindly provided by linking the SSH add-on example.

adamgreg: The schema specifies a list of strings, not a list of keys. It's a single key split by line into an array. It does not support multiple keys. It just works the same way as the git_pull plugin.

pvizeli: that is correct

adamgreg: Thanks! I gave a bit more of an explanation in my first comment: https://github.com/home-assistant/hassio-addons/pull/592#discussion_r289639143 . The private keys are just a bit too long to paste into a single JSON string!

Bashio/jq does a great job here of getting the whole key from JSON, with whitespace between the lines - just right for ssh-add.

Having a key split like that in the config doesn't look great, but I think it's better than squeezing it onto one line. It's good to be consistent with the format of the config of the git_pull add-on, so at least people who use both (like my client) can copy & paste the key between them.

adamgreg: Here's a blog post I just found about this approach to storing keys and other long strings in JSON, that might be helpful: https://www.gun.io/blog/multi-line-strings-in-json

frenck: I get that, although, JSON can still do that using newline char `\n`.
So how would I add multiple keys in that case?

frenck: Second thought, why not convert this in to a simple file list?
e.g., point directly to a key `/ssl/my.key` or `/config/ssh/id_rsa`.

snarky-snark:
@frenck @adamgreg It looks like this is on track, but I wanted to check in to see if there are any hold ups. Eagerly awaiting this pull request! :+1:

configurator/data/run.sh
adamgreg: Ok, that does sound like a good idea actually. I've made the change.

adamgreg:
Is there something else holding this up?

Levdbas:
Woud like to see this as well!

PR #600, opened 2019-06-03
dhcp_server: Fixes crash when using FQDN as domain

frenck:
Fixes a crash when using an FQDN as a domain, e.g., `home.local` instead of just `local` or `home.


pvizeli:
thanks

PR #601, opened 2019-06-03
Fix mqtt pw generator

pvizeli:


mosquitto/run.sh
```diff
@@ -111,8 +107,8 @@ fi
 
 # Prepare System Accounts
 if [ ! -e "${SYSTEM_USER}" ]; then
-    HOMEASSISTANT_PW="$(create_password)"
-    ADDONS_PW="$(create_password)"
+    HOMEASSISTANT_PW="$(pwgen 64 1)"
```
frenck: If you want more secure, add `-s`.

PR #608, opened 2019-06-06
Fixes deconz otau download issues

### LINKED ISSUE #607: deconz ota download scripts are broken
For the ota tools the following tools are missing:
- iproute2
- wget
- iputils-ping

further the directory /data/otau/ isn't created before downloading, resulting in more errors

mjrider:
Fixes #607 the problem that ota files are not downloaded

### DIFF
--- a/deconz/Dockerfile
+++ b/deconz/Dockerfile
@@ -22,6 +22,9 @@ RUN apt-get update \
         libqt5websockets5 \
         libqt5widgets5 \
         sqlite3 \
+	iputils-ping \
+	iproute2 \
+	wget \
     && rm -rf /var/lib/apt/lists/* \
     && if [ "${BUILD_ARCH}" = "armhf" ] || [ "${BUILD_ARCH}" = "aarch64" ]; \
         then \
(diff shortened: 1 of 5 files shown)


deconz/data/ika-otau-dl.sh
```diff
@@ -2,6 +2,8 @@
 
 URL_IKEA="http://fw.ota.homesmart.ikea.net/feed/version_info.json"
 
+mkdir "/data/otau"
```
pvizeli: Create this on `run.sh` before we start OTA scripts. `mkdir -p ....`

deconz/Dockerfile
```diff
@@ -22,6 +22,9 @@ RUN apt-get update \
         libqt5websockets5 \
         libqt5widgets5 \
         sqlite3 \
+	iputils-ping \
+	iproute2 \
```
pvizeli: why iproute2 ?

mjrider: because the command `ip` is in that package, and deconz is using that for the detecting of working internet

pvizeli:
Update version +0.1 and add a release node

mjrider:
@pvizeli do you agree with merging this ?

PR #623, opened 2019-07-03
Add cafile and require_certificate options

Diaoul:
This is a non-breaking change to add cafile customization (currently hardcoded to certfile) and require_certificate option.

I have no idea how to test this ATM. Hopefully the CI will provide a test docker image I'll be able to try out. :crossed_fingers: 

Documentation PR: https://github.com/home-assistant/home-assistant.io/pull/9771


Diaoul:
I cannot test locally due to this error:

> ERROR (SyncWorker_10) [hassio.docker.interface] Can't install homeassistant/armv7-addon-mosquitto:5.1 -> 404 Client Error: Not Found ("manifest for homeassistant/armv7-addon-mosquitto:5.1 not found").

frenck:
Temporarily remove the "image" option form the `config.json` in the add-on. That will trigger a local build.

Diaoul:
Thanks I'll try that.

Le jeu. 4 juil. 2019 à 09:58, Franck Nijhof <notifications@github.com> a
écrit :

Diaoul:
I didn't find a way to make optional keys in the schema of the config. I tried to add `?` as documented but it still wouldn't start unless I added the missing configuration keys.

Hence, the latest push makes those changes obvious to the user (no more cafile magic) and allows for `require_certificate`. All that comes with non-breaking defaults although for people having already configured the SSL it might not be backward compatible. This is why I bumped the major.

Let me know if this is satisfying. :cake:

pvizeli:
If you add `?`, you need also add a default to `jq` command

mosquitto/data/run.sh
```diff
@@ -91,7 +95,7 @@ else
 fi
 
 # Enable SSL if exists configs
-if [ -e "/ssl/$CERTFILE" ] && [ -e "/ssl/$KEYFILE" ]; then
+if [ -e "/ssl/$CAFILE" ] && [ -e "/ssl/$CERTFILE" ] && [ -e "/ssl/$KEYFILE" ]; then
```
pvizeli: Well, that work only now if you have also a ca file

Diaoul:
I did try that, it didn't work. Maybe I did something wrong? Can you try?

Le dim. 7 juil. 2019 à 12:38, Pascal Vizeli <notifications@github.com> a
écrit :

Diaoul:
It is already the case with cafile being set to certfile currently assuming
certfile contains the ca (fullchain).

Le dim. 7 juil. 2019 à 12:39, Pascal Vizeli <notifications@github.com> a
écrit :

mosquitto/data/run.sh
Diaoul: It is already the case with cafile being set to certfile currently assuming certfile contains the ca (fullchain).

Diaoul:
How rude.

DConcord:
This would be awesome! I hope you get it working Diaoul :)

mosquitto/data/run.sh
DConcord: thats true. cafile is already required, but just hard-coded to point to the same place as certifle. You could always just make the default: "cafile": "certfile.pem" along with "require_certificate": false. This way it will continue to work by default but is configurable if desired.

Diaoul: OK yes I see what you mean, I had several iterations before this so my previous answer is wrong and you are both correct 😄
I'm not sure what to do here:
1. Have self-explained and accurate default config but not backward compatible which makes it easy for new users (I think) and forces old users to update.
2. Have default config backward compatible but more misleading for new users (cafile and certfile both pointing to fullchain.pem, not even sure that actually works)

I think I went for 2 but had second thoughts and switched to 1 😆

Diaoul: To my own defense, I did say in my previous comment that I made that switch 😌 

> [...] All that comes with non-breaking defaults although for people having already configured the SSL it might not be backward compatible. [...]

mosquitto/config.json
```diff
@@ -24,8 +24,10 @@
       "active": false,
       "folder": "mosquitto"
     },
-    "certfile": "fullchain.pem",
-    "keyfile": "privkey.pem"
+    "cafile": "cafile.pem",
```
pvizeli: let it work like the old version. Don't set that option default and if they are not set, use the old hardcoded version.

Diaoul:
After hours on this I think I've got it right.

PR #627, opened 2019-07-12
Added config option to set ip for ipv4 and ipv6

lclc98:
Allows you to set the IP so that it doesn't autodetect each time.

### DIFF
--- a/duckdns/config.json
+++ b/duckdns/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Duck DNS",
-  "version": "1.6",
+  "version": "1.7",
   "slug": "duckdns",
   "description": "Free Dynamic DNS (DynDNS or DDNS) service with Let's Encrypt support",
   "url": "https://www.home-assistant.io/addons/duckdns/",
@@ -14,6 +14,8 @@
       "certfile": "fullchain.pem",
       "keyfile": "privkey.pem"
     },
+    "ipv4": "",
+    "ipv6": "",
     "token": null,
     "domains": [null],
     "seconds": 300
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @lclc98, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#627).

Thanks!

duckdns/config.json
```diff
@@ -14,6 +14,8 @@
       "certfile": "fullchain.pem",
       "keyfile": "privkey.pem"
     },
+    "ipv4": "",
+    "ipv6": "",
```
pvizeli: I don't think they should be as default config around

lclc98: I currently have that as the default because not setting the ip would make it autodetect which follows the same behavior.

pvizeli: Sure, but you can also use jq for default value. I don't want show this settings default on UI. They can be around if user explicit set it.

duckdns/run.sh
```diff
@@ -11,6 +11,8 @@ LE_DOMAINS=$(jq --raw-output '.domains[]' $CONFIG_PATH)
 LE_UPDATE="0"
 
 # DuckDNS
+IPV4=$(jq --raw-output '.ipv4' $CONFIG_PATH)
+IPV6=$(jq --raw-output '.ipv6' $CONFIG_PATH)
```
pvizeli: `$(jq --raw-output '.ipv6 // empty' $CONFIG_PATH)`

PR #649, opened 2019-08-09
update deCONZ

agners:
* add aarch64 version

  * add armv7 version


frenck:
So how would that work?

- Considering the fact this file isn't changed? <https://github.com/home-assistant/hassio-addons/blob/master/deconz/build.json>
- armv7 is added, but what is the use case? Since deCONZ is shipped for armhf. armv7 devices for Hass.io will fall back to armhf...
- The aarch64 is beta... Not sure if we want to ship that, especially since Hass.io is able to fall back to armhf on (most of) those devices

agners:
Sorry I rushed this pull request this morning because I realized @pvizeli is working on the same thing. I should have put WIP into the description

pvizeli:
@agners are fine with this?

agners:
@pvizeli after that last fix looks good to me! Thanks for picking this up!

agners:
@pvizeli I haven't seen aarch64 to become available (although 3.5 is), do I have to do something to refresh container metadata or something?

pvizeli:
It fetches all time the best arch match on every update. So if you click on update, they should see that the new aarch64 match better and take this image

agners:
@pvizeli you are right, checked again on the device, and it indeed installed the right image: `homeassistant/aarch64-addon-deconz`. What confused me is that on the plugin page it still shows that aarch64 | no badge...

Issue #680, opened 2019-08-31
Unable to recognize Conbee 2 USB stick on Intel NUC running CentOS 


ragnarbullnielsen:
deCONZ version 3.3
hass.io 0.98

Hass.io hardware info:

serial:
/dev/ttyS0
/dev/ttyACM0
/dev/ttyS2
/dev/serial/by-id/usb-dresden_elektronik_ingenieurtechnik_GmbH_ConBee_II_DE1963639-if00
/dev/ttyS3
/dev/ttyS1
input:
HDA Intel PCH HDMI/DP,pcm=10
HDA Intel PCH HDMI/DP,pcm=8
HDA Intel PCH HDMI/DP,pcm=3
ITE8708 CIR transceiver
HDA Intel PCH Front Headphone
HDA Intel PCH HDMI/DP,pcm=9
HDA Intel PCH HDMI/DP,pcm=7
PC Speaker
Power Button
MCE IR Keyboard/Mouse (ite-cir)
Sleep Button
Video Bus
disk:
gpio:
audio:

I have tried configuring the add-on using:

{
  "device": "/dev/serial/by-id/usb-dresden_elektronik_ingenieurtechnik_GmbH_ConBee_II_DE1963639-if00",
  "vnc_password": ************"
}

In the Phoscon web app, the gateway shows up as a RaspBee, and 'Not connected'

From the deCONZ log:

20:31:05:726 wait reconnect 2 seconds
20:31:06:726 wait reconnect 1 seconds
20:31:06:743 COM: --dev: /dev/serial/by-id/usb-dresden_elektronik_ingenieurtechnik_GmbH_ConBee_II_DE1963639-if00 (RaspBee)
This plugin does not support propagateSizeHints()
20:31:06:765 Device firmware version 0x26420700
20:31:06:767 unlocked max nodes: 200
20:31:06:841 Device protocol version: 0x0108
20:31:06:848 new node - ext: 0x00212effff0424b0, nwk: 0x0000
20:31:06:896 Current channel 11
20:31:06:912 CTRL ANT_CTRL 0x03
20:40:07:057 Announced to internet

lauer:
Same problem here. It does not work if using the by-id link, however worked fine, if I change back to directly link to e.g /dev/ttyACM2

jamescadd:
I'm having the same issue with a Nortek HUSBZB-1 on .98.1. Only the ttyUSB0/1 devices work.

pvizeli:
[...]

PR #684, opened 2019-09-02
Cleanup some udev rules

pvizeli:
Fix: #680

### DIFF
--- a/deconz/data/run.sh
+++ b/deconz/data/run.sh
@@ -3,7 +3,7 @@ set -e
 
 # Init own udev service
 /lib/systemd/systemd-udevd --daemon
-udevadm trigger --action=change
+bashio::hardware.trigger
 
 # Ensure otau folder exists
 mkdir -p "/data/otau"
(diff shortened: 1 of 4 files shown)


deconz/data/run.sh
```diff
@@ -24,6 +24,7 @@ VNC_PASSWORD=$(bashio::config 'vnc_password')
 WEBSOCKET_PORT=$(bashio::addon.port 8080)
 
 # Lookup udev link
+sleep 3
```
frenck: Isn't it better to add a loop here with a timeout? To check if it is available, and continue immediately is so?

frenck: Or do a while link does not exists counter++, sleep 1; if max counter -> bashio::exit.nok "timeout reached"

pvizeli: If the device does not exist on that point, he will never exist. Because it's docker, the device needs plugin before the add-on start. The sleep is not really needed, only a small helper for slower device.

frenck: Sure! totally understand, my suggestion was to make the sleep conditional using a loop, so faster devices will stall for 3 seconds.

PR #697, opened 2019-09-13
Added an optional Cloudflare mechanism for real visitor ip

Shiwigy:
Added an optional Cloudflare mechanism for real visitor ip.
If enabled it will auto-generate /data/cloudflare.conf with a list of ipv4/ipv6 addresses directly from Cloudflare that will be used for set_real_ip_from.

The /data/cloudflare.conf will be included in the /etc/nginx.conf

This is so the ip_ban_enabled can be used and work correctly in /config/customize.yaml

Example:
http:
        base_url: !secret base_url
        use_x_forwarded_for: true
        trusted_proxies:
                - 127.0.0.1
                - 172.30.33.1
        ip_ban_enabled: true
        login_attempts_threshold: 5

### DIFF
--- a/nginx_proxy/config.json
+++ b/nginx_proxy/config.json
@@ -18,8 +18,7 @@
     "keyfile": "privkey.pem",
     "hsts": "max-age=31536000; includeSubDomains",
     "cloudflare": {
-      "enable": false,
-      "real_ip_header": "CF-Connecting-IP"
+      "enable": false
     },
     "customize": {
       "active": false,
@@ -33,8 +32,7 @@
     "keyfile": "str",
     "hsts": "str",
     "cloudflare": {
-      "enable": "bool",
-      "real_ip_header": "str"
+      "enable": "bool"
     },
     "customize": {
       "active": "bool",
--- a/nginx_proxy/config.json
+++ b/nginx_proxy/config.json
@@ -1,6 +1,6 @@
 {
   "name": "NGINX Home Assistant SSL proxy",
-  "version": "2.3",
+  "version": "2.4",
   "slug": "nginx_proxy",
   "description": "An SSL/TLS proxy",
   "url": "https://home-assistant.io/addons/nginx_proxy/",
--- a/nginx_proxy/config.json
+++ b/nginx_proxy/config.json
@@ -31,9 +31,7 @@
     "certfile": "str",
     "keyfile": "str",
     "hsts": "str",
-    "cloudflare": {
-      "enable": "bool"
-    },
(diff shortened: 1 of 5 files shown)


nginx_proxy/config.json
```diff
@@ -17,6 +17,10 @@
     "certfile": "fullchain.pem",
     "keyfile": "privkey.pem",
     "hsts": "max-age=31536000; includeSubDomains",
+    "cloudflare": {
+      "enable": false,
+      "real_ip_header": "CF-Connecting-IP"
```
pvizeli: Why need that to be a customable value?

Shiwigy: I was thinking that maybe someone had a need to use x-forwarded-for when multiple connecting proxies was in use, but i also see that its a one-time setting for now, since i dont perform any update on the cloudflare.conf after first time enabled.

nginx_proxy/config.json
```diff
@@ -28,6 +31,9 @@
     "certfile": "str",
     "keyfile": "str",
     "hsts": "str",
+    "cloudflare": {
```
pvizeli: I think a simple `cloudflare: bool` should be better

Shiwigy: I agree, simplified the config.json to "cloudflare": "bool" and default "cloudflare": false, updated run.sh to reflect the change and check if "$CLOUDFLARE" == "true"

nginx_proxy/data/run.sh
```diff
@@ -24,6 +27,30 @@ if [ ! -f "$SNAKEOIL_CERT" ]; then
     openssl req -x509 -nodes -days 3650 -newkey rsa:2048 -keyout $SNAKEOIL_KEY -out $SNAKEOIL_CERT -subj '/CN=localhost'
 fi
 
+if [ "$CLOUDFLARE" == "true" ]; then
+    sed -i "s|#include /data/cloudflare.conf;|include /data/cloudflare.conf;|" /etc/nginx.conf
+    # Generate cloudflare.conf
+    if [ ! -f "$CLOUDFLARE_CONF" ]; then
```
pvizeli: At last, this list will be never updated after first creating. Is that okay or do we want update the list of IPs on every start but do not fails if we are not able to update it?

Shiwigy: I checked the list with the list using Wayback Machine, back up to three years ago, and essentially one entry got removed, which means using the list from three years ago has one unnecessary entry, but that’s about it. So i don't see a need to update it on every run.

Issue #576, opened 2019-04-26
MariaDB ability to set different path for storage


escoand:
I would like to use an external path for the MariaDB storage. I use Hass.io on ODROID-HC1 which has an SATA slot and I would like to use SSD in it for the database.

escoand:
I'm still interested in this feature.

danimtb:
I found this interesting too. I would like to store the database in an external usb drive. I think it is just a matter of chaging the datadir in the configuration file https://github.com/home-assistant/hassio-addons/blob/97156f0f093c8ebfaa8fb6f1dec0dbe3506094eb/mariadb/mariadb-server.cnf#L6

danimtb:
Or maybe making the variable `MARIADB_DATA` configurable from options
https://github.com/home-assistant/hassio-addons/blob/97156f0f093c8ebfaa8fb6f1dec0dbe3506094eb/mariadb/run.sh#L5

danimtb:
Didn't have the time to look into this but I am still interested in having support for this

danimtb:
I will be having a look at this soon once I set up everything for my new home assistant instance. Please do not close this issue

frenck:
We have discussed this and decided not to go this route. There are many things that can go wrong if a user misconfigures this, and from our end things can become unpredictable. 

If you need this, we suggest maintaining your own version of the add-on.

danimtb:
@frenck would you suggest any workaround for this?

frenck:
@danimtb Well, the add-on is open source, so you are free to maintain/create a fork of it or create your own version.

danimtb:
I know that alternative, I was asking about a workaround with the officially supported addons and home assistant features. Thanks

frenck:
There is currently not a supported workaround available.

danimtb:
Ok, thanks for the answer. I will take a look to the issue from my side.

PR #706, opened 2019-09-20
Mariadb patch datadir

Shiwigy:
Patch for #576 
Allow setting the datadir to desired path via options


frenck:
We have discussed this and decided not to go this route. There are many things that can go wrong if a user misconfigures this, and from our end things can become unpredictable. 

If you need this, we suggest maintaining your own version of the add-on.

Nevertheless, thank you for taking the time to contribute to the project, that is absolutely appreciated 🎖

Issue #709, opened 2019-09-24
Let's Encrypt 4.0 addon


AlixerVolkov:
Problem with Let's Encrypt 4.0 addon, was working well with 3.0:
Hassio 0.99.2

19-09-24 16:49:47 ERROR (MainThread) [aiohttp.server] Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 418, in start
    resp = await task
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_app.py", line 458, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.7/site-packages/aiohttp/web_middlewares.py", line 119, in impl
    return await handler(request)
  File "/usr/src/hassio/hassio/api/security.py", line 147, in token_validation
    return await handler(request)
  File "/usr/src/hassio/hassio/api/utils.py", line 39, in wrap_api
    answer = await method(api, *args, **kwargs)
  File "/usr/src/hassio/hassio/addons/addon.py", line 473, in start
    await self.write_options()
  File "/usr/src/hassio/hassio/addons/addon.py", line 357, in write_options
    options = schema(options)
  File "/usr/local/lib/python3.7/site-packages/voluptuous/schema_builder.py", line 272, in __call__
    return self._compiled([], data)
  File "/usr/local/lib/python3.7/site-packages/voluptuous/validators.py", line 205, in _run
    return self._exec(self._compiled, value, path)
  File "/usr/local/lib/python3.7/site-packages/voluptuous/validators.py", line 285, in _exec
    v = func(path, v)
  File "/usr/local/lib/python3.7/site-packages/voluptuous/schema_builder.py", line 817, in validate_callable
    return schema(data)
  File "/usr/src/hassio/hassio/addons/validate.py", line 337, in validate
    options[key] = _single_validate(coresys, typ, value, key)
  File "/usr/src/hassio/hassio/addons/validate.py", line 388, in _single_validate
    elif typ.strartswith(V_LIST):
AttributeError: 'str' object has no attribute 'strartswith'

olbjan:
This problem has been resolved with hass.io supervisor 189.

However problems exist around the new dns challenge:
-with a valid existing cert changing the hostname will not request a new cert
-the dns settings as described in the readme don't work yet because it isn't passing along a valid commandline to certbot:

certbot: error: unrecognized arguments: --null --null-credentials /data/dnsapikey

[...]

PR #713, opened 2019-09-25
Fix issue #709 - //empty function not required, as its handled by bas…

Zapfmeister:
…hio already and caused runtime failure


pvizeli:
- [ ] Bump version +0.1
- [ ] update changelog

PR #754, opened 2019-10-03
Fix route53 support

### LINKED ISSUE #715: Let's Encrypt Addon does not work with AWS Route53
The log produced the following error:
```
usage: 
certificate. 
certbot: error: unrecognized arguments: --dns-route53-credentials /data/dnsapikey
```
I will be making a pull request shortly.

Zapfmeister:
closes #715

### DIFF
--- a/letsencrypt/run.sh
+++ b/letsencrypt/run.sh
@@ -13,38 +13,45 @@ WORK_DIR=/data/workdir
 mkdir -p "$WORK_DIR"
 mkdir -p "$CERT_DIR"
 mkdir -p "/ssl"
+mkdir -p ~/.aws/
 chmod +x /run.sh
-touch /data/dnsapikey
 
-echo -e "dns_cloudflare_email = $(bashio::config 'dns.cloudflare_email')\n" \
-  "dns_cloudflare_api_key = $(bashio::config 'dns.cloudflare_api_key')\n" \
-  "dns_cloudxns_api_key = $(bashio::config 'dns.cloudxns_api_key')\n" \
-  "dns_cloudxns_secret_key = $(bashio::config 'dns.cloudxns_secret_key')\n" \
-  "dns_digitalocean_token = $(bashio::config 'dns.digitalocean_token')\n" \
-  "dns_dnsimple_token = $(bashio::config 'dns.dnsimple_token')\n" \
-  "dns_dnsmadeeasy_api_key = $(bashio::config 'dns.dnsmadeeasy_api_key')\n" \
-  "dns_dnsmadeeasy_secret_key = $(bashio::config 'dns.dnsmadeeasy_secret_key')\n" \
-  "dns_gehirn_api_token = $(bashio::config 'dns.gehirn_api_token')\n" \
-  "dns_gehirn_api_secret = $(bashio::config 'dns.gehirn_api_secret')\n" \
-  "dns_linode_key = $(bashio::config 'dns.linode_key')\n" \
-  "dns_linode_version = $(bashio::config 'dns.linode_version')\n" \
-  "dns_luadns_email = $(bashio::config 'dns.luadns_email')\n" \
-  "dns_luadns_token = $(bashio::config 'dns.luadns_token')\n" \
-  "dns_nsone_api_key = $(bashio::config 'dns.nsone_api_key')\n" \
-  "dns_ovh_endpoint = $(bashio::config 'dns.ovh_endpoint')\n" \
-  "dns_ovh_application_key = $(bashio::config 'dns.ovh_application_key')\n" \
-  "dns_ovh_application_secret = $(bashio::config 'dns.ovh_application_secret')\n" \
-  "dns_ovh_consumer_key = $(bashio::config 'dns.ovh_consumer_key')\n" \
-  "dns_rfc2136_server = $(bashio::config 'dns.rfc2136_server')\n" \
-  "dns_rfc2136_port = $(bashio::config 'dns.rfc2136_port')\n" \
-  "dns_rfc2136_name = $(bashio::config 'dns.rfc2136_name')\n" \
-  "dns_rfc2136_secret = $(bashio::config 'dns.rfc2136_secret')\n" \
-  "dns_rfc2136_algorithm = $(bashio::config 'dns.rfc2136_algorithm')\n" \
-  "aws_access_key_id = $(bashio::config 'dns.aws_access_key_id')\n" \
-  "aws_secret_access_key = $(bashio::config 'dns.aws_secret_access_key')\n" \
-  "dns_sakuracloud_api_token = $(bashio::config 'dns.sakuracloud_api_token')\n" \
-  "dns_sakuracloud_api_secret = $(bashio::config 'dns.sakuracloud_api_secret')" > /data/dnsapikey
-chmod 600 /data/dnsapikey
+if bashio::config 'dns.provider' == "dns-route53"; then
+    touch ~/.aws/config
+    echo -e "[default]\n" \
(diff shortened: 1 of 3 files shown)


Zapfmeister:
closes #715

kre8or69:
Would love this to be merged in! Just started using Hassio and already use Route53! haha

letsencrypt/run.sh
```diff
@@ -13,38 +13,45 @@ WORK_DIR=/data/workdir
...
-  "dns_rfc2136_algorithm = $(bashio::config 'dns.rfc2136_algorithm')\n" \
-  "aws_access_key_id = $(bashio::config 'dns.aws_access_key_id')\n" \
-  "aws_secret_access_key = $(bashio::config 'dns.aws_secret_access_key')\n" \
-  "dns_sakuracloud_api_token = $(bashio::config 'dns.sakuracloud_api_token')\n" \
-  "dns_sakuracloud_api_secret = $(bashio::config 'dns.sakuracloud_api_secret')" > /data/dnsapikey
-chmod 600 /data/dnsapikey
+if bashio::config 'dns.provider' == "dns-route53"; then
+    touch ~/.aws/config
+    echo -e "[default]\n" \
+        "aws_access_key_id = $(bashio::config 'dns.aws_access_key_id')\n" \
```
frenck: So do we require `dns.aws_access_key_id` & `dns.aws_secret_access_key` to be set when the `dns.provider` = `dns-route53`?

If so, we need a check for this.

letsencrypt/run.sh
```diff
@@ -13,38 +13,45 @@ WORK_DIR=/data/workdir
...
-  "dns_rfc2136_port = $(bashio::config 'dns.rfc2136_port')\n" \
-  "dns_rfc2136_name = $(bashio::config 'dns.rfc2136_name')\n" \
-  "dns_rfc2136_secret = $(bashio::config 'dns.rfc2136_secret')\n" \
-  "dns_rfc2136_algorithm = $(bashio::config 'dns.rfc2136_algorithm')\n" \
-  "aws_access_key_id = $(bashio::config 'dns.aws_access_key_id')\n" \
-  "aws_secret_access_key = $(bashio::config 'dns.aws_secret_access_key')\n" \
-  "dns_sakuracloud_api_token = $(bashio::config 'dns.sakuracloud_api_token')\n" \
-  "dns_sakuracloud_api_secret = $(bashio::config 'dns.sakuracloud_api_secret')" > /data/dnsapikey
-chmod 600 /data/dnsapikey
+if bashio::config 'dns.provider' == "dns-route53"; then
```
frenck: This logic now adds only these 2 settings, while previous behavior added all configuration of each provider to the configuration.

This this correct behavior? Does this mean only 1 provider can be active at the same time? If that is the case, the logic below (the else) should be adjusted as well to preview major loading of configs (and thus unnecessary calls). If not, well than the else should be removed I guess.

letsencrypt/run.sh
```diff
@@ -13,38 +13,45 @@ WORK_DIR=/data/workdir
 mkdir -p "$WORK_DIR"
 mkdir -p "$CERT_DIR"
 mkdir -p "/ssl"
+mkdir -p ~/.aws/
```
frenck: We can combine those `mkdir -p`'s into a single `mkdir` call.

letsencrypt/run.sh
```diff
@@ -54,8 +61,12 @@ if [ ! -d "$CERT_DIR/live" ]; then
     done
 
     echo "$DOMAINS" > /data/domains.gen
-    if [ "$CHALLENGE" == "dns" ]; then
-        certbot certonly --non-interactive --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" "--$DNS_PROVIDER" "--${DNS_PROVIDER}-credentials" "/data/dnsapikey" --email "$EMAIL" --agree-tos --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}"
+    if bashio::config 'challenge' == "dns"; then
```
frenck: This comparison would not work. Have you tested the add-on?

letsencrypt/run.sh
```diff
@@ -54,8 +61,12 @@ if [ ! -d "$CERT_DIR/live" ]; then
...
+        if bashio::config 'dns_provider' == "dns-route53"; then
```
frenck: same

letsencrypt/run.sh
```diff
@@ -54,8 +61,12 @@ if [ ! -d "$CERT_DIR/live" ]; then
...
+            certbot certonly --non-interactive --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" "--$DNS_PROVIDER" --email "$EMAIL" --agree-tos --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}"
```
frenck: Add some linebreaks to make this more readable.

letsencrypt/run.sh
```diff
@@ -54,8 +61,12 @@ if [ ! -d "$CERT_DIR/live" ]; then
...
+        else
+            certbot certonly --non-interactive --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" "--$DNS_PROVIDER" "--${DNS_PROVIDER}-credentials" "/data/dnsapikey" --email "$EMAIL" --agree-tos --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}"
```
frenck: Add some linebreaks to make this more readable.

letsencrypt/run.sh
Zapfmeister: Yes, those are the required variables for route53

letsencrypt/run.sh
Zapfmeister: Your right. Thanks for the hint. It is now fixed

letsencrypt/run.sh
Zapfmeister: Yes, but it slipped my eyes, as it was always true and i just tested against dns.
Fixed.

letsencrypt/run.sh
Zapfmeister: Fixed

letsencrypt/run.sh
Zapfmeister: Fixed

letsencrypt/run.sh
Zapfmeister: Fixed

Zapfmeister:
@frenck could you please have a look? It should be fine now for merging

tenacioustechie:
Any chance this can be updated and merged @Zapfmeister  and @frenck ?
Just hit this same issue on route53, let me know if there is anything I can do.

pvizeli:
Is fixed

BarrettLowe:
I'm still having this problem. Is the fix released yet?

imduffy15:
This does not appear to be fixed, https://github.com/home-assistant/hassio-addons/commit/50c295eee71cda83c42820e7e2b8ffb56df05aa1#diff-4d929de130b59358dd27cea01a3cff10 introduces a workaround for it, however, the --dns-route53-credentials argument is still passed which is not valid.

I've copied the code and made my own repo to work around it for now, fix is available here https://github.com/imduffy15/hassio-addons/commit/2bade84490ae8cb9469c1b25e9a937789e787d23 or if you want it as a hass repo (don't rely on it long term) https://github.com/imduffy15/hassio-addons

borger:
It doesn't appear to be fixed yet... :(

PR #759, opened 2019-10-07
Delay shutting down the PC

stefanroelofs:
Added 2 new options:
- New option to delay shutting down the Windows PC
- New option to send a message to the PC that is about to be shut down. If somebody is using the PC, you can warn them to save their work.

I tested the command locally, and the -t and -C arguments work as expected.

I've never editted a .bash file before, so this needs to be checked!

new variables are marked with a ? This should mean that they are optional (I don't want to break anyone's configuration.

### DIFF
--- a/rpc_shutdown/config.json
+++ b/rpc_shutdown/config.json
@@ -1,6 +1,6 @@
 {
   "name": "RPC Shutdown",
-  "version": "1.0",
+  "version": "2.0",
   "slug": "rpc_shutdown",
   "description": "Simple way for remote windows shutdowns",
   "url": "https://home-assistant.io/addons/rpc_shutdown/",
@@ -14,7 +14,9 @@
       {
         "alias": "test-pc",
         "address": "192.168.0.1",
-        "credentials": "user%password"
+        "credentials": "user%password",
+        "delay": "0",
+        "message": "Home Assistant is shutting down this PC. This cannot be canceled. Please save your work!"
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @stefanroelofs, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#759).

Thanks!

frenck:
I've requested a review from @pvizeli, I do not own any Windows machines myself, so that is hard to review.

rpc_shutdown/config.json
```diff
@@ -14,7 +14,9 @@
       {
         "alias": "test-pc",
         "address": "192.168.0.1",
-        "credentials": "user%password"
+        "credentials": "user%password",
+        "delay": "0",
```
pvizeli: `"delay": 0`

rpc_shutdown/run.sh
```diff
@@ -16,14 +16,21 @@ while read -r input; do
...
       
         # Not the correct alias
         if [ "$ALIAS" != "$input" ]; then
             continue
         fi
       
+	  # Check if delay is not empty
+        if [ "$DELAY" = "" ]; then
+            DELAY="0"
+        fi
```
pvizeli: Not needed if we have an default

stefanroelofs:
I committed the requested changes.

I could also have a look at converting the script to Bashio? But maybe its better if this is a separate pull request.

pvizeli:
@stefanroelofs would be nice, yeah

pvizeli:
Arr, you are right, it needs a default. I was wrong. Sorry. Can you fix that again @stefanroelofs ? Thanks

stefanroelofs:
no problem. This pull request is closed though. can you re-open? or do i start a new one?

pvizeli:
You need start with a fresh PR/Branch

PR #767, opened 2019-10-08
Adding config option for bash_profile in SSH

jfette:


homeassistant:
Hi @jfette, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#767).

Thanks!

frenck:
Since this add-on does not provide its own bash profile or rc files, isn't it wise to simply make those persistent instead of introducing a configuration option?

jfette:
It could be done that way.  I considered this method as we are appending the token each time run.sh is executed.  With a persistent file, the token could be appended multiple times.

frenck:
I'm not sure how the bash profile relates to those tokens?

jfette:
`echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" >> /root/.bash_profile`

This has existed for some time.  A .bash_profile is created with only this token.  Users do not have the ability to modify or keep any changes stored here.

I could look into creating a persistent file and add/replace the token to this file as needed.

frenck:
We could add something like this to the `.bash_profile` and make `.bashrc` persist?

```
if [[ -o interactive ]] && [[ -f ~/.bashrc ]]; then
  source ~/.bashrc
fi
```

pvizeli:
I agree we should make it persistent and we can use sed to update the token. If the file does not exist, we create a new file with token and link it otherwise we update the token and link it

jfette:
I have updated my PR to incorporate a persistent file.  We replace the token every time run.sh is ran if it exists, otherwise we append the token to the end of the file.

ssh/data/run.sh
```diff
@@ -47,8 +47,18 @@ touch /data/.bash_history
 chmod 600 /data/.bash_history
 ln -s -f /data/.bash_history /root/.bash_history
 
+# Persist .bash_profile by redirecting .bash_profile to /data
+touch /data/.bash_profile
+chmod 600 /data/.bash_profile
+ln -s -f /data/.bash_profile /root/.bash_profile
+
```
pvizeli: More like:
```
if bashio::file.exists /data/.bash_profile; then
   sed -i ... /data/.bash_profile
else
  echo "export ..." > /data/.bash_profile
  chmod 600 /data/.bash_profile
fi
ln -s -f /data/.bash_profile /root/.bash_profile
```

jfette: I've updated the PR to include binary logic on if the file exists to either create or replace the token if necessary.

ssh/data/run.sh
```diff
@@ -47,8 +47,14 @@ touch /data/.bash_history
...
 ln -s -f /data/.bash_history /root/.bash_history
 
-# Store token for Hass.io API
-echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" >> /root/.bash_profile
+# Persist .bash_profile by redirecting .bash_profile to /data
+if bashio::fs.file_exists /data/.bash_profile; then
+  sed -i "s/export HASSIO_TOKEN=.*/export HASSIO_TOKEN=${HASSIO_TOKEN}/" /data/.bash_profile
+else
+  echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" > /root/.bash_profile
+  chmod 600 /data/.bash_profile
```
frenck: Let's move this out of the if statement.
That way, we can guarantee it is always correct. For example, when a user changes it or on the restore of a backup.

frenck:
Agree with @pvizeli, looks fine. The file permissions does not need to be part of the if statement. It might even be beneficial to move it out of there.

ssh/data/run.sh
pvizeli: Good point

ssh/data/run.sh
```diff
@@ -47,8 +47,15 @@ touch /data/.bash_history
 chmod 600 /data/.bash_history
 ln -s -f /data/.bash_history /root/.bash_history
 
-# Store token for Hass.io API
-echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" >> /root/.bash_profile
+# Persist .bash_profile by redirecting .bash_profile to /data
+if bashio::fs.file_exists /data/.bash_profile; then
+  sed -i "s/export HASSIO_TOKEN=.*/export HASSIO_TOKEN=${HASSIO_TOKEN}/" /data/.bash_profile
+else
+  echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" > /root/.bash_profile
```
frenck: I just noticed this!

```suggestion
  echo "export HASSIO_TOKEN=${HASSIO_TOKEN}" > /data/.bash_profile
```

Or else the file is never created!

jfette: Good catch.  Corrected.

PR #786, opened 2019-10-15
Update README.md

AdighaLogic:
Grammar correction.


frenck:
This change is absolutely not needed. Both the original and suggested are correct.

PR #792, opened 2019-10-16
Delay shutting down the PC

stefanroelofs:
Add a check if delay is empty.


pvizeli:
please update version

stefanroelofs:
I updated the addon to use bashio. 
It seems that the checks are failing, but I don't know why..
I'm not too sure about the function call in run.sh line 19. Do I call the var.equals method with 2 parameters like this? And the ! negates the outcome?

rpc_shutdown/run.sh
```diff
@@ -1,37 +1,33 @@
...
+    for computer in $(bashio::config 'computers|keys'); do
+	    ALIAS=$(bashio::config "computers[${computer}].alias")
+	    ADDRESS=$(bashio::config "computers[${computer}].address")
+	    CREDENTIALS=$(bashio::config "computers[${computer}].credentials")
+	    DELAY=$(bashio::config "computers[${computer}].delay")
+	    MESSAGE=$(bashio::config "computers[${computer}].message")
+  
         # Not the correct alias
-        if [ "$ALIAS" != "$input" ]; then
+		if [! bashio::var.equals "$ALIAS" "$input" ]; then
```
pvizeli: Formatierung

rpc_shutdown/run.sh
```diff
@@ -1,37 +1,33 @@
...
             continue
         fi
       
-		# Check if delay is not empty 
-        if [ "$DELAY" = "" ]; then
+		# Check if delay is not empty
+		if bashio::var.is_empty "${DELAY}"; then
```
pvizeli: Formatierung

stefanroelofs:
Found the syntax error. 
Also converted some tabs to spaces.

rpc_shutdown/run.sh
stefanroelofs: You mean the tabs right? Fixed that.

pvizeli:
Thanks

PR #807, opened 2019-10-28
Almond

pvizeli:


almond/Dockerfile
```diff
@@ -0,0 +1,45 @@
+ARG BUILD_FROM
+FROM ${BUILD_FROM}
+
+# Install Alomond requirements
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        git \
```
frenck: `git` could be moved to the second RUN command, that will shave off a couple of MB's from the resulting image (since it can be removed at the end).

almond/data/run.sh
```diff
@@ -0,0 +1,64 @@
...
+        port "3001" \
+)
+
+almond_config=$(\
+    bashio::var.json \
+        kind "io.home-assistant" \
+        hassUrl "http://hassio/homeassistant" \
+        accessToken "${HASSIO_TOKEN}" \
+        refreshToken "${HASSIO_TOKEN}" \
+        accessTokenExpires "$((TODAY + 62208000))" \
```
balloob: How far ahead in the future is this? 

This value is in ms. 

This is what I have in the integration

```python
                    # 5 years from now in ms.
                    "accessTokenExpires": time.time() + (60 * 60 * 24 * 365 * 5) * 1000,
```

almond/data/run.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bashio
+WAIT_PIDS=()
+TOKEN_VALID="$(python3 -c "import time; print(f'{time.time() + (60 * 60 * 24 * 365 * 5) * 1000}')")"
```
balloob:
```suggestion
TOKEN_VALID="$(python3 -c "import time; print(time.time() + (60 * 60 * 24 * 365 * 5) * 1000)"
```

PR #812, opened 2019-11-02
Remove unsupported architectures from Homematic README

TravelinMax:
You might want to double check that I am right about these architectures not being supported.

### DIFF
--- a/homematic/README.md
+++ b/homematic/README.md
@@ -165,9 +165,9 @@ You have several options to get them answered:
 
 In case you've found an bug, please [open an issue on our GitHub][issue].
 
-[aarch64-shield]: https://img.shields.io/badge/aarch64-yes-green.svg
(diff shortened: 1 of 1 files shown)


homematic/README.md
```diff
@@ -2,7 +2,7 @@
 
 HomeMatic central based on OCCU.
 
-![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield] ![Supports armhf Architecture][armhf-shield] ![Supports armv7 Architecture][armv7-shield] ![Supports i386 Architecture][i386-shield]
```
frenck: We should not remove them, instead, leave them and mark them `red` and `no` in the icons listed in the bottom off the file.

TravelinMax:
Fixed. 👍

PR #827, opened 2019-11-14
Adding additional authentication via client certificate

Zapfmeister:
Hi,

i added the optional function to generate an certificate authority, server certificate and client certificate, in order to use an additional authentication via client certificate, for additional authentication security.


frenck:
Thanks for the PR @Zapfmeister!

We've discussed this and decided not to add this feature.
Reasoning it that this adds a lot of advanced features, that a regular user doesn't need, and might confuse them.

One can still use client certificates if one wants to, by providing a custom nginx configuration to the add-on.

Nevertheless, thank you for your contribution 👍

PR #829, opened 2019-11-17
Additional audioServer topics created by Snips platform

Romkabouter:
Snips has added extra topics to provide streaming of bytes.
See the reference here: https://docs.snips.ai/reference/hermes#streaming-a-sound

Those topics should also be out in the bridging, so that other devices are able to respond properly

### DIFF
--- a/snips/CHANGELOG.md
+++ b/snips/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 6.1
+
+- set hermes/audioServer/+/playBytesStreaming/# to allow subscribing from other systems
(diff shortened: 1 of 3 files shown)


pvizeli:
Can you bump the version, update changelog?

Romkabouter:
Yes sure, I was unsure if that was my responsibility or not.
Will do

Romkabouter:
Hi frenck, I could see the update in Hassio, but there is no docker image yet so I got a manifest error.
I don't know how to proceed on this, is such a thing automagically fixed?

frenck:
I'm currently in the process of booting up the machines to check what is going on. Normally it gets published automatically, unfortunately, the build failed.

PR #841, opened 2019-11-19
Bump deCONZ to 2.05.71

allanpersson:


allanpersson:
Merged in https://github.com/home-assistant/hassio-addons/pull/843

PR #891, opened 2019-12-17
Small improvements to the Almond add-on

frenck:
Question: Should we version bump for this?


pvizeli:
Since there are no real changes, I don't think we need to rebuild it.

PR #892, opened 2019-12-17
Implement auto-detecting IPv4/IPv6 addresses for DuckDNS

Wouter0100:
By implementing this on the client side, it allows us to use IPv6 instead of relying on DuckDNS to do the detection for us - which is IPv6 or IPv4 only. This because they're unable to detect both addresses in a single request.

I wasn't unfortunately able to test this on my local HassIO. Is there any documentation available to test this? Because when restarting the container, it is of course restored to the default.

### DIFF
--- a/duckdns/README.md
+++ b/duckdns/README.md
@@ -75,6 +75,14 @@ A list of DuckDNS subdomains registered under your account. An acceptable naming
 
 The number of seconds to wait before updating DuckDNS subdomains and renewing Let's Encrypt certificates.
 
+### Option: `ipv4`
+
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @Wouter0100, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#892).

Thanks!

pvizeli:
Maybe you want to implement my suggestion: https://github.com/home-assistant/hassio-addons/pull/1019

scop:
My (IMO improved) take in #1077

Wouter0100:
#1077 seems a better and more up-to-date version of this, indeed. I'm not using this anymore and thus closing it. Thanks for the support thus far!

PR #927, opened 2019-12-20
Hm regahss normal

pvizeli:


marcelveldt:
@pvizeli this change seemed to have caused an issue on x86/64 machines, the webinterface isn't loading  anymore and this error in the logs: 

/opt/hm/bin/tclsh: /lib/i386-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /opt/hm/lib/libtcl8.2.so)

pvizeli:
@marcelveldt can you test if they work if you overwrite the `tclsh` with old one?

marcelveldt:
> @marcelveldt can you test if they work if you overwrite the `tclsh` with old one?

Sorry for the late reply!
Issue lies is outdated libc6 version in the container. Seems the binaries have been build with a fairly new linux version, using glibc 2.29. Debian buster is still at 2.28, debian bullseye is using 2.29

pvizeli:
Yep, therefore the question is they run with the old tclsh

PR #954, opened 2019-12-29
ssh: Add web terminal, alpine 3.11, improved token handling, bump 8.0

frenck:
## 8.0

- Add support for a web-based terminal via Ingress
- Upgrade Alpine Linux to 3.11
- Improve Hass.io API token handling
- Persist .ssh folder across restarts
- Add helper symlink folders to user home folder


ssh/config.json
```diff
@@ -1,11 +1,15 @@
...
   "name": "SSH server",
-  "version": "7.1",
+  "version": "8.0",
   "slug": "ssh",
   "description": "Allows connections over SSH",
   "url": "https://github.com/home-assistant/hassio-addons/tree/master/ssh",
   "arch": ["armhf", "armv7", "aarch64", "amd64", "i386"],
   "startup": "services",
+  "ingress": true,
+  "ingress_port": 0,
```
pvizeli: We do not run on host network, we can use the default ingress port

frenck: Yep, copy/paste error simply forgot about it. Good catch!

frenck: Fixed.

PR #972, opened 2020-01-03
nginx_proxy: Add Option to disable/enable listening of IPv6

blacktirion:
Even with proper configuration of "trusted_proxies", some users will see an IPv6 translated address on NGINX and Home Assistant IP Ban/Notification. This address is simply an IPv4 that has been translated, however this can cause issues with trusted_networks as well. 

The fix in this pull should add the ability to tell the NGINX instance to listen on IPv6 or to disable listening. The change is done via run.sh, with sed replacing all instances of "[::]:". However this is fully reversible by setting the flag to true if desired.

This issue has been seen here: https://github.com/home-assistant/hassio-addons/issues/471

### DIFF
--- a/nginx_proxy/README.md
+++ b/nginx_proxy/README.md
@@ -49,7 +49,8 @@ Add-on configuration:
     "default": "nginx_proxy_default*.conf",
     "servers": "nginx_proxy/*.conf"
   },
-  "cloudflare": false
+  "cloudflare": false,
+  "listen_ipv6": true
 }
 ```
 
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @blacktirion, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#972).

Thanks!

pvizeli:
The question will be, should be that the default setting? @frenck you are better on Nginx as I, what are you IMHO?

pvizeli:
Do you have to test this? Because it's an internal Add-on and docker IPv6 support is mostly not that was people think about it, I'm not sure if this work

frenck:
IMHO, we should disable IPv6.
In the past I made add-ons configurable for this, in the end, I've removed it. 

Main reason for this, is that it indeed does not work as people expect.

blacktirion:
Either way will work. I changed it this way to make it optional to the user, as I don't like removing choice, but I am not terribly familiar with Docker's network stack, so if it is better to just change it hard coded, I can submit another PR.

pvizeli:
I agree with @frenck. The host stack run on IPv6 and we map the port from docker to the host stack. The Internal Network don't need have ipv6 because it's a virtual network for internal communication.

blacktirion:
Thanks @frenck and @pvizeli 

I've opened this up as a PR https://github.com/home-assistant/hassio-addons/pull/991. Should just be removing the "[::]:" sections of nginx.conf

Issue #987, opened 2020-01-09
minor correction needed in main Readme.MD


prabbit237:
The link for Lets Encrypt in the main readme at https://github.com/home-assistant/hassio-addons/blob/master/README.md points to the wrong page:

```
- **[Let's Encrypt](/duckdns/README.md)**

    Manage an create certificates from Let's Encrypt.
```

Issue #956, opened 2019-12-29
Fix Route53 support in Let's Encrypt 


pickerin:
Re-opening https://github.com/home-assistant/hassio-addons/issues/754 issue as it appears to not be fixed.

After configuration, the add-on is passing "--dns-route53-credentials /data/dnsapikey" to certbot which is an invalid option.

@imduffy15 copied the addon code and created a complete fix here:
https://github.com/imduffy15/hassio-addons/tree/master/letsencrypt

This is with Home Assistant 0.103.5.

aapit:
I can confirm these exact symptoms on HA `0.104.0`, in my case on `hass.io`.

Zapfmeister:
Please check if it works with the latest version and provide feedback

gitolicious:
I thought the same, but digging into the code I found out it's just a case of insufficient documentation combined with confusing log output.

https://github.com/home-assistant/hassio-addons/blob/a5569342fb46166e3c68c5880193e94847e12e6b/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L29-L35

So the check isn't actually on your `dns.provider` to be _dns-route53_ but on the existence of AWS credentials. Add the following and you should be good to go:

```
dns:
  provider: dns-route53
  aws_access_key_id: YOURKEYHERE
  aws_secret_access_key: YOURSECRETHERE
```

pmgledhill102:
And a note for anyone reading this. You don't include the `dns:` parent in the UI. You just put the 3 parameters
```yaml
provider: dns-route53
aws_access_key_id: YOURKEYHERE
aws_secret_access_key: YOURSECRETHERE
```

lekkerdriveway:
Note, the secret and access key params are case sensitive. AWS_SECRET_ACCESS_KEY will not work.

Issue #881, opened 2019-12-11
letsencrypt poor/bad documentation for config


prabbit237:
In the page for letsencrypt, it shows:

**### Configuration**
Add-on configuration:
````
{
  "email": "hello@home-assistant.io",
  "domains": [
    "home-assistant.io"
  ],
  "challenge": "dns",
  "dns": {
    "provider": "dns-cloudflare",
    "cloudflare_email": "cf@home-assistant.io",
    "cloudflare_api_key": "31242lk3j4ljlfdwsjf0"
  }
}
````

The default configuration settings are:

````
{
  "email": null,
  "domains": [
    null
  ],
  "certfile": "fullchain.pem",
  "keyfile": "privkey.pem",
  "challenge": "http",
  "dns": {}
}
````

On either of these, if you click **save** it gives an error reading something along the lines of:

**not a valid value for dictionary value @ data['options']. Got {'email': None, 'domains': [None], 'certfile': 'fullchain.pem', 'keyfile': 'privkey.pem', 'challenge': 'http', 'dns': {}}**

This error message is very **nonusefull** in that doesn't tell you what part of it is causing the problem. Now the error message may be generated by some library provided by someone else, thus making it very difficult to change but if the examples/defaults are changed to be syntactically valid, it would at least make it easier to figure out what's wrong when the user changes those defaults to the user's information.

The example configuration needs the two lines giving the cert and key filenames (and it took me quite some time to realize that.)

````
{
  "email": "hello@home-assistant.io",
  "domains": [
    "home-assistant.io"
  ],
  "challenge": "dns",
  "dns": {
    "provider": "dns-cloudflare",
    "cloudflare_email": "cf@home-assistant.io",
    "cloudflare_api_key": "31242lk3j4ljlfdwsjf0"
  },
  "certfile": "fullchain.pem",
  "keyfile": "privkey.pem"
}
````

The default settings should show an example of settings that **will* save error-free (but in such a way that it's obvious what needs to be changed in the default and what can be left as-is.)
````
{
  "email": "changeme@yourdomain.tld",
  "domains": [
    "www.yourdomain.tld","yourdomain.tld","subdomain.yourdomain.tld","etc.yourdomain.tld"
  ],
  "certfile": "fullchain.pem",
  "keyfile": "privkey.pem",
  "challenge": "http",
  "dns": {}
}

````

cwiggs:
Just wanted to mention I was having the same issue and your post helped me get the plugin working, thanks!

[...]

PR #988, opened 2020-01-09
Fix route53, add google dns, fix documentation

### LINKED ISSUE #977: Config issue with Let's Encrypt add-on
I'm trying to connect in https to my Hassio using the add-on. The Config results in an error message: "not a valid value for dictionary value at data  ['options']....
I have a public IP address so I don't need to use a DNS service.
Where is my mistake? the error messag doesn't allow to understand where the issue is.
thank you

Zapfmeister:
closes #977 #956 #881 #987 #834


letsencrypt/config.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Let's Encrypt",
-  "version": "4.2",
+  "version": "4.4",
```
pvizeli: 4.3?

letsencrypt/config.json
```diff
@@ -15,8 +15,8 @@
   },
   "map": ["ssl:rw"],
   "options": {
-    "email": null,
-    "domains": [null],
+    "email": "hello@home-assistant.io",
+    "domains": ["home-assistant.io"],
```
pvizeli: Let that on `null`, otherwise people hit the start button without change it

letsencrypt/data/run.sh
```diff
@@ -7,6 +7,17 @@ CERTFILE=$(bashio::config 'certfile')
...
+if [[ "$CHALLENGE" == "dns" ]]; then
+  bashio::log.info "Selected DNS Provider: $(bashio::config 'dns.provider')"
+else
+  bashio::log.info "Selected http verification"
+fi
+
+if [[ "$EMAIL" == "hello@home-assistant.io" ]] || [[ "$DOMAINS" == "home-assistant.io" ]]; then
+  bashio::log.error "Please update the Email and Domain section with your details"
+  exit
+fi
```
pvizeli: Is not needed with the comment above

letsencrypt/data/run.sh
```diff
@@ -46,13 +58,23 @@ echo -e "dns_cloudflare_email = $(bashio::config 'dns.cloudflare_email')\n" \
...
 
     export AWS_ACCESS_KEY_ID
     export AWS_SECRET_ACCESS_KEY
+    PROVIDER_ARGUMENTS="--${DNS_PROVIDER}"
+#Google
+elif bashio::config.exists 'dns.google_creds'; then
+    GOOGLE_CREDS="$(bashio::config 'dns.google_creds')"
+
+    export GOOGLE_CREDS
+    PROVIDER_ARGUMENTS=("--${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /ssl/"$GOOGLE_CREDS")
```
pvizeli: I think the share folder is a better place? We can also move it from share into `/data` on first start to make sure that it will work later if the user cleanup the share.

letsencrypt/config.json
Zapfmeister: This was done because of the discussion in #881. Its correct that the error message, when using null is hard to understand.
Therefore using this approach would provide the user with an easy understandable error message, making the setup easier.

letsencrypt/data/run.sh
Zapfmeister: This was done because of the discussion in #881. Its correct that the error message, when using null is hard to understand.
Therefore using this approach would provide the user with an easy understandable error message, making the setup easier.

letsencrypt/config.json
frenck: It also causes unexpected behaviour, which is weird as well.

My suggestion would be: Make it an empty string, detect that empty string in the startup and make a nice proper error message with instructions.

@pvizeli What do you think?

Zapfmeister: I see what you mean. But the error im refering too is appearing when you are trying to save the config, not when you start the container. Maybe a change of the config.json is needed to allow saving the null setting?

frenck:
No, it should not allow saving `null`.

JakeHertenstein:
Isn't it "Route 53" - not "route54"?

letsencrypt/config.json
pvizeli: The issue is, if we migrate to UI options, the null is an indicator to become a `*`

letsencrypt/config.json
Zapfmeister: Yes, changelog included 4.3 already. seems this was missed to be updated in this file before

letsencrypt/data/run.sh
Zapfmeister: Updated as requested.

Zapfmeister:
Updated as requested

borpin:
It would be useful if this instruction was a little more verbose or linked to more information.

> Please copy your the credentials file "google.json" into the "share" shared folder on the hass.io host before starting the service

The first time I needed to access the "share" folder, I had to resort to Discord to discover I'd find it at `/usr/share/hassio` on a generic install.

Would the folder required be `/usr/share/hassio/share`?

Zapfmeister:
@borpin the easiest way is to enable the samba addon and upload it there.

borpin:
> @borpin the easiest way is to enable the samba addon and upload it there.

I agree it is the easiest, but that nugget of information is so hard to find and there are occasions when you need to do it via SSH.

Zapfmeister:
@borpin i cannot do much more then putting it in the documentation. It actually says to use the samba addon.
But i will add the local folder for users that are not using that addon.
Yes, /usr/share/hassio/share is correct.

letsencrypt/data/run.sh
```diff
@@ -15,6 +21,7 @@ mkdir -p "$CERT_DIR"
 mkdir -p "/ssl"
 chmod +x /run.sh
 touch /data/dnsapikey
+PROVIDER_ARGUMENTS=""
```
pvizeli: `PROVIDER_ARGUMENTS=()`

letsencrypt/data/run.sh
```diff
@@ -46,13 +53,29 @@ echo -e "dns_cloudflare_email = $(bashio::config 'dns.cloudflare_email')\n" \
...
 
-# AWS workaround
+# AWS
 if bashio::config.exists 'dns.aws_access_key_id' && bashio::config.exists 'dns.aws_secret_access_key'; then
     AWS_ACCESS_KEY_ID="$(bashio::config 'dns.aws_access_key_id')"
     AWS_SECRET_ACCESS_KEY="$(bashio::config 'dns.aws_secret_access_key')"
 
     export AWS_ACCESS_KEY_ID
     export AWS_SECRET_ACCESS_KEY
+    PROVIDER_ARGUMENTS="--${DNS_PROVIDER}"
```
pvizeli: PROVIDER_ARGUMENTS+=("--${DNS_PROVIDER}")

letsencrypt/data/run.sh
```diff
@@ -46,13 +53,29 @@ echo -e "dns_cloudflare_email = $(bashio::config 'dns.cloudflare_email')\n" \
...
+    GOOGLE_CREDS="$(bashio::config 'dns.google_creds')"
+
+    export GOOGLE_CREDS
+    if [ -f /share/"$GOOGLE_CREDS" ]; then
+      cp -f /share/"$GOOGLE_CREDS" /data/"$GOOGLE_CREDS"
+      chmod 600 /data/"$GOOGLE_CREDS"
+    else
+      bashio::log.info "Google Credentials File doesnt exists in folder share."
+    fi
+    PROVIDER_ARGUMENTS=("--${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /data/"$GOOGLE_CREDS")
```
pvizeli: `+=`

letsencrypt/data/run.sh
```diff
@@ -46,13 +53,29 @@ echo -e "dns_cloudflare_email = $(bashio::config 'dns.cloudflare_email')\n" \
...
+#All others
+else
+    PROVIDER_ARGUMENTS=("--${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /data/dnsapikey)
```
pvizeli: `+=`

letsencrypt/data/run.sh
Zapfmeister: Updated as requested

letsencrypt/data/run.sh
Zapfmeister: Updated as requested

letsencrypt/data/run.sh
Zapfmeister: Updated as requested

letsencrypt/data/run.sh
Zapfmeister: Updated as requested

letsencrypt/data/run.sh
fantapop: Hey @pvizeli, random bystander here.  What do the parenthesis do in this case?  I was unfamiliar with this usage of bash.  It looks like it executes the contents in a subshell.

PR #991, opened 2020-01-11
nginx_proxy: Remove IPv6 Listen statements from nginx.conf

blacktirion:
Referencing PR: https://github.com/home-assistant/hassio-addons/pull/972, making this change hard-coded, rather than an option to be set in the Addon Config.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2.6
+
+- Remove ipv6 listener because we run only inside virtual network on a ipv4 range
+
+## 2.5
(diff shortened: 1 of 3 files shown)


frenck:
The `ipv6only=off` is no longer needed. Also, the resolver needs to be adjusted to turn off IPv6.

blacktirion:
@frenck I changed the ipv6only=off, however I am not sure what you mean about the resolver. Are you speaking about resolver in the location config or DNS resolver for the container?

Thanks for your time.

frenck:
Oh nevermind, apparently, we don't use the resolver Nginx configuration in this add-on. Sorry about that.

PR #1035, opened 2020-02-02
devcontainer_example

issacg:
Demo/example add-on utilizing devcontainers.

This allows add-on developer to set up local supervisor+homeassistant environment, placing the add-on under development in the "local addons"


issacg:
Updated in #1040

PR #1040, opened 2020-02-03
Add devcontainer

issacg:
Added devcontainer based loosely on my work at https://github.com/issacg/hassio-addon-devcontainer but optimized for working with the full community add-on repository

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -30,10 +30,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 COPY .devcontainer/start_hassio.sh /usr/local/bin/start_hassio.sh
 
-# Install dependencies for the add-on development below.  For example, if you're running Node.js,
-# you may want something like the following...
-# RUN apt-get install -y --no-install-recommends nodejs npm
-
 # Generate a machine-id for this container
 RUN rm /etc/machine-id && dbus-uuidgen --ensure=/etc/machine-id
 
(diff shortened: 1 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+        docker-ce \
+        docker-ce-cli \
+        containerd.io
+# This is a development container.  Don't bother to clean up apt cache, this way we have it handy later
+
+COPY .devcontainer/start_hassio.sh /usr/local/bin/start_hassio.sh
+
+# Install dependencies for the add-on development below.  For example, if you're running Node.js,
+# you may want something like the following...
+# RUN apt-get install -y --no-install-recommends nodejs npm
```
pvizeli: Remove this

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+        sudo \
+    && curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add - \
+    && add-apt-repository "deb https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" \
+    && apt-get update && apt-get install -y --no-install-recommends \
+        docker-ce \
+        docker-ce-cli \
+        containerd.io
+# This is a development container.  Don't bother to clean up apt cache, this way we have it handy later
+
+COPY .devcontainer/start_hassio.sh /usr/local/bin/start_hassio.sh
```
pvizeli: We don't need copy this, let put the script into `.devcontainer`

issacg: Agreed that this makes more sense in the generic version of the devcontainer (where users might want to change paths), but I guess in this case it can work.  I guess some pedantic part of me objects to putting a "binary" (even if it is a script) in such an oddball place...

Curious as to motives for the pednatic-ness on your end that wants to avoid the COPY, though.  Would be happy if you shared your motive.

pvizeli:
Let's test it :)

PR #1067, opened 2020-02-11
Rewrite MariaDB add-on

frenck:
## 2.0

- Pin add-on to Alpine Linux 3.11
- Redirect MariaDB error log to add-on logs (less writes to disk)
- Remove grant & host options
- Add support for the mysql service
- Use a more secure default on install
- Skip DNS name resolving
- Improve integrity checks and recovery
- Tune MariaDB for lower memory usage

ℹ️ The tuning is done, because in theory, MariaDB was allowed to use 7.2Gb of memory... This should limit it around max 150-200Mb, while still performing nicely. We might need to tune this more in the future, depends a bit on actual production experience.

ℹ️ The bin log has been enabled again, to improve recovery functionality (at the cost of more disk writes).

**⚠️ This is a big change, we need to test this!**

The output of logs:

image

### DIFF
--- a/mariadb/data/run.sh
+++ b/mariadb/data/run.sh
@@ -19,7 +19,7 @@ ln -s /proc/1/fd/1 /data/databases/local-mariadb.err
 
 # Start mariadb
 bashio::log.info "Starting MariaDB"
-mysqld_safe --datadir="$MARIADB_DATA" --user=root --skip-log-bin < /dev/null &
+mysqld_safe --datadir="$MARIADB_DATA" --user=root < /dev/null &
 MARIADB_PID=$!
 
 # Wait until DB is running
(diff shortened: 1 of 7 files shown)


mariadb/data/run.sh
```diff
@@ -2,18 +2,24 @@
...
-    mysql_install_db --user=root --datadir="$MARIADB_DATA" > /dev/null
+    mysql_install_db --user=root --datadir="${MARIADB_DATA}" --skip-name-resolve --skip-test-db > /dev/null
+    NEW_INSTALL=true
 else
     bashio::log.info "Using existing mariadb initial system"
 fi
 
+# Redirect log output
+rm -f /data/databases/local-mariadb.err
+ln -s /proc/1/fd/1 /data/databases/local-mariadb.err
```
frenck: This is not good for production I now realize, we might need to fix this in the mysql cnf by making it more predictable using the `log_error` directive.

frenck: Adjusted.

pvizeli:
Fix: #667

PR #1082, opened 2020-02-16
Update nginx.conf

jumping2000:
My propose is to use only TLS v1,2 and 1,3 with cipher suite recommended by https://wiki.mozilla.org/Security/Server_Side_TLS.
In march 2020 many web browsers  will drop support for the encryption protocols TLS 1.0 and TLS 1.1.


homeassistant:
Hi @jumping2000, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1082).

Thanks!

pvizeli:
Please update the changelog and bump the version on the config file.

jumping2000:
Ok, I'll do everything this evening. And I will modify also the HSTS parameter as recommended by Mozilla

PR #1107, opened 2020-02-19
Update lets encrypt logic for cloudflare

crdnl:
Add the additional check of if _either_ the token _or_ the keys (and email) exist, and prevent useless calls to lets encrypt.


pvizeli:
I did not like this for cloudflare only. We support more as only cloudflare. As next is that in cerbot issue on his cloudlfare plugin they did not process that correct and make wrong requests. I would suggest a complete solution or make an issue on the corebot repository to fix this on the plugin

Issue #1116, opened 2020-02-20
SSH addon should be able to start with default config. 


pvizeli:
Maybe we can solve it with new options UI they force to set a password if you click on start.

PR #1117, opened 2020-02-20
Support to use SSH as webterminal only

pvizeli:
Fix: #1116 

recommended by @balloob

### DIFF
--- a/ssh/README.md
+++ b/ssh/README.md
@@ -1,6 +1,6 @@
 # Home Assistant Add-on: SSH server
 
-Allow logging in remotely to Home Assistant using SSH.
+Allow logging in remotely to Home Assistant using SSH or just the web terminal with Ingress.
 
 ![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield] ![Supports armhf Architecture][armhf-shield] ![Supports armv7 Architecture][armv7-shield] ![Supports i386 Architecture][i386-shield]
 
@@ -12,7 +12,7 @@ client. It also includes a command-line tool to access the Home Assistant API.
 Try it out using:
 
 ```bash
-hassio help
+ha help
 ```
 
 ## Installation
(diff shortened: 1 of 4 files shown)


ssh/README.md
```diff
@@ -25,6 +25,9 @@ Follow these steps to get the add-on installed on your system:
 
 ## How to use
 
+You need enable the port for external access. You can just enter 22 as value or any other
```
rct: IMHO This doc change isn't helpful to a large number of people.   It just doesn't make sense unless the reader knows exactly what you are talking about. This was a breaking change. See new issue #1125.

ssh/config.json
```diff
@@ -14,7 +15,7 @@
   "hassio_role": "manager",
   "auto_uart": true,
   "ports": {
-    "22/tcp": 22
+    "22/tcp": null
```
rct: IMHO this change should have been a separate change that was advertised as a breaking change.   It should also have been described as "Disable network ssh access by default".

ssh/config.json
```diff
@@ -1,10 +1,11 @@
 {
-  "name": "SSH server",
-  "version": "8.3",
+  "name": "Terminal & SSH",
+  "version": "8.4",
   "slug": "ssh",
   "description": "Allow logging in remotely to Home Assistant using SSH",
```
rct: This description is no longer accurate.  By default the add-on no longer **allows logging in remotely using ssh**

ssh/README.md
frenck: You've opened up an issue, not sure why you would create a comment here as well.

Please don't dig up closed PR's. Thanks.

ssh/config.json
frenck: Please use issues for discussions and bug reports instead, thanks 👍

ssh/config.json
frenck: Feel free to open up a PR to improve a thing, or open up an issue. Let's not use merged PR's for that, as it is not trackable. Thanks 👍

PR #1126, opened 2020-02-22
Re-enable ssh add-on's network ssh access.

rct:
PR #1117 disabled SSH network access for the SSH server add-on.  

Users who previous had the "SSH Server" official add-on installed, will be surprised when they update to version 8.4 that ssh access from the network suddenly stops working.

PR #1117 has effectively changed this add-on from something that enabled ssh access over the network into something that only provides access through a web GUI terminal.

This PR simply makes ssh enabled be default restoring the previous behavior to avoid frustrating existing users.

If the intent is to have SSH network access disabled by default. it should be better communicated as a breaking change in a future PR.  There are a number of ways this could be done. At a minimum documentation updates and possibly advertising it as a breaking change in a future version.

### DIFF
--- a/ssh/config.json
+++ b/ssh/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Terminal & SSH",
-  "version": "8.4",
+  "version": "8.5",
   "slug": "ssh",
   "description": "Allow logging in remotely to Home Assistant using SSH",
   "url": "https://github.com/home-assistant/hassio-addons/tree/master/ssh",
@@ -15,7 +15,7 @@
   "hassio_role": "manager",
   "auto_uart": true,
   "ports": {
-    "22/tcp": null
+    "22/tcp": 22
   },
   "map": ["config:rw", "ssl:rw", "addons:rw", "share:rw", "backup:rw"],
   "options": {


homeassistant:
Hi @rct, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1126).

Thanks!

frenck:
We should document this behavior, not re-instate it.

ocean123:
So the solution is to edit the add-on config of "Terminal & SSH"

 - Add port number "host" in Container | Host | Description
 - save
 - The restart the ssh add-on.

pvizeli:
I will update the supervisor to copy the install port setting over, so that changes like this did not end up in a breaking change

rct:
@pvizeli - That seems like a good idea, anyone who has installed the ssh add-on but hasn’t updated to 8.4 yet will retain ssh network access.

mattrude:
This needs to be relooked at.  There needs to be a way of accessing the system when the addon has failed or the supervisor has stopped loading.

agners:
> This needs to be relooked at. There needs to be a way of accessing the system when the addon has failed or the supervisor has stopped loading.

You can install the add-on, enable SSH and it will stay enabled. Not sure why you think your use case is related to this (very old) PR?

If you are a developer, and have to deal with brokenn Add-ons and/or Supervisor, SSH access on the host system might be the better option for your case (https://developers.home-assistant.io/docs/operating-system/debugging#ssh-access-to-the-host).

PR #1137, opened 2020-02-25
Add /data folder for run.sh

eko1495:
Fix for:
https://github.com/home-assistant/hassio-addons/issues/1134


homeassistant:
Hi @eko1495, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1137).

Thanks!

frenck:
This is an incorrect change, as the data folder is provided by the supervisor as a mount. There is no need to create it.

Your problem is something else, this "fix" works for you probably for a single moment, but would not persist your data stored in there.

PR #1178, opened 2020-03-09
Update Almond to 1.7.3

gcampax:
By the way, would it make sense for the addon version to be the same as the version of Almond?
It would simplify the documentation I think.

### DIFF
--- a/almond/build.json
+++ b/almond/build.json
@@ -5,6 +5,6 @@
     "armv7": "homeassistant/armv7-base-debian:buster"
   },
   "args": {
-    "ALMOND_VERSION": "v1.7.2"
+    "ALMOND_VERSION": "v1.7.3"
   }
 }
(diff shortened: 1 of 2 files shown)


frenck:
Hi there @gcampax,

An add-on contains more parts, not just Almond. So, e.g., is an OS gets updated, or any other dependency, a release has to be done of the add-on as well. In those cases, Almond doesn't have to be changed.

Hence, add-ons do have their own versioning.

gcampax:
Yeah I understand the rationale, but in practice, it doesn't seem that the addon gets rebuilt when there are underlying changes, and Almond updates often enough that the same version number would work. Anyway, if this PR looks good, please merge, I don't have commit access.

frenck:
> Anyway, if this PR looks good, please merge, I don't have commit access.

We will review it soon.

PR #1184, opened 2020-03-15
letsencrypt: Adds cron based daily renewal check and restart on success. …

wonderslug:
This adds some a basic crond based daily check for the certificate
renewal.  It will run the check as a defined 24h:min based time that
the user can define.  As well it allows for the restart of the core
and defined addons by slug on a successful renewal and deployment of
the cert.

Its adds some config options that allow for the renewal cron to be enabled and set the time for the check.  The time is a 24 hour time based HH:MM.  This gets converted to a cron entry to allow for the check for renewal to run.

```json
    "renewal_daily_check": true,
    "renewal_check_time": "02:00",
```

Of note, for development the renewal_check_time can be set to something like "\*:\*/5" which will create a cronjob which tries the renew every 5 minutes.  The time values are not validated at the moment but cron will reject invalid values.

As well on a successful initial creation or renewal it allows for the deployment to run which can now not only copy the certificates over but also restart the core and any defined addons.

```json
    "restart_core": true,
    "restart_addons":[
       "a0d7b954_glances",
       "a0d7b954_nodered"
     ]
```

The additional options are not required and will if missing default to the same behaviour as before with the checks and renewal happening only on a start of the addon.

In addition there are some development config options available to force the certbot to use the "staging" lets encrypt server (https://letsencrypt.org/docs/staging-environment/)  as well as force the renew to renew the cert all the time instead of waiting for expiry.

```json
    "dev_use_staging": "bool",
    "dev_force_renewal": "bool",
```

### DIFF
--- a/letsencrypt/config.json
+++ b/letsencrypt/config.json
@@ -7,6 +7,8 @@
   "arch": ["armhf", "armv7", "aarch64", "amd64", "i386"],
   "startup": "once",
   "boot": "manual",
+  "hassio_api": true,
+  "hassio_role": "manager",
   "ports": {
     "80/tcp": 80
   },
@@ -20,7 +22,11 @@
     "certfile": "fullchain.pem",
     "keyfile": "privkey.pem",
     "challenge": "http",
-    "dns": {}
+    "dns": {},
+    "renewal_daily_check": true,
+    "renewal_check_time": "02:00",
+    "restart_core": true,
+    "restart_addons":[null]
   },
   "schema": {
     "email": "email",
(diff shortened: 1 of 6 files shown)


pvizeli:
Hi @wonderslug, thanks for your contribution.

The Certification handling is pain and functionality like you provide needed for all kinds of add-ons with have certifications. That is the reason why we plan last year to make a certification manager on Supervisor and abstract the hole certification handling with a lot more logic. This logic can be consumed on certification add-ons.

This PR is nice to work but in the wrong place. I would be happy to advise you on Supervisor to work on the certification manager to archive this. Are you onboard?

wonderslug:
Hey @pvizeli 

Sure, im happy to help.  Let me know what your thinking.

letsencrypt/config.json
```diff
@@ -20,14 +22,24 @@
     "certfile": "fullchain.pem",
     "keyfile": "privkey.pem",
     "challenge": "http",
-    "dns": {}
+    "dns": {},
```
chickenandpork, line 25: Might be a nitpicking, but you've put these new lines AFTER dns{} here, meaning dns{} sees delta due to the added comma, but you did differently on line 42.  Won't change behavior but it's inconsistent   :-/

PR #1216, opened 2020-04-11
Remove advanced flag from SSH add-on

ludeeus:
https://www.home-assistant.io/hassio/installation/
During step #7, we ask that the user install this addon.
But it is now hidden.

I think it's better to remove the flag here, instead of telling every user that they have to enable advanced mode in HA for it to show.


frenck:
We could also adjust the documentation instead.
SSH/Web terminals should not be used for users running Home Assistant as a non-advanced mode.

pvizeli:
True

PR #1218, opened 2020-04-12
git_pull: delete key file first so we are not always appending

chickenandpork:
`git_pull` is always appending existing key files, eventually making them usuable

I found that my config would eventually grow stale; it seems to be a few restarts that does it, so when I'm fiddling with the hass config and I hit `restart` on the `git_pull` addon to rush the 5-minute interval.

My impatience made the addon start a number of times, and eventually it would break.

Checking, (`docker exec -it addon_core_git_pull bash`) I found that `/data/options.json`) has a valid key, but my `/root/.ssh/id_rsa` file was 39935 lines long.

Fixing the `/root/.ssh/id_rsa` made the pull/check/restart work again.

### DIFF
--- a/git_pull/CHANGELOG.md
+++ b/git_pull/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.12
+
+- Fix error of deployment_key eventually failing by overwriting the deployment_key every cycle
+
 ## 7.11
 
 - Update Home Assistant CLI to 4.2.0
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @chickenandpork, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1218).

Thanks!

chickenandpork:
Quick ping: Changelog and version were updated; I wonder if you have a chance to approve the merge.

PR #1246, opened 2020-04-23
Updated readme to include invalid Cert files

ecelonghorn:
While trying to use this add-on I noticed that even if the certfile setting is false, if blank /ssl/privkey.pem and /ssl/fullchain.pem files existed, it would not work. I needed to delete these files. I don't necessarily view this as a bug, so I figured adding it to the documentation covers it.

### DIFF
--- a/mosquitto/README.md
+++ b/mosquitto/README.md
@@ -92,6 +92,14 @@ A file containing the private key.
 
 If set to `true` encryption will be enabled using the cert- and keyfile options.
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @ecelonghorn, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1246).

Thanks!

PR #1248, opened 2020-04-24
samba: Added warning to the documentation on veto_files when transferring files with rsync.

Nossnevs:
samba: Added warning to the documentation on veto_files when transferring files with rsync.

### DIFF
--- a/samba/README.md
+++ b/samba/README.md
@@ -71,7 +71,9 @@ List of hosts/networks allowed to access the shared folders.
 
 List of files that are neither visible nor accessible. Useful to stop clients
 from littering the share with temporary hidden files
-(e.g., macOS `.DS_Store` or Windows `Thumbs.db` files)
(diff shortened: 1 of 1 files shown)


samba/README.md
```diff
@@ -71,7 +71,9 @@ List of hosts/networks allowed to access the shared folders.
 
 List of files that are neither visible nor accessible. Useful to stop clients
 from littering the share with temporary hidden files
-(e.g., macOS `.DS_Store` or Windows `Thumbs.db` files)
+(e.g., macOS `.DS_Store` or Windows `Thumbs.db` files). 
+If you transfer files with rsync, *._ will then prevent rsync to transfer
```
frenck, line 75: `*._` does not match `__init__.py`...?

Nossnevs: Yes but when rsync copies the file it tries to creates a temporary file .__ini_.py.[hash] .

pvizeli:
Please make a new PR to DOC.md and we can merge it

PR #521, opened 2019-02-06
Duck DNS alias support

jmorahan:
For when you have another domain CNAME'd to the Duck DNS one.

Context: https://community.letsencrypt.org/t/dont-know-what-content-my-text-record-should-have/85439/2?u=jmorahan

The config would look like:

```
{
  "lets_encrypt": {
    "accept_terms": true,
    "certfile": "fullchain.pem",
    "keyfile": "privkey.pem"
  },
  "token": "sdfj-2131023-dslfjsd-12321",
  "domains": ["subdomain.example.com"],
  "aliases": {
    "subdomain.example.com": "example.duckdns.org"
  },
  "seconds": 300
}
```

PR #1253, opened 2020-04-26
Duckdns alias support

noamokman:
Hey, rebased #521 let me know what else needs to be done to support this

### DIFF
--- a/duckdns/CHANGELOG.md
+++ b/duckdns/CHANGELOG.md
@@ -1,55 +1,18 @@
 # Changelog
 
-## 1.13
-
-- Add DNS alias option to allow pointing a CNAME at a Duck DNS subdomain
-
-## 1.12
+## 1.12.0
 
 - Add option to specify a service or URL as IPv4 and IPv6 address source
+- Add DNS alias option to allow pointing a CNAME at a Duck DNS subdomain
 
-## 1.11
+## 1.11.0
 
 - Do not skip TLS security checks on Duck DNS API access
 
-## 1.10
+## 1.10.0
 
 - Fix issue with dehydrated lock file
 
-## 1.9
+## 1.9.0
 
 - Fix issue with empty IPv4 / IPv6
-
-## 1.8
-
-- Update dehydrated to 0.6.5
-- Migrate scripts to bashio
-
-## 1.7
-
-- Add support for manual set IPv4 and IPv6 address
-
-## 1.6
-
-- Update Alpine images / OpenSSL
-
-## 1.5
-
-- Fix bug with multible domains for Let's encrypt
-
-## 1.4
-
-- Update dehydrated to version 0.6.2
-
-## 1.3
-
-- Fix dehydrated warning from hook script
-
-## 1.2
-
-- Update dehydrated to version 0.6.1
-
-## 1.1
-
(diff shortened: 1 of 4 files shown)


homeassistant:
Hi @noamokman, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1253).

Thanks!

noamokman:
@pvizeli would love your input

noamokman:
Added the new option to the readme

duckdns/README.md
```diff
@@ -89,6 +89,17 @@ The DuckDNS authentication token found at the top of the DuckDNS account landing
...
+### Option: `aliases` (optional)
+
+A list aliases of domains configured on the `domains` option.
+This is useful in cases where you would like to use your own domain.
+Create a CNAME record to point at the DuckDNS subdomain and set this value accrodingly.
+
+For example:
+```json
+[{"domain": "my-domain.duckdns.org", "alias": "ha.my-domain.com"}]
+```
```
pvizeli, line 101: YAML please

noamokman: Fixed 👍

noamokman:
I'll try and test it on my server

noamokman:
OK, tested it here:
https://github.com/noamokman/hassio-repository
It works, but the configuration to make it work is a bit confusing.
In my system it looks like this:
```yaml
domains:
  - mydns.duckdns.org
  - my-custom-domain.com
aliases:
  - domain: my-custom-domain.com
    alias: mydns.duckdns.org
```
This way the certificate holds both domains and the dns challenge succeeds

noamokman:
Another cool tip, I use the nginx proxy, all I had to do is add the new name to the domain in the configuration:
```yaml
# before
domain: mydns.duckdns.org
# after
domain: mydns.duckdns.org my-custom-domain.com
```
That way, both urls still work, with ssl
Should I add it in the docs here?

noamokman:
Thanks!

TheGroundZero:
Does this properly request SSL certs for the alias domains, when they need a specific config?

E.g. a domain hosted at OVH uses the dns-ovh provider with the DNS challenge in the LetsEncrypt plugin.
This provider needs API credentials to be able to create a verification subdomain.  
There's no option (afaik) to add these credentials or to define that a certain domain/alias requires this verification method inside the DuckDNS plugin.

noamokman:
I Don't think so, I just forwarded the subdomain for the challenge

Pabsilon:
> OK, tested it here:
> https://github.com/noamokman/hassio-repository
> It works, but the configuration to make it work is a bit confusing.
> In my system it looks like this:
> 
> ```yaml
> domains:
>   - mydns.duckdns.org
>   - my-custom-domain.com
> aliases:
>   - domain: my-custom-domain.com
>     alias: mydns.duckdns.org
> ```
> 
> This way the certificate holds both domains and the dns challenge succeeds

I'm doing this same configuration and the dns challenge succeeds, but the duckDNS addon seems to have an error when updating the duckdns.org domain IP.
When I remove my custom domain from the domains list, it works correctly, so it looks like the addon is trying to update my custom domain on duckdns.org.

PR #1285, opened 2020-05-10
Rename  DOC.md to DOCS.md

### LINKED ISSUE #1284: DOC.md doesn't seems available in the UI [add-on: google_assistant]
## The problem

The documentation about [Presenting your addon](https://developers.home-assistant.io/docs/hassio_addon_presentation/#adding-documentation) specified that the documentation should reside in a `DOC.md` file in the project.

After merging https://github.com/home-assistant/hassio-addons/pull/1265, I've tried to access the updated documentation within the Supervisor UI without success.

I just don't see the documentation, and if I click on the `Click here for the full documentation` link (which works on GitHub), it redirects me to the same page (`/hassio/addon/core_google_assistant/info`) with a double side-panels:

image

@pvizeli mentioned that the UI support for this is pretty new. Unless I am missing something, it doesn't seems to show at all in the UI.

## Environment

I tested with Home Assistant Core versions 0.109.1 and 0.109.6.

## Additional information

If you want, I can also make a PR to revert back the `DOC.md` into the `README.md`.

ludeeus:
Fixes #1284


google_assistant/config.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Google Assistant SDK",
-  "version": "2.4.0",
+  "version": "2.4.1",
```
ludeeus, line 3: I'm not sure if this is needed.

google_assistant/README.md
```diff
@@ -69,6 +69,7 @@ In case you've found a bug, please [open an issue on our GitHub][issue].
 [armv7-shield]: https://img.shields.io/badge/armv7-yes-green.svg
 [dialogflow-integration]: https://www.home-assistant.io/integrations/dialogflow/
 [discord]: https://discord.gg/c5DvZ4e
+[documentation]: https://github.com/home-assistant/hassio-addons/blob/master/google_assistant/DOCS.md
```
ericmatte, line 72: Is it not better to reference a relative link since the file is in the same repo?

With a full link to GitHub, I think that the user will be redirected directly into GitHub instead of staying on the add-on page in the Supervisor.
I am not sure if this is the expected behaviour. 🤔

ludeeus: With a relative link, you get the issue you had with the dual sidebars ;)

ericmatte: True

ericmatte: But I'm still unsure what is supposed to be the "UI support" with the `DOCS.md` file :P

ludeeus: There will be a new tab on the addon:
image

ericmatte: Oh great!!

Do you think it's possible to make it that when you click on the `Click here for the full documentation`, it redirects the UI to the documentation tab?

If not, a direct link to GitHub is great.

ludeeus: No that is not possible, that would break the link if browsed on Github.

ericmatte:
@ludeeus do you think you could add a title to the `DOCS.md` file?

Something like:
```md
## How to use
```

or
```md
## Installation process
```

ludeeus:
IMO that is not needed.

google_assistant/config.json
frenck: It is not needed for doc updates.

ludeeus:
There is a link issue

ericmatte:
@ludeeus I think that you've missed some of the reference links in the `README.md`:

Screen Shot 2020-05-11 at 15 12 17

```
[google-assistant]: https://assistant.google.com/
[google-actions]: https://actions.google.com/
[dialogflow-integration]: https://www.home-assistant.io/integrations/dialogflow/
```

These can be safely moved from `DOCS.md` to `README.md`, as they are only used in the readme.

PR #1356, opened 2020-05-26
Add feedback_sound option [add-on: google_assistant]

ericmatte:
### Description

With a conventional Google Home device, if you want to hear a sound after saying "Ok Google", you would have to [enable the option in the settings](https://support.google.com/googlenest/answer/7410241?hl=en-CA).

But, since the Google Assistant add-on is recognized as a "speaker" for its device type, this option is not available.

<details>
<summary>See settings screenshot</summary>
<br>

image

</details>

I also tried changing the device type and did not found any type that could have the option.

### Proposal

This PR adds an option to enable custom feedback sounds before and after saying a command; pretty much identical to the one on a Google Home.

It really improves the user experience, as you get a real feedback and always know when your Google Assistant is or is not listening to you.


google_assistant/rootfs/etc/services.d/google-assistant/run
```diff
@@ -23,5 +24,5 @@ elif ! bashio::fs.file_exists "${CRED_JSON}"; then
 fi
 
 bashio::log.info "Starting Home Assistant GAssisant SDK..."
-exec python3 /usr/bin/hassio_gassistant.py \
-    "${CRED_JSON}" "${PROJECT_ID}" "${MODEL_ID}"
+python3 /usr/bin/hassio_gassistant.py \
```
ericmatte, line 27: I've removed the `exec` command here to be constant with how the `hassio_oauth.py` is called (`python3 /usr/bin/hassio_oauth.py`).

It works great without it, but I'm not sure if this could have any impact on how the script is run with Docker. Let me know if I should revert it or not.
I'm not sure

frenck: Please revert that. `exec` hands over the currently running process to that command (completely replacing it).

google_assistant/Dockerfile
```diff
@@ -13,6 +13,7 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         libportaudio2 \
         libasound2-plugins \
+        alsa-utils \
```
pvizeli, line 16: Please use Pulseaudio -> https://linux.die.net/man/1/pacat

ericmatte:
I see that `paplay` has a `--volume` option. What do you guys think this?:
Maybe I could change the option `feedback_sound` from a bool to a int between 0-100, to let the user have control over the volume of the sounds (0 being disabled).

pvizeli:
Great idea I think. Let set `feedback_volume: 80` and the on/off. We have soon a UI for this. Let use somethings like:
```
feedback:
  enable: true
  volume: 80
```

With pulseaudio system, you can also mute all other sound output, and all is done, restore it. But maybe somethings for the future :)

ericmatte:
> Great idea I think. Let set `volume: 80` and the on/off. We have soon a UI for this. Let use somethings like:
> 
> ```
> feedback:
>   enable: true
>   volume: 80
> ```
> 
> With pulseaudio system, you can also mute all other sound output, and all is done, restore it. But maybe somethings for the future :)

Great idead, but I'm not sure if this is the best way to go for this configuration. I think it could be simplified to only:
```
feedback: 80
```
And if `feedback` == 0, then we don't play any feedback sounds.

Also, if I do my schema in `config.json` like below, does that means that users will be forced to update their configuration in order for it to work?
```json
"schema": {
  ...
  "feedback": {
    "enable": "bool",
    "volume": "int(0,100)?"
  }
}
```
vs
```json
"schema": {
  ...
  "feedback": "int(0,100)?"
}
```

What do you think?

ericmatte:
So I've made the changes to use the volume feature and the defaults value are now:
```yaml
feedback:
  enable: false
  volume: 0
```

Everything is wokring as especting on my manual tests.

But I still think that simply one option would be simpler:
```yaml
feedback: 0
```

What do you think?

google_assistant/DOCS.md
```diff
@@ -33,6 +33,10 @@ The ID of the model you've registered at Google for this add-on.
 
 The model id can also be found under the "Develop - Device registration" tab in the [Google Actions Console][google-actions-console].
 
+### Option: `feedback`
```
pvizeli, line 36: Make a grouped https://github.com/home-assistant/hassio-addons/blob/master/duckdns/DOCS.md#option-group-lets_encrypt

ericmatte:
@pvizeli should I add a changelog and update the version number?

pvizeli:
Good work

PR #1359, opened 2020-05-27
nginx_proxy: Fix the use of subfolders in certificate files

frenck:
The use of slashes in cert files, causes the sed replacement to break.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.1
+
+- Fix the use of subfolders with certificate files
+
 ## 3.0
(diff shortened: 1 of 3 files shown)


nginx_proxy/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.1
+
+- Fix the use of subfolders with certificate files
```
ludeeus, line 5:
```suggestion
- Add support for certificate files located in subdirectories
```
Maybe?

pvizeli: It's more a fix. Anyway with the cert manager, such issues gone

PR #1364, opened 2020-05-27
SSH: Add local TCP forwarding option

adamgreg:
Add allow_tcp_forwarding option, to allow local port forwarding by the SSH add-on. Code lifted from the more advanced [SSH & Web Terminal](https://github.com/hassio-addons/addon-ssh) add-on.

Motivating use case: I wanted to use the ptvsd integration to help debug an issue on a remote Home Assistant instance I'm supporting. For this I would like to use port forwarding via the SSH add-on. It would also be useful for accessing the host SSH server without having to modify the router configuration to forward port 22222.

### DIFF
--- a/ssh/config.json
+++ b/ssh/config.json
@@ -23,12 +23,16 @@
   "options": {
     "authorized_keys": [],
     "password": "",
-    "allow_tcp_forwarding": false
+    "server": {
+      "tcp_forwarding": false
+    }
   },
   "schema": {
     "authorized_keys": ["str"],
     "password": "str",
-    "allow_tcp_forwarding": "bool"
+    "server": {
+      "tcp_forwarding": "bool"
+    }
   },
   "image": "homeassistant/{arch}-addon-ssh"
 }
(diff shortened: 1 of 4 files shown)


ssh/config.json
```diff
@@ -22,11 +22,13 @@
   "map": ["config:rw", "ssl:rw", "addons:rw", "share:rw", "backup:rw"],
   "options": {
     "authorized_keys": [],
-    "password": ""
+    "password": "",
+    "allow_tcp_forwarding": false
```
pvizeli, line 26: I think somethings like this would be better:
```yaml
server:
  tcp_forwarding: false
```

adamgreg: Sure, sounds good to me. I just copied the option as-is from the SSH & Web terminal add-on.

adamgreg: I just noticed the `server:` key in your comment. Is the extra level of hierarchy necessary? All of the options are for the SSH server.

pvizeli: yes

PR #1380, opened 2020-06-07
zwave: Add instance config

### LINKED ISSUE #1332: OpenZwave: configure instance
The zwave addon should allow configuration of `ozw_instance` (probably default to `1`), similar to https://github.com/marcelveldt/hassio-addons-repo/blob/master/ozwdaemon/config.json

uSpike:
Fixes #1332

### DIFF
--- a/zwave/config.json
+++ b/zwave/config.json
@@ -25,8 +25,7 @@
   "devices": ["/dev/bus/usb:/dev/bus/usb:rwm", "/dev/mem:/dev/mem:rw"],
   "options": {
     "device": null,
-    "network_key": "",
-    "instance": 1
+    "network_key": ""
   },
   "schema": {
     "device": "str",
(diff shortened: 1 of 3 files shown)


zwave/config.json
```diff
@@ -25,11 +25,13 @@
   "devices": ["/dev/bus/usb:/dev/bus/usb:rwm", "/dev/mem:/dev/mem:rw"],
   "options": {
     "device": null,
-    "network_key": ""
+    "network_key": "",
+    "instance": 1
```
frenck, line 29: Let's remove this from the default configuration. Most people should not need this and can be confused by this setting.

We can still allow setting it, document it, and check on runtime if it is set.

uSpike: OK, I think I've properly changed this.

PR #1393, opened 2020-06-10
Fix broken link mosquitto -> DOCS.md

joth76:
Docs were moved in d31f1c79349ed

### DIFF
--- a/mosquitto/README.md
+++ b/mosquitto/README.md
@@ -6,7 +6,7 @@ MQTT broker for Home Assistant.
 
 ## About
 
-You can use this add-on to install Eclipse Mosquitto, which is an open-source (EPL/EDL licensed) message broker that implements the MQTT protocol. Mosquitto is lightweight and is suitable for use on all devices from low power single board computers to full servers. For more information, please see [mosquitto].
+You can use this add-on to install Eclipse Mosquitto, which is an open-source (EPL/EDL licensed) message broker that implements the MQTT protocol. Mosquitto is lightweight and is suitable for use on all devices from low power single board computers to full servers. For more information, please see [DOCS.md](DOCS.md).
 
 [aarch64-shield]: https://img.shields.io/badge/aarch64-yes-green.svg
 [amd64-shield]: https://img.shields.io/badge/amd64-yes-green.svg


fabaff:
@joth76, can you please sign the CLA? Thanks

mosquitto/README.md
```diff
@@ -6,7 +6,7 @@ MQTT broker for Home Assistant.
 
 ## About
 
-You can use this add-on to install Eclipse Mosquitto, which is an open-source (EPL/EDL licensed) message broker that implements the MQTT protocol. Mosquitto is lightweight and is suitable for use on all devices from low power single board computers to full servers. For more information, please see [mosquitto].
+You can use this add-on to install Eclipse Mosquitto, which is an open-source (EPL/EDL licensed) message broker that implements the MQTT protocol. Mosquitto is lightweight and is suitable for use on all devices from low power single board computers to full servers. For more information, please see [DOCS.md](DOCS.md).
```
frenck, line 9: As requested above, please sign the CLA, we can't merge your PRs without it.

We should remove the link in this case, as this text shows up inside the Supervisor panel as well. This link won't work there.

```suggestion
You can use this add-on to install Eclipse Mosquitto, which is an open-source (EPL/EDL licensed) message broker that implements the MQTT protocol. Mosquitto is lightweight and is suitable for use on all devices from low power single board computers to full servers.
```

joth76: Thanks - CLA signed.
Just to give context on why I think this is helpful, scattered around the internet are hundreds of deep link to this README on github with text saying "for full instructions on setup, please refer to the readme"
Since those instructions were extracted, fools like me end up on this terse README feeling a bit like we've been rickrolled. 
The long term fix would be for all those sites to link directly to the new DOC.md (assuming it doesn't move again) but in the short term I think you really should consider some kind of hint on this page that this isn't all the documentation, there really is some more somewhre.

Would a fully qualified link to https://github.com/home-assistant/hassio-addons/blob/master/mosquitto/DOCS.md work here, including when it is used in the supervisor panel?

frenck: That would work, but it would be a weird experience from the UI.

pvizeli: Yes, that file is for the UI and not for browser over Github

joth76: So here is a notable example of a page that is erroneously linking to this README:
https://www.home-assistant.io/docs/mqtt/broker#embedded-broker-deprecated

There are numerous others like it on the web, and they currently leave users rather stranded.

PR #1395, opened 2020-06-10
Clarify exactly how to get to the configuration

joth76:
The example snipppet is hard to apply for someone that may have completed the "install" step months ago and then come back to this howto to do configuration.

### DIFF
--- a/mosquitto/DOCS.md
+++ b/mosquitto/DOCS.md
@@ -29,6 +29,8 @@ If you have old MQTT settings available, remove this old integration and restart
 
 ## Configuration
 
+After the initial installation process described above, the configuration editor can be found in **Supervisor** -> **Mosquitto broker** -> **Configuration** tab.
+
 Add-on configuration:
 
 ```yaml


mosquitto/DOCS.md
```diff
@@ -29,6 +29,8 @@ If you have old MQTT settings available, remove this old integration and restart
 
 ## Configuration
 
+After the initial installation process described above, the configuration editor can be found in **Supervisor** -> **Mosquitto broker** -> **Configuration** tab.
```
fabaff, line 32: From my point of view, this applies to all add-ons. With the recent UI changes it could make sense to add it somewhere at https://www.home-assistant.io/hassio/

joth76: Ah OK, yes that's fair.
I think what caught me out here is I arrived at the Add-on instructions having already read through the instructions on https://www.home-assistant.io/docs/mqtt/broker/ (being the top Google hit for "home assistant mosquitto config") and that page describes editing `configuration.yaml` so I was still in that mindset when eventually finding these Add-on docs (and being new to Add-ons didn't know this wider convention)
Thanks

frenck:
I agree with fabaff in this case. If an upstream doc is not correct, we should fix that.

PR #1407, opened 2020-06-13
Fix #1406: modify documented network_key

### LINKED ISSUE #1406: add an option to allow using network key from the documentation
Recently, it was introduced a check to make sure users were not using the network-key from the documentation https://github.com/home-assistant/hassio-addons/blob/master/zwave/rootfs/etc/cont-init.d/network_key.sh#L5

I am in the unfortunate situation of having a big zwave mesh using the network key from the documentation (which was already used in the previous documentation) which mean that I am unable to use this new official add-on. I know it is my fault and I should not have done that but I do not want to have to reset my whole network and I'm comfortable with the fact that my network is not really secure because of that key.

Could you please evaluate two solutions:
- either provide the possibility to force this network key even though it is dangerous (the 'know-what-i-am-doing' option)
- modify the network key from the documentation (and the one checked in that script) so that it only affect new users and not users that had an integration not failing at start with the previous zwave component and built a big zwave mesh.

I know it is my fault but I would really appreciate if this was considered so I can migrate to this new add-on. Thanks for the great work btw

rubenfiszel:
Fix #1406 

> Recently, it was introduced a check to make sure users were not using the network-key from the documentation https://github.com/home-assistant/hassio-addons/blob/master/zwave/rootfs/etc/cont-init.d/network_key.sh#L5
> 
> I am in the unfortunate situation of having a big zwave mesh using the network key from the documentation (which was already used in the previous documentation) which mean that I am unable to use this new official add-on. I know it is my fault and I should not have done that but I do not want to have to reset my whole network and I'm comfortable with the fact that my network is not really secure because of that key.
> 
> Could you please evaluate two solutions:
> 
>  1. either provide the possibility to force this network key even though it is dangerous (the 'know-what-i-am-doing' option)
>  2. modify the network key from the documentation (and the one checked in that script) so that it only affect new users and not users that had an integration not failing at start with the previous zwave component and built a big zwave mesh.
> 
> I know it is my fault but I would really appreciate if this was considered so I can migrate to this new add-on. Thanks for the great work btw

This is a PR corresponding to the second proposal. I generated randomly that new key


homeassistant:
Hi @rubenfiszel, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1407).

Thanks!

frenck:
You are changing the example key in our documentation for a long time already:

<https://www.home-assistant.io/docs/z-wave/installation/>

This should not be changed and for real, not be used.

PR #1436, opened 2020-06-29
mariadb: Alpine 3.12, increase innodb_buffer_pool_size

frenck:
## 2.2.0

- Upgrade Alpine Linux to 3.12
- Increase innodb_buffer_pool_size to 128M

The increase of the innodb_buffer_pool_size from 64M -> 128M, will add 64M of additional memory usage to the add-on.

Table locks are stored in the InnoDB buffer pool and with 64M, it can run short and causes "1206: The total number of locks exceeds the lock table size" to be thrown. Especially when locks are quickly made after one another.

While 128M is not the holy grail, it should significantly reduce the risks of running into such an issue. In case it does happen, <https://github.com/home-assistant/core/pull/37228> triggers a retry.

### DIFF
--- a/mariadb/CHANGELOG.md
+++ b/mariadb/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.2.0
+
+- Upgrade Alpine Linux to 3.12
+- Increase innodb_buffer_pool_size to 128M
+
 ## 2.1.2
 
 - Fix S6-Overlay shutdown timeout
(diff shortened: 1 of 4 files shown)


Shamshala:
Just out of curiosity, is there any chance to have `innodb_buffer_pool_size` as a user configurable value? In case of hosts with more RAM, it can pretty significantly speed up information exchange between HA and database. :slightly_smiling_face:

PR #1450, opened 2020-07-09
Don't delete the mariadb.sys yser

iangregory:
Since MariaDB 10.4.13 `mysql.user` (and probably others) have a new definer `mariadb.sys` instead of `mysql.sys` that should be added to the list of users which are retained when MariaDB is set up.

The Docker library has fixed their bootstrap code in https://github.com/docker-library/mariadb/pull/306 - this change applies the same fix here.

Deleting the `mariadb.sys` user causes a problem when selecting from the `user` table, and may cause issues when MariaDB is upgraded. If so, the following can be used to restore the user:
```sql
insert into mysql.global_priv values ('localhost','mariadb.sys','{"access":0,"plugin":"mysql_native_password","authentication_string":"","account_locked":true,"password_last_changed":0}');
flush privileges;
```

### DIFF
--- a/mariadb/CHANGELOG.md
+++ b/mariadb/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Changelog
+## 2.2.1
+- Don't delete the mariadb.sys user, it's needed in MariaDB >= 10.4.13
 
 ## 2.2.0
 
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @iangregory, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1450).

Thanks!

pvizeli:
Can you also bump the version with +0.0.1 and update the changelog?

PR #1458, opened 2020-07-12
Improves Terminal & SSH Add-On documentation

FredrikWendt:
- removed mentions of the add-on's old name
- tried to make the whole page flow better (remove duplication, minor reflow of sections)

### DIFF
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -9,7 +9,6 @@ Follow these steps to get the add-on installed on your system:
 3. Find the "Terminal & SSH" add-on and click it.
 4. Click on the "INSTALL" button.
 
-
 ## How to use
 
 This add-on enables two main features to your Home Assistant installation:
@@ -24,7 +23,6 @@ This add-on comes bundled with [The Home Assistant CLI](https://www.home-assista
 ```bash
 ha help
 ```
-
 
 ### The In-Browser Terminal
 
@@ -42,7 +40,6 @@ You can then connect to the port specified, using the username `root`. Please no
 If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is therefor generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
 
 Enabling login via password, will disable key based login. You can not run both variants at the same time.
-
 
 ## Configuration
 
@@ -76,19 +73,14 @@ Specifies whether TCP forwarding is permitted or not.
 
 **Note**: _Enabling this option lowers the security of your SSH server! Nevertheless, this warning is debatable._
 
-
 ## Network
 
-This section is only relevant if you want to connect to Home Assistant using an SSH client, such as PuTTY. To enable SSH access via the network, you need to specify which port to use on the Home Assistant host. The number you enter, will be used to map that port from the host into the running "Terminal & SSH" container.
-
-The standard port used for the SSH protocol is `22`.
-
+This section is only relevant if you want to connect to Home Assistant using an SSH client, such as PuTTY. To enable SSH access via the network, you need to specify which port to use on the Home Assistant host. The number you enter, will be used to map that port from the host into the running "Terminal & SSH" container. The standard port used for the SSH protocol is `22`.
 
 ## Known issues and limitations
 
 - This add-on will not enable you to install packages or do anything as root.
   This is not working with Home Assistant.
-
 
 ## Support
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -15,7 +15,7 @@ This add-on enables two main features to your Home Assistant installation:
 - a Terminal that you can use from your browser, and
 - enable connecting to your system using an SSH client.
 
-Regardless of how you connect (using the Terminal, or a using an SSH client), you end up in this add-on's container. The Home Assistant configuration 
+Regardless of how you connect (using the web Terminal or using an SSH client), you end up in this add-on's container. The Home Assistant configuration
 directory is located on the path `/config`.
 
 This add-on comes bundled with [The Home Assistant CLI](https://www.home-assistant.io/hassio/commandline/). Try it out using:
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -11,11 +11,11 @@ Follow these steps to get the add-on installed on your system:
 
 ## How to use
 
-This add-on enables two main features to your Home Assistant installation:
-- a Terminal that you can use from your browser, and
+This add-on adds two main features to your Home Assistant installation:
+- a web terminal that you can use from your browser, and
 - enable connecting to your system using an SSH client.
 
-Regardless of how you connect (using the web Terminal or using an SSH client), you end up in this add-on's container. The Home Assistant configuration
+Regardless of how you connect (using the web terminal or using an SSH client), you end up in this add-on's container. The Home Assistant configuration
 directory is located on the path `/config`.
 
 This add-on comes bundled with [The Home Assistant CLI](https://www.home-assistant.io/hassio/commandline/). Try it out using:
@@ -26,7 +26,7 @@ ha help
 
 ### The Web Terminal
 
-To enable the web Terminal all you need to do is start the add-on, and click the Terminal menu item.
+You can access the web terminal by clicking the "Open Web UI" button on this add-on's Info tab. If you set the "Show in sidebar" setting (found on the same Info tab) to "on", a shortcut is added to the sidebar allowing you to quickly access to the web terminal.
 
 ### SSH Server Connection
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -16,7 +16,7 @@ This add-on enables two main features to your Home Assistant installation:
 - enable connecting to your system using an SSH client.
 
 Regardless of how you connect (using the Terminal, or a using an SSH client), you end up in this add-on's container. The Home Assistant configuration 
-directory is mounted on the path `/config`.
+directory is located on the path `/config`.
 
 This add-on comes bundled with [The Home Assistant CLI](https://www.home-assistant.io/hassio/commandline/). Try it out using:
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -24,9 +24,9 @@ This add-on comes bundled with [The Home Assistant CLI](https://www.home-assista
 ha help
 ```
 
-### The In-Browser Terminal
+### The Web Terminal
 
-To enable the in-browser Terminal all you need to do is start the add-on, and click the Terminal menu item.
+To enable the web Terminal all you need to do is start the add-on, and click the Terminal menu item.
 
 ### SSH Server Connection
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -39,7 +39,7 @@ You can then connect to the port specified, using the username `root`. Please no
 
 If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is therefore generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
 
-Enabling login via password, will disable key based login. You can not run both variants at the same time.
+Enabling login via password will disable key-based login. You can not run both variants at the same time.
 
 ## Configuration
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -57,7 +57,7 @@ server:
 
 Your **public keys** that you wish to accept for login. You can authorize multiple keys by adding multiple public keys to the list.
 
-If you get errors when adding your key, it is likely that the public key you're trying to add, contains characters that intervene with Yaml syntax. Try enclosing your key in double quotes to avoid this issue.
+If you get errors when adding your key, it is likely that the public key you're trying to add, contains characters that intervene with YAML syntax. Try enclosing your key in double quotes to avoid this issue.
 
 ### Option: `password`
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -1,4 +1,4 @@
-# Home Assistant Add-on: SSH Server
+# Home Assistant Add-on: Terminal & SSH
 
 ## Installation
 
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -4,7 +4,7 @@
 
 Follow these steps to get the add-on installed on your system:
 
-1. This addon is only visible to "Advanced Mode" users. To enable advanced mode, go to **Profile** -> and turn on **Advanced Mode**.
+1. This add-on is only visible to "Advanced Mode" users. To enable advanced mode, go to **Profile** -> and turn on **Advanced Mode**.
 2. Navigate in your Home Assistant frontend to **Supervisor** -> **Add-on Store**.
 3. Find the "Terminal & SSH" add-on and click it.
 4. Click on the "INSTALL" button.
@@ -12,6 +12,7 @@ Follow these steps to get the add-on installed on your system:
 ## How to use
 
 This add-on adds two main features to your Home Assistant installation:
+
 - a web terminal that you can use from your browser, and
 - enable connecting to your system using an SSH client.
 
@@ -26,7 +27,7 @@ ha help
 
 ### The Web Terminal
 
-You can access the web terminal by clicking the "Open Web UI" button on this add-on's Info tab. If you set the "Show in sidebar" setting (found on the same Info tab) to "on", a shortcut is added to the sidebar allowing you to quickly access to the web terminal.
+You can access the web terminal by clicking the "Open Web UI" button on this add-on's Info tab. If you set the "Show in sidebar" setting (found on the same Info tab) to "on", a shortcut is added to the sidebar allowing you to access the web terminal quickly.
 
 ### SSH Server Connection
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @FredrikWendt, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1458).

Thanks!

ssh/DOCS.md
```diff
@@ -6,50 +6,57 @@ Follow these steps to get the add-on installed on your system:
...
+## How to use
 
-1. Add a ssh key to  `authorized_keys` or set a `password` in the add-on configuration.
-2. Start the add-on.
-3. Connect to your device using your preferred SSH client and use `root` as
-   the username.
+Once installed (following the steps above) you need to
+* enable external access by selecting which TCP port to open up (see Network section), and
+* supply login credentials,
+* start the add-on.
```
frenck, line 18: This is now incorrect. You don't have to use SSH. Furthermore, external access can be confused with opening ports up in a router.

We use dashes for lists (`-`) in Markdown and please start items in a list with a capital letter.

ssh/DOCS.md
```diff
@@ -65,14 +72,22 @@ Specifies whether TCP forwarding is permitted or not.
...
+
+The standard port for the SSH protocol is `22`.
+
 
 ## Known issues and limitations
 
 - This add-on will not enable you to install packages or do anything as root.
   This is not working with Home Assistant.
 
+
```
frenck, line 90: Not needed extra new line.

ssh/DOCS.md
```diff
@@ -65,14 +72,22 @@ Specifies whether TCP forwarding is permitted or not.
...
 ## Network
-To enable ssh access via the network, you need to enter the port number ‘22’ or the port you want to use. This will map that port from the hassio host into the running “Terminal & SSH” container.
+
+To enable ssh access via the network, you need to specify which external port to use. 
+The number you enter, will be used to map that port from the hassio host into the 
+running "Terminal & SSH" container.
+
+The standard port for the SSH protocol is `22`.
+
```
frenck, line 83: Not needed extra new line.

ssh/DOCS.md
```diff
@@ -65,14 +72,22 @@ Specifies whether TCP forwarding is permitted or not.
 
 **Note**: _Enabling this option lowers the security of your SSH server! Nevertheless, this warning is debatable._
 
+
 ## Network
-To enable ssh access via the network, you need to enter the port number ‘22’ or the port you want to use. This will map that port from the hassio host into the running “Terminal & SSH” container.
+
+To enable ssh access via the network, you need to specify which external port to use.
```
frenck, line 78: The term `SSH` needs to be capitalized, the external port might be confusing for the end user.

ssh/DOCS.md
```diff
@@ -65,14 +72,22 @@ Specifies whether TCP forwarding is permitted or not.
 
 **Note**: _Enabling this option lowers the security of your SSH server! Nevertheless, this warning is debatable._
 
+
 ## Network
-To enable ssh access via the network, you need to enter the port number ‘22’ or the port you want to use. This will map that port from the hassio host into the running “Terminal & SSH” container.
+
+To enable ssh access via the network, you need to specify which external port to use. 
+The number you enter, will be used to map that port from the hassio host into the
```
frenck, line 79: We don't use the term `hassio` anymore. Use `Home Assistant` instead.

ssh/DOCS.md
```diff
@@ -6,50 +6,57 @@ Follow these steps to get the add-on installed on your system:
...
 
 ### Option: `authorized_keys`
 
 Your **public keys** for the authorized key file. You can authorize multiple
-keys by adding multiple public keys to the list.
+keys by adding multiple public keys to the list. If you're on Linux, you can
+likely issue the command `ssh-add -L` to get a list of your public keys that 
+are available to you.
+
+If you get errors when adding your key, try enclosing it in double quotes to avoid .
```
frenck, line 59: missing word in the end?

ssh/DOCS.md
```diff
@@ -6,50 +6,57 @@ Follow these steps to get the add-on installed on your system:
...
+server:
+  tcp_forwarding: false
 ```
 
 ### Option: `authorized_keys`
 
 Your **public keys** for the authorized key file. You can authorize multiple
-keys by adding multiple public keys to the list.
+keys by adding multiple public keys to the list. If you're on Linux, you can
+likely issue the command `ssh-add -L` to get a list of your public keys that
```
frenck, line 56: This is unclear to the average user? Do I need to type that command in the terminal provided by the add-on? Or what if I don't have Linux?

ssh/DOCS.md
```diff
@@ -6,50 +6,57 @@ Follow these steps to get the add-on installed on your system:
...
-Try it out using:
+This add-on comes bundled with [The Home Assistant CLI](https://www.home-assistant.io/hassio/commandline/). Try it out using:
 
 ```bash
 ha help
 ```
 
+### General Note on Security
+
+You are strongly recommended to use private/public keys to log in.
```
frenck, line 32: Why?

ssh/DOCS.md
FredrikWendt: I can certainly adapt to use `-` in Markdown bullet lists! I searched for, but didn't find a style guide for documentation (or use of Markdown features) in this project. I looked here: https://developers.home-assistant.io/ and here: https://developers.home-assistant.io/docs/development_guidelines (and somehow missed the top menu item Documentation).

As I failed to find a style guide for docs, I went with what most references online use, which is `*`. Example references (top two results when I use Google to search for markdown bullet list): 
- https://www.markdownguide.org/basic-syntax/#adding-elements-in-lists
- https://guides.github.com/features/mastering-markdown/

I now spent some more time, and see that https://developers.home-assistant.io/docs/documenting/ exists, there's no recommendation on what should be preferred. In fact, the only reference I can find to Markdown is this: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code which suggests any use of `*-+`, and lists them in that order with `*` being the first one used to demonstrate bullet lists.

I'm happy to make a change (PR) in https://developers.home-assistant.io/docs/documenting/standards/ if other that want to contribute to keeping documentation up-to-date will get remarks about this undocumented preferred style.

ssh/DOCS.md
FredrikWendt: Fixed, thanks.

ssh/DOCS.md
FredrikWendt: Fixed, thanks.

ssh/DOCS.md
FredrikWendt: Thanks, I removed this.

ssh/DOCS.md
FredrikWendt: The intent with my contribution was to clean up this page, as I discovered the documentation was not up to date with the implementation when I - as a normal user - installed this core Add-On.

In doing so (updating documentation to match behavior present in latest stable release), I have no reason to question the original authors' suggestions - perhaps they've implemented logging of passwords in clear text? While I happen to be a software developer, I have not read the implementation, and I hope this project accepts documentation improvements without having to have read, understood and need to reason about the applicability of the documented recommended use of the software it documents.

On the other hand, as a professional developer, I would argue that keys stored on an external card (or similar device), with two factor protection (both key and the external device are individually protected) provides a much stronger scheme, compared to username+password combination that is always sent over the network. Sending your public key over the network, compared to a password, is always preferable. (It is unclear what process HA suggests be put in place for rotating the server's keys, and help humans understand what "the server's key fingerprint has changed" really means, but I believe that's out of scope for this add-on, and this page as well.)

Some parts of the internet tend to agree, such as discussed here: https://security.stackexchange.com/questions/69407/why-is-using-an-ssh-key-more-secure-than-using-passwords

Now, I can personally agree that making a recommendation may be "over the top". I'll reword this section along the lines of "if you enable external access to you machine, using keys is preferred over password based authentication".

ssh/DOCS.md
FredrikWendt: Fixed ssh, and reworded "external port" to try to avoid the confusion, thanks.

ssh/DOCS.md
FredrikWendt: I need help to disambiguate this comment, as there are four new line characters in play in this comment, and I can see that at least three could be seen as "extra new line". I'm unsure which one you refer to, and guess it's one of these:
* the one separating the two sections (line 81), or
* the one adding white space between the next 2nd level headline (line 83).

ssh/DOCS.md
FredrikWendt: The use of new lines in this documentation file was inconsistent. I normalized this so sections could be found by quickly scanning the document. I went with what I've seen before, which is that headings of level 1 and 2 (`#` and `##`) have two empty lines above them.

I've not seen the suggestion of a linter or similar, just that Black is used for source code - nothing around consistency of flow, ' vs *, ... in documentation. If there is expected formatting to be used, I suggest it is added to https://developers.home-assistant.io/docs/documenting/standards

ssh/DOCS.md
FredrikWendt: In response to your first statement:

I didn't realize the in-browser Terminal could be enabled from the current documentation. As I learned this now, I more or less rewrote the whole section to take this into account. While doing so, the text now nudges towards that one should only enable the SSH Server/Client part "if you know what you're doing".

ssh/DOCS.md
frenck: multiple new lines have no effect. please remove them as requested. Thanks.

ssh/DOCS.md
frenck: multiple new lines have no effect. please remove them as requested. Thanks.

ssh/DOCS.md
frenck: Thanks, I was very well aware why. However, the user that reads that maybe not. Hence why question.

FredrikWendt: Ah. From your "Why?" comment, I didn't understand you wanted me to educate or motivate the recommendation to the end user. I assume the new wording is sufficient (please reopen if you feel otherwise).

ssh/DOCS.md
FredrikWendt: We disagree, I've updated as per your request.

ssh/DOCS.md
FredrikWendt: > multiple new lines have no effect

I disagree on whether white space and formatting affects readability of/ability to (quickly) consume a text, but can adapt.
I agree it makes not difference for the rendered documentation, but it makes a difference for the person that needs to maintain the source documentation.

In my opinion, you're prioritizing a program's processing aspects of a text, over humans' ability to process a text. There's a book titled Clean Code that has several examples of this, for source code. I'm arguing you're optimizing for the smaller cost in your project, rather than the by orders of magnitude larger cost. The highest cost for any change to a system, is almost always the process that involves a human interpreting text (maintaining the Markdown version), never the compiler/processor.

For future, I suggest this project adapt https://github.com/markdownlint/markdownlint and integrate it into the CI process.

frenck:
"reduced Markdown readability - fewer \n characters"

Really now? That was needed?

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
-# Home Assistant Add-on: SSH server
+# Home Assistant Add-on: SSH Server
```
frenck, line 1: Let's correct the name so it matches the actual add-on name, see:

<https://github.com/home-assistant/hassio-addons/blob/master/ssh/config.json#L2>

```suggestion
# Home Assistant Add-on: Terminal & SSH
```

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
-value like you want. This recommend to add login credentials.
+This add-on enables two main features to your Home Assistant installation:
+- a Terminal that you can use from your browser, and
+- enable connecting to your system using an SSH client.
 
-To use this add-on, you must have a private/public key to log in.
-To generate them, follow the [instructions for Windows][keygen-windows]
-and [these for other platforms][keygen]. It is possible to set a password for
-login since version 2.0 but for high security use private/public keys.
+Regardless of how you connect (using the Terminal, or a using an SSH client), you end up in this add-on's container. The Home Assistant configuration
```
frenck, line 18:
```suggestion
Regardless of how you connect (using the web Terminal or using an SSH client), you end up in this add-on's container. The Home Assistant configuration
```

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
+This add-on enables two main features to your Home Assistant installation:
+- a Terminal that you can use from your browser, and
+- enable connecting to your system using an SSH client.
 
-To use this add-on, you must have a private/public key to log in.
-To generate them, follow the [instructions for Windows][keygen-windows]
-and [these for other platforms][keygen]. It is possible to set a password for
-login since version 2.0 but for high security use private/public keys.
+Regardless of how you connect (using the Terminal, or a using an SSH client), you end up in this add-on's container. The Home Assistant configuration 
+directory is mounted on the path `/config`.
```
frenck, line 19: Maybe avoid the term "mounted" here, as it might be confusing (as one can not control the mounts or might not even know what a mount is).

```suggestion
directory is located on the path `/config`.
```

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
-Try it out using:
+This add-on comes bundled with [The Home Assistant CLI](https://www.home-assistant.io/hassio/commandline/). Try it out using:
 
 ```bash
 ha help
 ```
 
+### The In-Browser Terminal
+
+To enable the in-browser Terminal all you need to do is start the add-on, and click the Terminal menu item.
```
frenck, line 29: This menu item does not appear by default. It has to be specifically placed into the menu by the end-user. Alternatively, the button to open the web UI on the add-on page can be used.

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
+
+To connect using an SSH client, such as PuTTY, you need to supply additional configuration for this add-on. To enable SSH connectivity, you need to:
+
+- Provide authentication credentials - a password or SSH key(s)
+- Specify which TCP port to bind to, on the Home Assistant host
+
+You can then connect to the port specified, using the username `root`. Please note that enabling the SSH Server potentially makes your Home Assistant system less secure, as it might enable anyone on the internet to try to access your system. The security of your system also depends on your network set up, router settings, use of firewalls, etc. As a general recommendation, you should not activate this part of the add-on unless you understand the ramifications.
+
+If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is therefor generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
```
frenck, line 40:
```suggestion
If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is therefore generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
```

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
+
+Enabling login via password, will disable key based login. You can not run both variants at the same time.
```
frenck, line 42:
```suggestion
Enabling login via password will disable key-based login. You can not run both variants at the same time.
```

ssh/DOCS.md
```diff
@@ -1,55 +1,63 @@
...
+  tcp_forwarding: false
 ```
 
 ### Option: `authorized_keys`
 
-Your **public keys** for the authorized key file. You can authorize multiple
-keys by adding multiple public keys to the list.
+Your **public keys** that you wish to accept for login. You can authorize multiple keys by adding multiple public keys to the list.
+
+If you get errors when adding your key, it is likely that the public key you're trying to add, contains characters that intervene with Yaml syntax. Try enclosing your key in double quotes to avoid this issue.
```
frenck, line 60:
```suggestion
Your **public keys** that you wish to accept for login. You can authorize multiple keys by adding multiple public keys to the list.

If you get errors when adding your key, it is likely that the public key you're trying to add, contains characters that intervene with YAML syntax. Try enclosing your key in double quotes to avoid this issue.
```

ssh/DOCS.md
```diff
@@ -66,7 +74,8 @@ Specifies whether TCP forwarding is permitted or not.
 **Note**: _Enabling this option lowers the security of your SSH server! Nevertheless, this warning is debatable._
 
 ## Network
-To enable ssh access via the network, you need to enter the port number ‘22’ or the port you want to use. This will map that port from the hassio host into the running “Terminal & SSH” container.
+
+This section is only relevant if you want to connect to Home Assistant using an SSH client, such as PuTTY. To enable SSH access via the network, you need to specify which port to use on the Home Assistant host. The number you enter, will be used to map that port from the host into the running "Terminal & SSH" container. The standard port used for the SSH protocol is `22`.
```
frenck, line 78:
```suggestion
This section is only relevant if you want to connect to Home Assistant using an SSH client, such as PuTTY. To enable SSH access via the network, you need to specify which port to use on the Home Assistant host. The number you enter will be used to map that port from the host into the running "Terminal & SSH" add-on. The standard port used for the SSH protocol is `22`.
```

ssh/DOCS.md
frenck: Thanks for the education. Markdownlint would not have agreed as well. Furthermore, markdownlint isn't the goto linter for Markdown anymore, instead, we use Remarklint, which is slowly being rolled out in our repositories.

FredrikWendt:
> "reduced Markdown readability - fewer \n characters"
> 
> Really now? That was needed?

Not sure, it answers these two (IMO valid) questions which a reviewer may ask:

a) what's been done to this text, from a human perspective? What's the driver for the change?
b) what's been done, technically? What's the implementation (details)?

Perhaps I misunderstood the intent of your requested changes.

ssh/DOCS.md
FredrikWendt: Absolutely agree!

ssh/DOCS.md
FredrikWendt: Oh, this is a surprise to me - I don't remember placing it in the sidebar. Where is this setting located?
I suggest something like:
> To access the in-browser Terminal, start the add-on and then click the Open web UI button. If you find yourself using the Terminal a lot, you can add a menu item that takes you directly to the Terminal, by doing X......

ssh/DOCS.md
FredrikWendt: yes, perhaps we should change "In-Browser Terminal" to "Web Terminal"?

ssh/DOCS.md
FredrikWendt: I was initially hesitant about this, as a rule-of-thumb (in Swedish) is that you only uppercase it if you would pronounce the acronym letter by letter, like Y A M L.

I read up on this a bit and it seems for English, it's all about what the reader is familiar with. There are two other uses of the uppercase spelling (`YAML`) in this project's documentation, so it makes absolutely sense to stick to that.

ssh/DOCS.md
FredrikWendt: @frenck - this section was more or less completely rewritten. I believe the issues you raised are (better) addressed with the new text. Are we OK to close this?

ssh/DOCS.md
frenck: Sure 👍

ssh/DOCS.md
frenck: There is a toggle on the add-on page to toggle the sidebar menu item... It is not enabled by default (this is the same for all add-ons supporting Ingress)

FredrikWendt: Updated.

frenck (approved):
Thanks, @FredrikWendt 👍

PR #1471, opened 2020-07-18
ssh: improve documentation

agners:
The add-on gained more capability over time. This lead to the
documentation being a bit confusing. Update the doc to document the
current add-on capability.


frenck:
There is already in a PR in progress for this: #1458

agners:
Yeah I just saw that too, sorry about this. Will contribute to that PR and create another one in case I still feel there is more cleanup needed.

Issue #1427, opened 2020-06-23
LetsEncrypt - Cannot obtain certificate via dns-ovh


TheGroundZero:
Hi all,

I'm trying to use the LetsEncrypt addon with the dns challenge to provide SSL certs for my Home Assistant.
I want to use DNS-OVH as my DNS provider.

This is a censored version of the config I want to use:

```yaml
email: email@example.org
domains:
  - sub.parent.tld
certfile: fullchain.pem
keyfile: privkey.pem
challenge: dns
dns:
  - dnsprovider: dns-ovh
  - dns_ovh_endpoint: ovh-eu
  - dns_ovh_application_key: application_key_123
  - dns_ovh_application_secret: application_secret_123
  - dns_ovh_consumer_key: consumer_key_123
```

The moment I add the dns-options, I'm getting errors.
![image](https://user-images.githubusercontent.com/2406013/85400306-7ef5aa00-b558-11ea-9d73-2b22ac58c741.png)

Sticking with the default `dns: {}` I'm not getting errors, but I suppose the DNS challenge would then fail since it cannot connect to my DNS provider (OVH).

---

Also, the documentation on getting the OVH key (especially the consumer key) isn't very clear and the permissions are a bit too wide.

To get the key:
- register a new app via https://eu.api.ovh.com/createApp/  
   This will give you `application_key` and `application_secret`
- go to the API console at https://eu.api.ovh.com/console/#/auth/credential#POST
  - Add the necessary access rules  
     I limited my permissions to GET/POST/PUT/DELETE on `/domain/zone/parent.tld` instead of `/domain/zone/*`.
  - Enter a redirection URL, e.g. your own domain name, e.g. `https://example.org`
  - Click execute
  - Copy the consumerKey from the output
- Browse to the validationURL and authenticate to active your key

TheGroundZero:
I captured these error logs in Supervisor > System > System Log

```
20-06-23 21:37:37 INFO (MainThread) [supervisor.addons] Create Home Assistant add-on data folder /data/addons/data/core_letsencrypt
20-06-23 21:37:37 INFO (SyncWorker_7) [supervisor.docker.interface] Pull image homeassistant/amd64-addon-letsencrypt tag 4.10.0.
20-06-23 21:37:54 INFO (MainThread) [supervisor.addons] Add-on 'core_letsencrypt' successfully installed
20-06-23 21:38:21 ERROR (MainThread) [aiohttp.server] Error handling request
[...]

PR #1491, opened 2020-07-22
Improve DNS-OVH documentation

lambtho12:
Give more details about the API Key requirements for DNS challenge with OVH, following [certbot documentation](https://certbot-dns-ovh.readthedocs.io/en/stable/). 
Added example configuration and details about api_endpoint.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -277,7 +277,31 @@ dns:
 ```
 
 ### OVH
-Access to https://eu.api.ovh.com/createToken/ to create the required keys. (Use https://ca.api.ovh.com/createToken/ for north america region)
+You will need to generate an OVH API Key first at https://eu.api.ovh.com/createToken/ (for Europe) or https://ca.api.ovh.com/createToken/ (for north america). 
+
+When creating the API Key, you must ensure that the following rights are granted:
+* ``GET /domain/zone/*``
+* ``PUT /domain/zone/*``
+* ``POST /domain/zone/*``
+* ``DELETE /domain/zone/*``
+
+Example configuration
+```yaml
+email: hello@home-assistant.io
+domains:
+  - home-assistant.io
+certfile: fullchain.pem
+keyfile: privkey.pem
+challenge: dns
+dns:
+  provider: dns-ovh
+  ovh_endpoint: ovh-eu
+  ovh_application_key: 0123456789abcdef0123456789abcdef01234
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @lambtho12, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1491).

Thanks!

TheGroundZero:
Wouldn't it be better to also limit the scope of the OVH API credentials?

See #1427
> I limited my permissions to GET/POST/PUT/DELETE on /domain/zone/parent.tld instead of /domain/zone/*

So for `homeassistant.domain.com`

* ``GET /domain/zone/domain.com``
* ``PUT /domain/zone/domain.com``
* ``POST /domain/zone/domain.com``
* ``DELETE /domain/zone/domain.com``


Instead of 

* ``GET /domain/zone/*``
* ``PUT /domain/zone/*``
* ``POST /domain/zone/*``
* ``DELETE /domain/zone/*``

lambtho12:
Yes that would probably be preferable if it works. Did you test that? I just copied what was in the certbot docs.

TheGroundZero:
I recently redid the installation, since I previously had issues (the issues mentioned in #1427 are mine).

You first need to visit https://eu.api.ovh.com/createApp/ to create an API Application.
image

Then you get the `Application Key` and `Application Secret`
image

They you either use the [API Explorer](https://eu.api.ovh.com/console/#/auth/credential#POST) or send a cURL request to the [API endpoint](https://eu.api.ovh.com/1.0/auth/credential).
image

You then get a validation URL where you'll need to authenticate with your OVH account to validate the consumerKey
image


Almost there ...
image

lambtho12:
Well I don't get it... It seems that what you are doing does not work. 

On the other hand the change I proposed in this pull request works properly. Did you tired to proceed as I suggested with similar yaml syntax for the add-on config (and maybe just change the ami permission to be more restrictive as you suggested)?

TheGroundZero:
I copied the syntax from your documentation and used the more restrictive permissions as I proposed.

The "application" is valid and active
image

lambtho12:
Yes, but certbot cannot do the challenge and provide you the certificates. Right?

If that is the case, then the issue is with certbot itself and not this add-on I believe.

TheGroundZero:
I'll try some more and note my findings in #1427, to avoid cluttering this PR.

PR #1493, opened 2020-07-22
Allow configurable hostname for hassio mqtt calls

imduffy15:
I'm wanting to use the docker image in a home assistant environment on kubernetes. The hardcoded hostname makes this difficult to do. Adding an option so users can change it.

### DIFF
--- a/mosquitto/config.json
+++ b/mosquitto/config.json
@@ -26,7 +26,10 @@
     },
     "certfile": "fullchain.pem",
     "keyfile": "privkey.pem",
-    "require_certificate": false
+    "require_certificate": false,
+    "hassio": {
+      "hostname": "hassio"
+    }
   },
   "schema": {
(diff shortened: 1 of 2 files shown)


frenck:
This is not how it works, unfortunately. The Supervisor and the ecosystem around it are not designed to run on Kubernetes.

These hostnames are resolved and managed internally for a lot of things, not just this part you changed here, but actually almost any add-on (but also other internals) will fail.

imduffy15:
> These hostnames are resolved and managed internally for a lot of things, not just this part you changed here, but actually almost any add-on (but also other internals) will fail. So in the end, this is adding options, that actually won't solve anything.

You are 100% correct, didn't realise it was the supervisor that it was hitting and not home assistant, the wording in the log lines had me confused.
Will close this and move on.

Thanks for taking a look.

Issue #1487, opened 2020-07-21
HASS.IO addon upgrade to 6.1.1 error


andysaab:
After upgrade addon got error - deCONZ: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
![Deconz error](https://user-images.githubusercontent.com/56047524/88031555-6d96c200-cb45-11ea-9945-9da530412afa.jpg)
Restore from backup and work again sucsessefully.

frenck:
Could you provide details on your setup? Hardware? 32 or 64 bits?

andysaab:
Thx for your reply, Frenk!
My setup is Xpenology on HP Gen8
![image](https://user-images.githubusercontent.com/56047524/88038964-a50a6c00-cb4f-11ea-98e8-e3684d5869cb.png)

fleshyflat:
same problem her with 6.1.1

althaus:
I've got the same error. System is Hassio running on a Synology DS via Docker.

System health is same as andysaab's one besides HA 0.112.4 and Python 3.7.7.

og9gjz567:
I've got the same error. System is Hassio running on a Synology DS via Docker, too.
I'm currently running version 0.112.4 ;)

omersh82:
Same error for me, running hassio on Synology DS via docker.

2020-07-23 08:43:01 | stderr | [11:43:00] INFO: [32mStarting VNC server...[0m
2020-07-23 08:43:01 | stderr | [11:43:00] INFO: [32mWaiting for device...[0m
2020-07-23 08:43:01 | stderr | [08:43:00] INFO: [32mRunning the deCONZ OTA updater...[0m
2020-07-23 08:43:01 | stderr | [08:43:00] INFO: [32mRunning the IKEA OTA updater...[0m
2020-07-23 08:43:01 | stderr | [08:43:00] INFO: [32mStarting udevd...[0m
2020-07-23 08:43:00 | stderr | [08:43:00] INFO: [32mRunning the OSRAM LEdvance OTA updater...[0m
2020-07-23 08:43:00 | stdout | [services.d] done.
2020-07-23 08:42:59 | stdout | [services.d] starting services
2020-07-23 08:42:59 | stdout | [cont-init.d] done.
2020-07-23 08:42:59 | stdout | [cont-init.d] vnc.sh: exited 0.
2020-07-23 08:42:59 | stdout | [cont-init.d] vnc.sh: executing...
[...]

PR #1499, opened 2020-07-24
Fix ABI compatibility with old unsupported linux kernel on amd64

pvizeli:
Fix: #1487


pvizeli:
I'm not sure if we should merge this

frenck:
Agree this might work, but can be a tremendously head breaking if things go south

Issue #1529, opened 2020-08-03
LetsEncrypt cannot obtain certificate with RFC2136 Server (HA 0.113.2)


derekatkins:
Hi,

I am trying to use LetsEncrypt to get a certificate for a private domain behind a DNS server that I control.  I set up RFC2136 dynamic update but when I configure the LetsEncrypt plugin I get an error.  The log says:

```
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] file-structure.sh: executing... 
[cont-init.d] file-structure.sh: exited 0.
[cont-init.d] done.
[services.d] starting services
[services.d] done.
[12:05:40] INFO: Selected DNS Provider: dns-rfc2136
[12:05:40] INFO: Use propagation seconds: 60
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Plugins selected: Authenticator dns-rfc2136, Installer None
Obtaining a new certificate
Performing the following challenges:
dns-01 challenge for name.dom.ain
Cleaning up challenges
Encountered exception during recovery: 
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/certbot/_internal/auth_handler.py", line 70, in handle_authorizations
    resps = self.auth.perform(achalls)
  File "/usr/local/lib/python3.7/site-packages/certbot/plugins/dns_common.py", line 57, in perform
    self._perform(domain, validation_domain_name, validation)
  File "/usr/local/lib/python3.7/site-packages/certbot_dns_rfc2136/_internal/dns_rfc2136.py", line 76, in _perform
    self._get_rfc2136_client().add_txt_record(validation_name, validation, self.ttl)
  File "/usr/local/lib/python3.7/site-packages/certbot_dns_rfc2136/_internal/dns_rfc2136.py", line 83, in _get_rfc2136_client
    int(self.credentials.conf('port') or self.PORT),
ValueError: invalid literal for int() with base 10: 'null'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/certbot/_internal/error_handler.py", line 125, in _call_registered
    self.funcs[-1]()
[...]

PR #1532, opened 2020-08-07
Add RFC2136 documentation

derekatkins:
Fix for Issue #1529

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -302,6 +302,51 @@ dns:
 ```
 Use `ovh_endpoint: ovh-ca` for north america region.
 
+### RFC2136
+You will need to set up a server with RFC2136 (Dynamic Update) support with a TKEY (to authenticate the updates).  How to do this will vary depending on the DNS server software in use.  For Bind9, you first need to first generate an authenticate key by running
+```
+# dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST letsencrypt
+Kletsencrypt.+165+20675
+```
+The key file (Kletsencrypt.+165+20675.key in this example) looks like the following:
+```
+# cat Kletsencrypt.+165+20675.key
+letsencrypt. IN KEY 512 3 165 Cj2SJThIYZqZO39HIOA8dYryzsLT3CI+m43m3yfGfTMvpyYw5DXjn5da hokrwyLe3MTboGkloKIsT6DUcTSdEA==
+```
+You don't need to publish this; just copy the key data into your named.conf file:
+```
+key "letsencrypt" {
+  algorithm hmac-sha512;
+  secret "Cj2SJThIYZqZO39HIOA8dYryzsLT3CI+m43m3yfGfTMvpyYw5DXjn5da hokrwyLe3MTboGkloKIsT6DUcTSdEA==";
+};
+```
+And ensure you have an update policy in place in the zone that uses this key to enable update of the correct domain
+```
+   update-policy {
+      grant letsencrypt name _acme-challenge.hassio.dom.ain. txt;
+   };
+```
+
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @derekatkins, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1532).

Thanks!

letsencrypt/DOCS.md
```diff
@@ -302,6 +302,51 @@ dns:
...
+};
+```
+And ensure you have an update policy in place in the zone that uses this key to enable update of the correct domain
+```
+   update-policy {
+      grant letsencrypt name _acme-challenge.hassio.dom.ain. txt;
+   };
+```
+
+For this add-on you will need to supply all the rfc2136_* options.  Note that the rfc2136_port item is required (there is no default port in the add-on) and, most importantly, the the port number must be quoted because the add-on is expecting it to be a string and not a number.  Also be sure to copy in the key so certbot can authenticate to the DNS server.
```
pvizeli, line 330: Instead of document a bug, you can fix it in the schema ;)

derekatkins: True, I could, but this is easier, safer, and definitely more testable :)

derekatkins:
@pvizeli does closing this PR mean you're not going to merge this update to the docs?  I don't (well, didn't until last week) have any way to test changes to the HA codebase.

letsencrypt/DOCS.md
```diff
@@ -302,6 +302,51 @@ dns:
...
+```yaml
+email: your-mail@dom.ain
+domains:
+  - hassio.dom.ain
+certfile: fullchain.pem
+keyfile: privkey.pem
+challenge: dns
+dns:
+  provider: dns-rfc2136
+  rfc2136_server: dns-server.dom.ain
```
EmnaX, line 342: I couldn't get this add-on to work when I used a FQDN. I had to change the line from:
`rfc2136_server: dns-server.dom.ain`
to:
`rfc2136_server: 123.123.123.123`
Before it would work.

derekatkins: That's odd, because it works fine for me using FQDN.
My issue was that it did NOT work if I ignored the (default port 53) setting.

EmnaX: hmm, I went back and tried changing the rfc2136_server from an IP to my DNS server's FQDN and the Let's Encrypt add-on failed again producing a number of exceptions. I'm certainly no expert here so I may have something else incorrectly configured in my case. It is working great with the IP though!

letsencrypt/DOCS.md
derekatkins: @pvizeli -- coming at this with a lot of time to sit and reflect (and also understanding HA much better), another reason not to change the schema would be that it would introduce a breaking change.  If the schema says "integer" but you supply a string, wouldn't it barf on the input?  If so, that would break every current user.

letsencrypt/DOCS.md
derekatkins: What exceptions did it give you? 

I am happy to update this PR to suggest that one may need to supply the IP address in lieu of the FQDN.

EmnaX: Sure, my log from the Let's Encrypt add-on is here: https://pastebin.com/7r7JnV0V
For some reason the add-on doesn't show the start of the log so I don't see the initial exception. What I put on pastebin is everything I could see.

derekatkins: @EmnaX -- I'm sorry, but I think we need more of the log.  The initial failure is higher up, before what you posted.  :(
Are you willing to share the FQDN of your DNS server and how you had it in your config when it failed?

PR #1565, opened 2020-09-06
Add support for certbot --dry-run

samuelalexmclean:
Added support for (and documented) --dry-run.

Tweaked the final cerbot invocation to refer to a master set of certbot_arguments to reduce duplication on args like --agree-tos, but happy to drop that bit if you liked it more how it was. Will add to change log and increment version number if this looks acceptable.

Still learning my way around git, thanks for your patience :)

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -135,6 +135,21 @@ dns:
   cloudflare_api_key: 31242lk3j4ljlfdwsjf0
 ```
 
+### Staging Environment
+
+Let's Encrypt includes a number of [rate limits](https://letsencrypt.org/docs/rate-limits/) which you might run into - for example, if you fail validation too many times in a row, or request too many duplicate certificates. If you hit one of these limts, there's no way to reset it - you'll have to wait anywhere from an hour to a week to get things moving again. To help avoid that, Let's Encrypt offers a [staging environment](https://letsencrypt.org/docs/staging-environment/) for testing purposes. You can target this environment by adding an "environment: staging" entry to your yaml; we recommend this when you're getting started, especially if things don't work on your first attempt.
+
+```yaml
+email: your.email@example.com
+domains:
+  - home-assistant.io
+certfile: fullchain.pem
+keyfile: privkey.pem
+challenge: http
+dns: {}
+environment: staging
+```
+
 ### google dns challenge
 
 ```yaml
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @samuelalexmclean, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1565).

Thanks!

pvizeli (commented):
I'm fine with better arguments handling but I see no reason why we should support the stage certification. If you run into rate limit because of missing knowledge, the remote UI would be more fit for such users. Feel free to make a new PR with your nice consolidation

samuelalexmclean:
LE "highly recommend" using this env before targeting prod ([source](https://letsencrypt.org/docs/staging-environment/)) which (AIUI) is impossible without monkey patching the hassio addon today, right? If you think it's not worth the cognitive load for newcomers to have to deal with yet another arg then fair enough (although I could put in some default logic so people only had to specify it if they were trying to do dry runs...).

I still use this multiple times a week but that's because I'm trying to test and implement [other horrifying changes](https://github.com/certbot/certbot/issues/8281) for my ratchet setup - perhaps this would be better suited to my local dev branch vs the release version. How do you avoid hitting these limits when you're mucking around? Am I just a sloppy developer?

I'll split the other bits out into a separate PR if you feel strongly, would love to contribute something - really neat bit of software, thanks for all the hard work :)

letsencrypt/DOCS.md
```diff
@@ -135,6 +135,21 @@ dns:
...
 ```
 
+### Staging Environment
+
+Let's Encrypt includes a number of [rate limits](https://letsencrypt.org/docs/rate-limits/) which you might run into - for example, if you fail validation too many times in a row, or request too many duplicate certificates. If you hit one of these limts, there's no way to reset it - you'll have to wait anywhere from an hour to a week to get things moving again. To help avoid that, Let's Encrypt offers a [staging environment](https://letsencrypt.org/docs/staging-environment/) for testing purposes. You can target this environment by adding an "environment: staging" entry to your yaml; we recommend this when you're getting started, especially if things don't work on your first attempt.
+
+```yaml
+email: your.email@example.com
+domains:
+  - home-assistant.io
```
ludeeus, line 145:
```suggestion
  - example.com
```

PR #1583, opened 2020-09-17
Roll back zwave add-on to 0.5.2

MartinHjelmare:
- After the build troubles with 0.5.3 and vnc troubles with 0.6.0, we've decided to roll back the add-on to 0.5.2.
- We'll freeze updates until the current version (0.6.0) is fixed.

### DIFF
--- a/zwave/config.json
+++ b/zwave/config.json
@@ -1,6 +1,6 @@
 {
   "name": "OpenZWave",
-  "version": "0.6.0",
+  "version": "0.5.2",
   "slug": "zwave",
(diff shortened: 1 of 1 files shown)


lukewaite:
Thanks @MartinHjelmare

rgruyters:
Ah that makes sense now. I saw there was an update and this morning it was gone. Although when clicking on the changelog link you will see there is a newer version available, only you can't update to it.

Screenshot 2020-09-18 at 09 20 35

PR #1613, opened 2020-10-07
letsencrypt: Add support for manual hook scripts.

blueshiftlabs:
These are useful for smaller or self-hosted DNS providers that don't have official support in Certbot. Several of these providers already publish official hook scripts, allowing those providers to be supported without any additional changes to the add-on.

This adds `curl` and `requests` to the Dockerfile, but this should not increase the size of the container, since `curl` is already in the addon base image, and `requests` is used by Certbot itself. However, since most of the hook scripts that I found online seemed to use one or the other, it's worth calling them out explicitly.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.12.0
+
+- Add support for custom Certbot auth and cleanup hooks
+
 ## 4.11.0
 
 - Add support for Njalla DNS
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -61,6 +61,8 @@ dnsmadeeasy_secret_key: ''
 google_creds: ''
 gehirn_api_token: ''
 gehirn_api_secret: ''
+hook_script_auth: ''
+hook_script_cleanup: ''
 linode_key: ''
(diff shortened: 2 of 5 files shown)


blueshiftlabs:
Friendly ping - @pvizeli, could you please take a look at this PR?

pvizeli:
I stop to deal with files overshare. I'm fine with the code if they use the yaml file function:
i.e.
```
my_hook_script: -
   #!/bin/bash

  echo "hallo"
```

blueshiftlabs:
Is the concern here ease of use, because the user who's setting this up would need to write another file to `/share`? To that, I have a few responses:

- The idea of having an entire hook script like [this one](https://github.com/desec-utils/certbot-hook/blob/master/hook.sh) sitting as a blob in the config file seems like poor design. Debuggability would be very hard, as well.
- There's precedent for needing a file dropped into `/share` - the Google Cloud DNS support requires a credentials file to be placed there.
- Hook scripts are an advanced-user concept anyway, and for those users, using SFTP or Samba to upload an extra file wouldn't be difficult.

Or is there another concern here?

pvizeli:
> Or is there another concern here?

It comes the day which I have time to fix that on old stuff. Today, no new old-style ways are added

pvizeli:
Stale

PR #1629, opened 2020-10-18
Add action to build changed add-ons

ludeeus:
Adds GitHub action to build add-ons that has changed files in a PR.

Example run with a file changed for the ssh add-on: <https://github.com/ludeeus/hassio-addons-1/runs/1271169501>
Example run with a failed build for nginx_proxy add-on: <https://github.com/ludeeus/hassio-addons-1/actions/runs/313791749>

When the build is done, `docker images` are executed so we can catch the new size of the images.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,66 @@
+name: Build
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - master
+
+jobs:
+  init:
+    runs-on: ubuntu-latest
+    name: Initialize builds
+    outputs:
+      files: ${{ steps.files.outputs.all }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - id: files
+        uses: jitterbit/get-changed-files@v1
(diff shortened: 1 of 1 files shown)


pvizeli:
I have no clue about create workflows. I ask myself if we should convert the builder into workflow or use it like this/azure before. We can also get the shell script directly from the repository (after we renamed it) and run that on the workflow without pulling the builder. Not sure what is better for the future or  better shareable over our repositories

ludeeus:
The way I see it we have 4 options:
1) Continue with this PR as is
2) Create a new action under the home-assistant/actions repo that does the same as introduced here
3) Make home-assistant/builder be an action (while keeping the images for local builds)
4) Close this PR and do nothing

All options have trade-off's 🤷

frenck:
I think we should make hassio-builder an action, that is by far the most flexible, and also re-usable by others in an easy way.

Alternatively, a 5th option, is looking into using Docker's buildx for cross-building.

pvizeli:
`buildx` somethings if we have too much time. For now `make hassio-builder an action` is the way to go

PR #1641, opened 2020-10-24
Fix example Home Assistant configuration

Joshndroid:
When setting up a config as a button within home assistant, the original Home Assistant configuration example provides an error "required key not provided @ data['platform']"
Changing the data: line to service_data: inline with other service calls functions as expected as well as not generating the error.

### DIFF
--- a/rpc_shutdown/DOCS.md
+++ b/rpc_shutdown/DOCS.md
@@ -64,7 +64,7 @@ Use the following inside Home Assistant service call to use it:
 
 ```yaml
 service: hassio.addon_stdin
-data:
+service_data:
   addon: core_rpc_shutdown
   input: test-pc
 ```


homeassistant:
Hi @Joshndroid, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1641).

Thanks!

rpc_shutdown/DOCS.md
```diff
@@ -64,7 +64,7 @@ Use the following inside Home Assistant service call to use it:
 
 ```yaml
 service: hassio.addon_stdin
-data:
+service_data:
```
frenck, line 67: This change is incorrect, service calls in Home Assistant have a data object.

Considering your PR description, it sounds like you are trying to add something else, but that is not related to this part of the documentation.

Joshndroid:
By all means I understand where your coming from but within rpc_shutdown, setting up a button in loveui it literally did not work as described in the docs with just data, hence the service_data which does work.

Joshndroid:
Screenshot_20201025-222742

frenck:
That is the Lovelace frontend, which is not what is described in the documentation here.

Joshndroid:
Would you be interested in a pull request with a note outlining the caveat for others who end up there like I did? Thanks for the reply btw

PR #1642, opened 2020-10-24
Update firmware upgrade section for deCONZ

bverkron:
Second attempt of this PR https://github.com/home-assistant/hassio-addons/pull/1640. Cannot determine why it errors. Trying again after updating my fork from the original add-on repo

-------------
Spin off of this PR home-assistant/home-assistant.io#15306. I believe this is the right place for this information.

### DIFF
--- a/deconz/DOCS.md
+++ b/deconz/DOCS.md
@@ -89,17 +89,9 @@ settings.
 
 ## Upgrading RaspBee and ConBee firmware
 
-This add-on allows you to upgrade your firmware straight from the Phoscon
-web interface with ease.
-
-Go to "Settings -> Gateway" and click the upgrade button.
-
-However, some USB sticks (like the Aeotec Z-Wave sticks), can interfere with
-the upgrade process, causing the firmware upgrade to fail silently. If you end
-up with the same firmware version as before you started the upgrade, consider
-unplugging the other sticks and try again.
-
-If that is still not working, try [upgrading the firmware manually][manual-upgrade].
+Although the Phoscon web interface allows you to upgrade your firmware with ease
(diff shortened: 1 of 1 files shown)


frenck:
This is a bug in the phoscon application, not the add-on. A bug should be raised with Dresden instead.

bverkron:
They are aware of the issue and state in [their firmware update doc](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Update-deCONZ-manually#update-in-docker) (the one already link to from this deCONZ add-on doc) that it does not work.

“This method cannot be used for the Home-Assistant add-on and no current workaround is known. This also includes the button in Phoscon within the addon.”

They are investigating disabling / hiding the update button for environments where it doesn’t work.

Perhaps it’s best to say nothing about the button specifically and direct add-on users to the official deCONZ doc and let them decide how to update. As it is now this add-on doc gives stale advice that is detrimental to HA users if followed so it should not be left as-is.

pvizeli:
Is not true, I upgrade my firmware all-time over Phoscon 🤷

bverkron:
I suppose you’re lucky then or have something special about your setup that other users don’t. Maybe deCONZ can learn from what you have and make it work for everyone.

As it stands now the official suggestions from deCONZ is to not use it as it does not work, hence it seems silly to intentionally direct HA users to use it. Prop decision from an HA user community perspective IMO.

frenck:
That message isn't correct anymore, fixes have been made by dresden in their latest releases. If something is broken, it should be fixed. As this is third party software, you are now blaming our community for a bug that isn't in our software?

Interesting.

Please report it with them, as bugs aren't acceptable and simply should be reported until fixed.

PR #1654, opened 2020-10-29
DuckDns script read config off ipv4 and ipv6 correctly

jonasem:
Configuring an ipv4 failed for me until i removed the ` // empty`. This is already added by bashio under the hood.

### DIFF
--- a/duckdns/CHANGELOG.md
+++ b/duckdns/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.12.2
+
+- Rely on bashio to handle empty IPv4 / IPv6
+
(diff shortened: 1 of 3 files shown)


frenck (approved):

frenck:
We need to bump the version number in `config.json` and update the changelog

jonasem:
> We need to bump the version number in `config.json` and update the changelog

Okey, didnt think about that. Thank you :)

pvizeli (approved):

Kars-de-Jong:
Updating to this version causes my add-on to log "KO" all the time, see #1657.

Issue #1601, opened 2020-09-29
Duck DNS not committed to set ip address


ClearlyDazed:
Duck DNS ver 1.12.1 is set up for a static ip address (ipv4) in the configuration but the logs show it connecting to my secondary network interface card on occasion and then reconnecting to my primary interface card. This occurs a number of times throughout the day. Using HA ver 115.2

jonasem:
ipv4 hasnt been read correctly from the config before, meaning that duckdns would try to dynamically guess ip even though config is set. A fix is on the way #1659. Try again when version 1.12.3 is released (might be pretty soon)

PR #1659, opened 2020-10-29
Set blank ipv4 and ipv6 if config is unset.

### LINKED ISSUE #1657: Duck DNS returns "KO" after updating duckdns add-on to 1.12.2
## The problem
After updating to version 1.12.2 of the duckdns add-on, the log shows "KO" every 5 minutes. This means the update failed.
Note that I do not have the `ipv4` or `ipv6` settings configured (they are both unset).

## Environment

- Add-on with the issue: duckdns
- Add-on release with the issue: 1.12.2
- Last working add-on release (if known): 1.12.1
- Operating environment (OS/Supervised): OS

The configuration of the add-on:
```yaml
lets_encrypt:
  accept_terms: true
  certfile: fullchain.pem
  keyfile: privkey.pem
token: xxxxxxxxxx
domains:
  - xxx.duckdns.org  
aliases: []
seconds: 300

```

## Traceback/Error logs

```txt
[17:14:09] INFO: KO
...
[17:19:12] INFO: KO
```

## Additional information
It also logs "KO" when I use the URL option for the `ipv4` setting. Ditto for the `ipv6` setting.
It also logs "KO" when I hard code the IP address in the `ipv4` setting.
I can login just fine to duckdns.org, which shows the same token as is used in the config.
The problematic commit seems to be #1654 .

jonasem:
Fixes #1657

### DIFF
--- a/duckdns/CHANGELOG.md
+++ b/duckdns/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.12.3
+
+- Dont set IPv4 / IPv6 in duckdns request if config is unset
+
 ## 1.12.2
 
 - Rely on bashio to handle empty IPv4 / IPv6
--- a/duckdns/config.json
+++ b/duckdns/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Duck DNS",
-  "version": "1.12.2",
+  "version": "1.12.3",
   "slug": "duckdns",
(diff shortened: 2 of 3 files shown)


jonasem:
Also fixes #1601 as far as i can see.

Kars-de-Jong (approved):
Yes, looks like this might work.

pvizeli (approved):

PR #1667, opened 2020-11-03
Add HTTP configuration instructions to DuckDNS documentation

cogneato:
Re-add `http:` integration instructions.

### DIFF
--- a/duckdns/DOCS.md
+++ b/duckdns/DOCS.md
@@ -33,7 +33,7 @@ aliases: []
 seconds: 300
 ```
 
-[HTTP][HTTP] integration configuration in configuration.yaml:
+Additionally, you'll need to configure the Home Assistant Core to pick up the SSL certificates. This is done by setting the following configuration for the [HTTP][HTTP] integration configuration in your `configuration.yaml`:
 
 ```
 http:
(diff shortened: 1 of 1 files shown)


duckdns/DOCS.md
```diff
@@ -33,6 +33,14 @@ aliases: []
 seconds: 300
 ```
 
+[HTTP][HTTP] integration configuration in configuration.yaml:
```
frenck, line 36: Maybe explain a bit more about what is going on here?

```suggestion
Additionally, you'll need to configure the Home Assistant Core to pick up the SSL certificates. This is done by setting the following configuration for the [HTTP][HTTP] integration configuration in your `configuration.yaml`:
```

Additionally, I wonder, should we recommend using this even? On Supervisor enabled platforms, I would highly recommend running a proxy (e.g., the official NGinx one).

frenck (changes requested):

duckdns/DOCS.md
cogneato: I agree that nginx proxy would be preferred, but I don't know how to reconcile that with duckdns still existing as an option which users try. Since this step has been removed from the original instructions, they now first need to go to discord or the forums and ask what to do next. I thought a link to the integration would be an ok minimum to at least point them to the right place. I wasn't sure if more explanation would be too much.

frenck (approved):
Thanks, @cogneato 👍

PR #1669, opened 2020-11-04
Allow Configurator to skip git initialization

tomer-w:
In some cases, although there is .git folder in the config folder, we should have a way to skip git initialization as it might fail. When it does it completely fails the loading of the configurator.
With this change the user can specify we should skip git initialization and therefore allow the configurator to work correctly.

### DIFF
--- a/configurator/rootfs/etc/services.d/configurator/run
+++ b/configurator/rootfs/etc/services.d/configurator/run
@@ -4,8 +4,11 @@
 # ==============================================================================
 export HC_DIRSFIRST
 export HC_ENFORCE_BASEPATH
+export HC_GIT
 export HC_HASS_API_PASSWORD
 export HC_IGNORE_PATTERN
+
+bashio::log.info "Starting Configurator run script"
 
 # If any SSH key files are defined in the configuration options, add them for use by git
 if bashio::config.has_value "ssh_keys"; then
@@ -27,6 +30,7 @@ fi
 # Setup and run configurator
 HC_DIRSFIRST=$(bashio::config 'dirsfirst')
 HC_ENFORCE_BASEPATH=$(bashio::config 'enforce_basepath')
+HC_GIT=$(bashio::config 'git')
 HC_HASS_API_PASSWORD="${SUPERVISOR_TOKEN}"
 HC_IGNORE_PATTERN="$(bashio::config 'ignore_pattern | join(",")')"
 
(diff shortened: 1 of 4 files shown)


configurator/rootfs/etc/services.d/configurator/run
```diff
@@ -4,9 +4,12 @@
 # ==============================================================================
 export HC_DIRSFIRST
 export HC_ENFORCE_BASEPATH
+export HC_GIT
 export HC_HASS_API_PASSWORD
 export HC_IGNORE_PATTERN
 
+bashio::log.info "Starting Configurator run script"
+
```
frenck, line 12: This is already outputted by S6 and not needed.

```suggestion
```

frenck (changes requested):
Added a small comment.

What confuses me, is that this option is originally build to cope with installation missing `git` support at all (GitPython). This is not the case here.

So I'm kinda missing the reason to add this?!

Lastly, if added, the changelog and version number need to be adjusted.

configurator/rootfs/etc/services.d/configurator/run
tomer-w: I'm using a bit of unique setup. I have my Raspberry Pi config folder setup with remote git repository which reside on my PC. This is achieved with separate-git-dir git option. Problem is that configurator doesn't handle this too well and when initialized I get the following error:
WARNING:2020-11-06 20:28:05,602:hass_configurator.configurator:Cmd('git') failed due to: exit code(128)
  cmdline: git status --porcelain --untracked-files
  stderr: 'fatal: not a git repository: /config/C:\ttt\HomeAssistant\config\.git

This fix allows me to use configurator (which is really handy) even when it doesn't support my git setup.

BTW: The added tracing just helped a novice like to find out what is running when but I dont have any problem with removing it.

tomer-w:
I can update the changelog and version. What should it be? 5.1.1? 5.2.0?

tomer-w:
I have unique setup which use separate-git-dir git option to have may raspberry pi config folder separated from my .git folder which reside on my PC. When I use this setup I get the following error:
WARNING:2020-11-06 20:28:05,602:hass_configurator.configurator:Cmd('git') failed due to: exit code(128)
  cmdline: git status --porcelain --untracked-files
  stderr: 'fatal: not a git repository: /config/C:\ttt\HomeAssistant\config\.git

I really like configurator so I thought that disabling git is going to be handy for me and other users which might have this or other git issues but still want to use it.

frenck:
Right, that can indeed happen on some other occasions as well, so from that perspective, it makes sense.

It should be 5.2.0, as it adds a new feature.

frenck (approved):
Thanks, @tomer-w 👍

PR #1674, opened 2020-11-06
OpenZWave 0.6.1

gongtao0607:
* Fix novnc "missing vnc_lite.html" build issue on 0.6.0.
* Map /config folder.
* Make config-dir and user-dir configurable.

### DIFF
--- a/zwave/rootfs/etc/services.d/zwave/run
+++ b/zwave/rootfs/etc/services.d/zwave/run
@@ -10,6 +10,7 @@ export OZW_NETWORK_KEY
 
 MQTT_PASSWORD="$(bashio::jq /data/auth.json '.ozw_password')"
 OZW_CONFIG=/data/ozw/config
+OZW_USER=/data/ozw/config
 OZW_DEVICE=$(bashio::config 'device')
 OZW_INSTANCE=1
 OZW_NETWORK_KEY=$(bashio::config 'network_key')
@@ -17,6 +18,14 @@ OZW_NETWORK_KEY=$(bashio::config 'network_key')
 # Set custom instance ID if configured
 if bashio::config.has_value 'instance'; then
     OZW_INSTANCE=$(bashio::config 'instance')
+fi
+
+# Set custom config_dir and user_dir
+if bashio::config.has_value 'config_dir'; then
+    OZW_CONFIG=$(bashio::config 'config_dir')
+fi
+if bashio::config.has_value 'user_dir'; then
+    OZW_USER=$(bashio::config 'user_dir')
 fi
 
 # Wait until mosqitto is up and running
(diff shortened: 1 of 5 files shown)


homeassistant:
Hi @gongtao0607, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;hassio-addons#1674).

Thanks!

zwave/rootfs/etc/services.d/zwave/run
```diff
@@ -19,6 +20,14 @@ if bashio::config.has_value 'instance'; then
     OZW_INSTANCE=$(bashio::config 'instance')
 fi
 
+# Set custom config_dir and user_dir
+if bashio::config.has_value 'config_dir'; then
+    OZW_CONFIG=$(bashio::config 'config_dir')
+fi
+if bashio::config.has_value 'user_dir'; then
+    OZW_USER=$(bashio::config 'user_dir')
+fi
```
frenck, line 29: We should not do this.

We don't want this into the Home Assistant configuration folder. 

This feature isn't implemented yet, because it awaits:

https://github.com/home-assistant/supervisor/issues/1842

frenck (changes requested):

zwave/rootfs/etc/services.d/zwave/run
gongtao0607: It by default writes to "/data/ozw/config". It only changes when user wants to. 
If /config can cause conflict with HA, can we let user chose to use /share folder? (It is the user repurposing the folder, not the addon)

frenck: Well, the same answer as before, this feature isn't implemented because of the linked RFC.

gongtao0607: I'm not sure I understand what you said unimplemented "feature".
The config.json -> map allows me to access these folders. If I map `config` and/or `share` in config.json, user can set the flags here to ask ozwdaemon to use those folders, so the other addons (e.g., samba, ssh) can access them.

gongtao0607: For feasibility, I just want to clarify that this is tested working - I am testing my zwave device support xml in /config

frenck: This feature wasn't added because of the linked issue/RFC. This is not possible at this time. We don't want this configuration to end up in these folders you are suggesting; nor do we want this configurable.

As said, this feature isn't added before because of the given reasoning. Please remove that feature from this PR.

pvizeli:
The novnc fix is fine, it allows us to provide new versions. Please remove the other changes until the supervisor implemented the linked RFC.

deltachonglee81 (approved):

deltachonglee81 (changes requested):
Amazing

PR #1681, opened 2020-11-15
MQTT documentation: clarifying write access for anonymous users

coreGreenberet:
I've had issues with enabling mqtt for a device. From the current documentation I understood, that anonymous: true should be enough, since I didn't specify any logins


pvizeli:
Let's fix that and automatic generate such config on startup if this option is enabled

PR #1687, opened 2020-11-20
Simplify devcontainer setup

ludeeus:
The same setup as was added to https://github.com/home-assistant/addons-development

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,47 +1,25 @@
-FROM mcr.microsoft.com/vscode/devcontainers/base:debian-10
+FROM mcr.microsoft.com/vscode/devcontainers/base:debian
 
 WORKDIR /workspaces
 
-# Default ENV
-ENV LANG C.UTF-8
-ENV DEBIAN_FRONTEND noninteractive
-
-# Set shell
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
-# Install docker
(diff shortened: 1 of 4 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,47 +1,25 @@
...
         libpulse0 \
-        git \
-        tar \
-        xz-utils \
-    && rm -rf /var/lib/apt/lists/* \
-    \
-    && curl -SL "https://storage.googleapis.com/shellcheck/shellcheck-v${SHELLCHECK_VERSION}.linux.x86_64.tar.xz" | tar xJf - \
-    && cp "shellcheck-v${SHELLCHECK_VERSION}/shellcheck" /usr/bin/ \
-    && rm -rf shellcheck-v${SHELLCHECK_VERSION}
+        shellcheck
```
pvizeli, line 19: How old is the shellcheck?

ludeeus: Not that bad, it's from march (0.5.0)

pvizeli: Do we not want to be close as possible to CI?

frenck: Alternatively, there are statically linked versions available on their github releases page...

<https://github.com/koalaman/shellcheck/releases>

ludeeus: Yeah, if not using the one in apt, we should use 'stable'

pvizeli (approved):

PR #1726, opened 2020-12-21
[homematic] Add troubleshoot coprocessor firmware to docs

FloThinksPi:
Add a section to docs with errors related to a custom HM-MOD-RPI-PCB firmware version.
In this case one needs to flash firmware from occu otherwise one gets strange errors in logs.
This e.g. happens when having used the HM-MOD-RPI-PCB before on e.g. raspmatic.

### DIFF
--- a/homematic/DOCS.md
+++ b/homematic/DOCS.md
@@ -132,6 +132,21 @@ configuration. If you installed Hassio on another distribution of Linux, you nee
 follow the installation guide for the UART USB to setup the UART USB interface on
 your computer.
 
+## BidCos Errors / Coming from raspmatic/pivccu ?
+
+It is recommended to flash the HM-MOD-RPI-PCB again as raspmatic uses a modified firmware as far as we know.
+This causes strange issues with BidCos init
+The logs look simmilar to:
+- CCU2CoprocessorCommand::CCU2CoprocessorCommand(): Message too small: 2 byte
+- CCU2CommController::handleIncomingSerialFrame(): Command not parseable. 
+- CCU2CommController::waitForCoProcessorResponse(): Timeout while waiting for response.
+- <Error> No BidCoS-Interface available
+  
+If you have those kind of issues and your module was never used with raspmatic you can try this anyway.
+  
+To flash the coprocessor firmware one can e.g follow this guide(works from within hassos):
+https://gist.github.com/FloThinksPi/a2cc6e0e8d106ca9e6378c8c6c61ee67
+
 ## Support
 
 Got questions?


homeassistant:
Hi @FloThinksPi, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#1726).

Thanks!

dave-castle:
@FloThinksPi was trying to move my setup to just one Pi today and running also in the same error that bidcos isn’t available. Current setup is pivccu, I think/hope that with the firmware upgrade you mentioned it will be fixed. Honestly I’m little bit afraid when updating and switching back to old so card my running setup will be destroyed. Any suggestion how to switch back to old FW?

FloThinksPi:
@dave-castle as far as i know the firmware on the radio module is just for sending out messages and does not hold any information about your setup. It is just used as a relay for homematic to send out commands and recieve information. As long as you backup and restore the CCU and restore it in the addon your setup should be keept.
Flashing can be done according to the insturctions i added as a link in the PR: https://gist.github.com/FloThinksPi/a2cc6e0e8d106ca9e6378c8c6c61ee67 

There are many ways to do it but this one is specifically usefull in a HasOS setup where one does not have a full Operating system to install stuff into.

pvizeli:
Can we not integrate the flashing on the add-on?

FloThinksPi:
@pvizeli  Hm it would be possible theoretically. We would need to include the flash utilitz in the image and call it on startup. Firmware file should be already present as it is part of occu.  However i dont know if you can brick the modules somehow by aborting the flash process midway(e.g. cutting power etc.) and i dont want to test this on my unit. Since it is uart i would not guess so but i am not sure.

When we do it automaticallz and the user cannot know if a flash is currently running we should be sure nothing bad can happen when he pulls the plug.

homematic/DOCS.md
```diff
@@ -132,6 +132,21 @@ configuration. If you installed Hassio on another distribution of Linux, you nee
 follow the installation guide for the UART USB to setup the UART USB interface on
 your computer.
 
+## BidCos Errors / Coming from raspmatic/pivccu ?
+
+It is recommended to flash the HM-MOD-RPI-PCB again as raspmatic uses a modified firmware as far as we know.
```
jens-maus, line 137: Sorry, but this is not correct. RaspberryMatic as well as piVCCU uses the latest official firmware for the HM-MOD-RPI-PCB which you can find here: https://github.com/eq-3/occu/tree/master/firmware/HM-MOD-UART

And this is version 2.8.6 (`dualcopro_si1002_update_blhm.eq3`) which enables the HM-MOD-RPI-PCB to communicate with HomeMatic (BidCos-RF) as well as homematicIP devices. So, by referring users to force install version 1.4.1 (`coprocessor_update.eq3`) of the HM-MOD-RPI-PCB the rf module will not be able to communicate with homematicIP devices anymore. This should at least mentioned here.

FloThinksPi: oh good to know @jens-maus. This instruction was just based on pure observations when using the module with RaspberryMatic it wont work with the occu plugin anymore unless the coprocessor_update.eq3 is flashed. I did not try dualcopro_si1002_update_blhm.eq3 as i didnt know about that one. Think ill try that out, maybe this plugin somehow cannot work with that firmware.

PR #1730, opened 2020-12-22
nginx_proxy: Add a new "listen_parameters" config option

albinou:
This new config option allows to change the default `listen` arguments
used in nginx configuration. In particular, this allows to use the
`proxy_protocol` to retrieve the end-user IP address (along with a
config file stored into a `customize.default` file).
Default options are not changed by this commit.

I made this change generic, but it would also make sense to add a `proxy_protocol` config option directly. Discussion is open :-)

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -33,6 +33,7 @@ domain: home.example.com
 certfile: fullchain.pem
 keyfile: privkey.pem
 hsts: "max-age=31536000; includeSubDomains"
+listen_parameters: http2
 customize:
   active: false
   default: "nginx_proxy_default*.conf"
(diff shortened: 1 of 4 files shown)


homeassistant:
Hi @albinou, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#1730).

Thanks!

frenck:
So, this is rather an advanced use case, that probably 99% of the users of this add-on won't need. As in, it never came up in all the years this add-on exists.

Why add this as an option? As for advanced configurations, the alternative to use your own configuration files exists as well.

albinou:
Hi frenck,

Thanks for your comment. I totally agree that my use case (using the proxy_protocol) is advanced. That is why I did not wanted to mention the protocol name in the config option I added in the proposed commit.

However, it seems I can't do what I want by just using the `customize.active` option. In for my setup I need these 3 nginx config lines:

```
listen 443 ssl http2 proxy_protocol;
set_real_ip_from 192.168.1.1;
real_ip_header proxy_protocol;
```

But, as far as I know, the default listen instruction (i.e. `listen 443 ssl http2;`) can't be overwritten. Indeed, adding a line like `listen 443 ssl http2 proxy_protocol;` into a file  `/share/nginx_proxy_default*.conf`  prevents nginx from starting: logs says that `listen` can't be used twice.

So, my patch provides a way to change de default `listen` options and I use it in addition to a personal configuration file.

Is there a better solution for me?
Thanks.

PR #1748, opened 2021-01-04
Zwave: update ozw 66619ed

pvizeli:
- Update xml definition of ozw
- Typo fix on ozw-admin

### DIFF
--- a/zwave/CHANGELOG.md
+++ b/zwave/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.8.0
+
+- Update OpenZWave to 66619ed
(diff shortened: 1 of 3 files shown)


frenck (approved):

MartinHjelmare:
I'm building this and will test after that.

MartinHjelmare:
I've tested and my devices works for me with the integration. I had some issues with ozw-admin in Chrome, but it doesn't seem consistent. During some time it just said loading in the top panel of the VNC frame. But after a restart and some browser refreshes it now loads the window successfully. In Firefox it seems to work.

MartinHjelmare (approved):

PR #1772, opened 2021-01-24
Turn off NGINX server tokens by default

jumping2000:
SECURITY BEST PRACTICE:
add "server_tokens off" option to not display the version of nginx that you use

### DIFF
--- a/nginx_proxy/data/nginx.conf
+++ b/nginx_proxy/data/nginx.conf
@@ -12,6 +12,7 @@ http {
         ''      close;
     }
     
+    server_tokens off;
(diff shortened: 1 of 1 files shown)


frenck:
No need to all caps that.

jhampson-dbre:
This is included in the nginx.conf that is shipped with the Alpine [nginx package](https://gitlab.alpinelinux.org/alpine/aports/-/blob/3.11-stable/main/nginx/nginx.conf#L35)

frenck (changes requested):
Looks fine, please update the add-on changelog and version information as well.

PR #1775, opened 2021-01-24
Add ModSecurity web application firewall to NGINX Home Assistant SSL proxy addon

jhampson-dbre:
## Feature Proposal: Add ModSecurity web application firewall to NGINX Home Assistant SSL Proxy addon

[ModSecurity](https://github.com/SpiderLabs/ModSecurity) is an open source web application firewall (WAF). When used with the [OWASP ModSecurity Core Rule Set (CRS)](https://github.com/coreruleset/coreruleset) - an open source firewall policy for ModSecurity, web application are protected from a wide range of attacks, including SQL Injection (SQLi), Cross Site Scripting (XSS), Local File Inclusion (LFI), Remote File Inclusion (RFI), and Code/Shell Injection.

From my understanding of the recent Home Assistant security disclosures, the types of vulnerabilities that were found in 3rd party custom integrations are exactly the types of attacks that ModSecurity is designed to provide a layer of protection against.

Including ModSecurity in the NGINX SSL Proxy add-on would provide an easy to adopt layer of security for Home Assistant users with internet-facing installations, while only requiring minimal user configuration.

To use ModSecurity in NGINX SSL Proxy, one new configuration option is implemented:

- `security_mode`: Controls the behavior of ModSecurity web application firewall. Valid values are:
  1. `DetectionOnly` - (default) process security rules and log detections but never executes any disruptive actions (block, deny, drop)
  2. `On` - process security rules; blocks potentially malicious requests
  3. `Off` - do not process security rules

The ability easily run in "report only" mode or completely disable ModSecurity greatly reduces the user impact in case of any false positive detections or other unforeseen issues occur.

## Additional supportability considerations

1. ModSecurity is compiled from source as an nginx "dynamic module", making the initial installation a fairly complex process (compared to a simple package manager installation).
   - ModSecurity must be compiled with the source code for the version of nginx that is installed (v1.16.1 in the current NGINX SSL Proxy addon)
   - Currently the NGINX SSL Proxy addon does not specify a particular nginx version to install. If/when the version of nginx changes, ModSecurity library would need to be compiled against that version.
2. I observed that the compile process for GitHub Actions is extremely slow for ARM-based images. The amd64 build completes in ~15 minutes, but aarch64 takes around 1 hours and 45 minutes. Searching google indicates that this is a common problem.

My current solution is to use a 2-stage build:
 - Stage 1 - Build an intermediate container from Dockerfile.modsecurity that contains the compiled modsecurity libraries. ~~Although currently configured to pull from the official `nginx` repository on Docker Hub, the Home Assistant base images could be used here as well.~~ Edit: Updated `modsecurity/Dockerfile` to use Home Assistant base images.
       Stage 1 would only need to be ran: 
       - to update modsecurity to new releases
       - to support newer nginx version that are introduced in Stage 2
 - Stage 2 - Build the NGINX SSL Proxy addon and copy the compiled libraries from the Stage 1 build.   Currently, the original Dockerfile now points to my (~~yet-to-published~~ Edit: modsecurity library build images are published now) Stage 1 build, but were this to be included in the core NGINX addon, it would be pointed to the images built by the Home Assistant project.
       - Stage 2 would pull the prebuilt Stage 1 copy as part of the build process to elimate the need to compile the binaries for every build.

I'm open for any other solutions to resolving the slow compiles on ARM builds.

3. The default configuration of OWASP CRS "should face [false positives] rarely, and therefore it is recommended for all sites and applications", however, if any false positives are encountered, they would need to be addressed by creating an exclusion policy and/or filing an issue in the CRS repository.

## Alternatives considered for securing Home Assistant with ModSecurity

A community fork of the NGINX SSL Proxy addon could be created that includes ModSecurity. While a fork would provide the community with an additional option, ModSecurity should be considered for inclusion in the core NGINX Home Assistant SSL proxy addon because:
1. Anyone running the core NGINX Home Assistant SSL proxy addon to expose their Home Assistant to the Internet could benefit from the additional protection provided by ModSecurity WAF.
2. Anyone who does not want the protection provided by ModSecurity or encounters any problems due to false positives can disable the protection via configuration and continue to use the existing functionality.
3. While recently disclosed security vulnerabilities have been mitigated, ModSecurity would provide one more security layer to guard against yet-to-be-discovered vulnerabilities.

Thanks for taking the time to review this proposal. Looking forward to any feedback.

Ref: [Compiling and Installing ModSecurity for NGINX Open Source](https://www.nginx.com/blog/compiling-and-installing-modsecurity-for-open-source-nginx/)


frenck:
Please make sure the build passes, thanks 👍

jhampson-dbre:
This pull request is now at "minimum viable" levels of readiness. While there are no configuration options added yet to enable/disable ModSecurity, I am currently running this as a local addon on my personal setup and I have validated that the the ModSecurity module is loaded, running in `DetectOnly` mode, and it is detecting and logging "simulated" attacks. For example, the User-Agent 'Nikto' is used by a web server vulnerability scanner software and is detected as malicious by ModSecurity.

```
curl -H "User-Agent: Nikto" https://<my-personal-setup>.duckdns.org
```

```
# modsecurity_audit.log
...
ModSecurity: Warning. Matched "Operator `PmFromFile' with parameter `scanners-user-agents.data' against variable `REQUEST_HEADERS:user-agent' (Value: `Nikto' ) [file "/usr/local/owasp-modsecurity-crs/rules/REQUEST-913-SCANNER-DETECTION.conf"] [line "33"] [id "913100"] [rev ""] [msg "Found User-Agent associated with security scanner"] [data "Matched Data: nikto found within REQUEST_HEADERS:user-agent: nikto"] [severity "2"] [ver "OWASP_CRS/3.3.1"]
...
```

I'll continue to run this on my personal setup, but I think the concept is far enough along that it can be evaluated so I removed the WIP indicator from the title.

DevSecNinja:
Great feature! I hope to see it in the Nginx Proxy Manager add-on some day as well. Someone already created a feature request for it: https://github.com/jc21/nginx-proxy-manager/issues/847

jhampson-dbre:
Added a `security_mode` option that controls if modsecurity policy is being enforced, audited, or disabled. Currently defaults to `DetectionOnly` to minimize any potential user impact and requiring opt-in to start blocking requests detected as malicious.

@frenck I don't have any other planned changes to make for this. Not sure if the "in progress" label needs to be removed or anything. Thanks.

norgan:
Brilliant, I was just coming to github to suggest this, and here it is, you've basically nailed it. Thanks

pvizeli:
We can't allow copying a binary from a 3party into our container. Please build the module inside our dockerfile

nginx_proxy/Dockerfile
```diff
@@ -1,7 +1,96 @@
...
+    git clone -b master --depth 1 https://github.com/SpiderLabs/ModSecurity-nginx.git && \
+    git clone -b ${OWASP_BRANCH} --depth 1 https://github.com/coreruleset/coreruleset.git /usr/local/owasp-modsecurity-crs && \
+    wget -O - https://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz | tar -xz
+
+# Install ModSecurity Nginx modules
+# Since Alpine nginx package is not compiled using
+# `--with-compat` flag, we must specify all the same
+# compile options that are used to compile the nginx Alpine
+# packge when we configure the modsecurity nginx connector
+# https://git.alpinelinux.org/aports/tree/main/nginx/APKBUILD?h=3.11-stable
```
joelmoses, line 69: Oh, good grief. That's a pain. Any concerns about keeping in-sync with potential changes in the base build?

nginx_proxy/config.json
```diff
@@ -34,7 +35,8 @@
       "active": "bool",
       "default": "str",
       "servers": "str"
-    }
+    },
+    "security_mode": "str"
```
joelmoses: I'd suggest doing some basic schema validation here since you're only using three options (On, Off, DetectionOnly), otherwise you may end up with some unparsable configuration in the modsecurity.conf file later.

"security_mode" : "match(On|Off|DetectionOnly)"
or
"security_mode" : "list(On|Off|DetectionOnly)"

I don't know much about Modsec's config case-sensitivity, so you may need to do "match" with a regex to handle case-insensitivity if you want to allow that.

nginx_proxy/data/nginx.conf
```diff
@@ -17,6 +17,10 @@ http {
 
     server_names_hash_bucket_size 64;
 
+    # enable modsecurity
+    modsecurity on;
+    modsecurity_rules_file /etc/nginx/modsec/main.conf;
```
joelmoses, line 22: Might I suggest only including this section if the user has actually turned "On" or "DetectionOnly" mode? You can either "sed" it into the nginx.conf file or set it using tempio.

joelmoses (commented):
Basic review

nginx_proxy/Dockerfile
jhampson-dbre: I've open an [issue in the Alpine repo](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12375) to request it be added upstream. This "should" only be a concern, if the nginx version in the addon is updated before the `--with-compat` flag is added upstream (assuming it will be added at some point). It really is a 1 line change to add `--with-compat` with the rest of the existing options, so I'm hoping it won't take too much time for the Alpine maintainer for the nginx package to evaluate adding it.

I've seen some people dynamically determine the required compile options by installing nginx from package manager, dumping the compile-time options, `sed` out the stuff that's not needed, and then add those options to the `./configure` command, but that seems like a workaround for something that can/should be fixed upstream.

nginx_proxy/config.json
jhampson-dbre: This should already be updated to `"security_mode": "list(DetectionOnly|On|Off)"`. In an older commit it was `"security_mode": "str"`, so maybe an older revision got selected during the review.

nginx_proxy/Dockerfile
joelmoses: I understand. It's as good as we can hope right now. Luckily, we can control our destiny somewhat with build.json here so an "unprompted" upstream change shouldn't impact us.

I'm curious to see whether that issue gets picked up.

nginx_proxy/data/nginx.conf
jhampson-dbre: Having `modsecurity on;` only causes the modsecurity module to be loaded by nginx,  but the `SecRuleEngine` setting in modsecurity.conf is what actually determines if any rules are being processed by modsecurity or not. 

If we leave `modsecurity on;` with `security_mode: Off`, the net effect is the same as if we remove the modsecurity entries entirely from nginx.conf: modsecurity rules are not processed.

However, with `modsecurity on;`, there is the side effect that the modsecurity module is loaded by nginx, even though  it may or may not be processing rules depending on the value of `security_mode` option. 

My question would be is there is any concern with nginx loading the modsecurity module if modsecurity is not actually processing rules (i.e. `security_mode: Off`)?

P.S. Thanks for taking the time to review this. I appreciate the feedback and discussion.

joelmoses: Because we're running in a state where we're restarting the container each time with a newly built configuration, the only real cost is system memory to only turn on the module when it's in On or DetectionOnly mode. The purist in me doesn't like enabling a module in a configuration and not using it... but I'm used to working in embedded systems. :)

Alternately, maybe adopting an approach similar to the "customize" section...

security:
  active: true
  mode: (On|DetectionOnly[default])

When the section is set active, drop in the modsecurity stanza into nginx.conf and it adopts DetectionOnly with no "mode" set. Otherwise, the user can define the mode manually. This approach would also allow you to customize modsecurity later (custom rule folders? logging destinations?).

nginx_proxy/config.json
joelmoses: I see it now. Yeah, I think I might have been in an older commit by mistake. Disregard.

nginx_proxy/data/nginx.conf
jhampson-dbre: I like the flexibility this would add to allow for future customizations if needed, with out really adding much complexity initially. Plus the added benefit of not loading modsecurity into nginx at all if it's not being used. Probably, would only have a small reduction in the memory footprint, but should eliminate any possibility of unintended side effects from modsecurity when it is not used.

jhampson-dbre: Updated to use this format and tested good on my local. Thanks for the suggestion.

nginx_proxy/Dockerfile
jhampson-dbre: I really don't see why it wouldn't. I attached a link to a similar request for the Debian nginx package that has already been accepted and merged so hopefully that will help speed the consideration to see that it has already been implemented in other distros, including the code change required to implement it. The Alpine maintainer for the nginx package seems pretty active and there is usually a couple of commits added about every month or two. I'll open a PR for it if it doesn't get picked up before I get a chance to.

nginx_proxy/data/nginx.conf
joelmoses: Looks great! I'd imagine that it also means it's safer for @frenck to ship as it'll basically run like the "normal" addon by default and only load and engage when the user is ready to configure.

I note that you've got three "exception" rules incorporated (I too had figured Node-RED would be a difficult one). How many addons/integrations have you tested with so far?

jhampson-dbre: Good call out. A couple thoughts around this before I answer your question:

1. It would be impossible to find every false positive that might be detected by ModSecurity while using Home Assistant due to the hundreds of "official" integrations, plus essentially unlimited community content. Really the only things that potentially **need** an exception would be official integrations and add-ons that **require** remote access to function (like, Google Assistant, Alexa, etc).
2. Even then, this an exception is only absolutely required if all of the following are true:
   - Built-in, integration, or add-on functionality is detected as a false positive by the default CRS ruleset
   - The integration or add-on requires continuous remote access to function
3. If there is no false positive, then there is no exception required
4. If there is a false positive, and the integration/add-on doesn't require remote access, or only needs one-off remote access (for example, initial registration only, after which it functions locally) i've added some recommendations for local workarounds to DOCS.md
```md
   1. Avoid requests being processed through ModSecurity by accessing Home Assistant using the internal connection URL (e.g. homeassistant.local:8123), then retry the blocked action.
   2. Temporarily disable ModSecurity by changing `security.mode` to `DetectionOnly`, restart the NGINX Home Assistant SSL Proxy add-on, then retry the blocked action. Afterwards, re-enable ModSecurity by changing `security.mode` to `On` and restart the add-on again.
 ```
5. Based on that criteria, of the 3 exceptions I added, the only one that I think is **required** is `SQLi false positive for wildcards in add-on configurations`. This is because you could end up in a situation where you only have remote access to your system, enable ModSecurity, and then cannot disable it because the configuration options get detected SQLi false positive, so the configuration change fails to save, making both workarounds unavailable.
6. The other 2 exceptions I added more due to the "annoyance factor", but maybe they should be removed to not set a precedent?
   - For Node Red: I usually connect using homeassistant.local, but for some reason I was connected to my external address (probably because I was working on this addon). I made some changes to a flow, went to deploy, deploy got blocked by ModSecurity (because your literally POSTing code in the payload, so not unexpected to get flagged). Go to a new browser tab to configure the settings to put ModSecurity in DetectionOnly mode, restart the add-on. Node Red tab loses connection due to nginx addon restart, and when it reconnects, the page reloads and lost my flow changes. Node Red is a "Community Add-on" and does not require remote access to function, so it does not meet my previously stated criteria for required exceptions, so maybe this exclusion should be removed? Or does the higher impact factor in, since you can't simply retry the request without redoing your work?
   - For "Allow integrations to be deleted": This is built-in functionality, so the first criteria of "built-in functionality" is met, but deleting an integration does not require remote access and can be done through local connection, or easily retried after temporarily disabling ModSecurity. I added this one because there is no error message when deleting the integration is blocked; it fails silently. So it took me a few times of hitting the "delete" button and nothing happening before I went to check the NGINX add-on logs to find the block message there.
7. Both of those problems, they can be easily avoided by operating primarily over the local connection, so my recommendation would be if you want to be able have 100% functionality to remotely manage Home Assistant over remote web access at anytime or any place, then you should leave ModSecurity disabled, which would make you no worse off than your currently are. Alternatively, use a VPN connection for most remote management, and use the remote web access for only the integrations/add-ons that require it.
8. However, my impression is that most people only use remote access for minimum necessary functionality of integrations require it. And for those people, they could opt-in to an additional layer of security without any impact to functionality.
9. As you can see from the comment in the exlusions file 'User specified exclusions can be added by using the `security_exclusions_file`', I got a little carried away since there is no `security_exclusions_file` option, and I think 99.9% of people wouldn't write their own ModSecurity rules even if it was an option. But if it turns out this customization is needed, it will be easier to implement due to the adjusts we made to the add-on options. 🎉

With all that being said I can only test remote access for integrations that I use, but I have no false positives with them from ModSecurity:
   - Mobile App
   - SmartThings
   - Withings

I think SmartThings and Withings just use the remote access for external webhook callback, and I think integrations that follow this same behavior have low risk of false positives.

Would be great if somebody could test with the cloud-based assistants, e.g. Google, Alexa. Those are the ones I would think have a higher potential for false positives where a workaround would not be suitable.
I could create an account and configure an integration for those, but I don't know that just setting up the integration is a great test for these without actively using them.

Comment (joelmoses) [reply nginx_proxy/data/nginx.conf]: Good observations. I'm obviously not saying it's possible to cover every possible integration, but it might be good to cover the HA addons that are part of the Core at the very least - the ones that use the _Ingress_ functions to present their own WebUI. Those are the most-often-used ones and might generate issues. In particular, I'm thinking of the Zwave addons and maybe the Conbee one.

I also agree that very few people are going to want to make their own exclusions, but it's probably something that will occur with advanced users. You might need to cross that bridge (the "security_exclusions" one) sooner than you think. I'd suggest adopting the same approach as the NGINX customizations one in this same addon - but obviously, that can come later.

I use the Nobu Casa functionality for Alexa/Google Assistant functionality which does not require direct external access (uses SNITun) so I'm not able to test the cloud-based assistants. But that's a good point. 

That said, this is an awesome addition to the NGINX addon. I hope we see it integrated.

Comment (jhampson-dbre) [reply nginx_proxy/data/nginx.conf]: I think there is an argument that could be made for disabling ModSecurity for ingress, since it's protected by Home Assistants authentication. And if a bad actor is able to authenticate to your Home Assistant instance, then ModSecurity isn't going to help you at that point. On the other hand, it's nice to have it as a layer of security, because you never know when a vulnerability could be found.

As far as maintainability, I could see a community currated exclusions list that supports non-core functionality and advanced use cases that people could optionally include in their own instances. then only the minimum necessary (based on the criteria above) ship with the addon.

The key thing to keep on mind is that it would be opt-in. And if you do opt-in and you find it too disruptive for your particular set of integrations, add-ons, and remote management preference, you can opt back out and be no worse off than you are currently.

File: nginx_proxy/Dockerfile (hunk 1 above)
Comment (jhampson-dbre) [reply nginx_proxy/Dockerfile:69]: looks someone has opened merge request in the Alpine repo to add --with-compat flag. are we good to go ahead and resolve this thread and the other one?

Comment (joelmoses) [reply nginx_proxy/Dockerfile:69]: Sure thing, although it'll be some time before this change gets pushed down to a home assistant base-image.

File: nginx_proxy/modsecurity/Dockerfile [hunk 4]
```diff
@@ -0,0 +1,93 @@
+ARG BUILD_FROM
```
pvizeli, line 1: Not sure why that is needed as folder: `nginx_proxy/modsecurity`

nginx_proxy/Dockerfile
```diff
@@ -1,11 +1,120 @@
 ARG BUILD_FROM
+
+FROM ${BUILD_FROM} as build_modsecurity
+
+ARG NGINX_VERSION=1.16.1
+ARG MODSEC_BRANCH=v3/master
+ARG OWASP_BRANCH=v3.3/master
```
pvizeli, line 7: Args values are offloaded into `build.json`

nginx_proxy/modsecurity/Dockerfile
jhampson-dbre: We moved everything to the main `nginx_proxy/Dockerfile` so the `nginx_proxy/modsecurity` folder is obsolete. It's deleted now.

nginx_proxy/Dockerfile
jhampson-dbre: Thanks for the tip. done.

nginx_proxy/Dockerfile
```diff
@@ -1,11 +1,120 @@
...
+# Copy owasp-modsecurity-crs and modsecurity from the build image
+COPY --from=build_modsecurity /usr/local/modsecurity /usr/local/modsecurity
+COPY --from=build_modsecurity /usr/local/owasp-modsecurity-crs /usr/local/owasp-modsecurity-crs
+COPY --from=build_modsecurity /usr/lib/nginx/modules/ /usr/lib/nginx/modules/
 
 # Copy data
 COPY data/run.sh /
 COPY data/nginx.conf /etc/
+COPY data/modsec/ /etc/nginx/modsec/
+COPY data/owasp/ /usr/local/owasp-modsecurity-crs/
```
pvizeli, line 118: Please create a `rootfs` like i.e. the ssh add-on or zwavejs. We moving to s6-overlay if we touch an add-on.

jhampson-dbre: Ok. The files in `data/` are now restructured under `rootfs/` to match the expected directory paths inside the container.

mainmind83:
🧐

PR #1776, opened 2021-01-24
Use ${MARIADB_DATA} everywhere in run script

lmagyar:
At 2 places the default ```/data/databases``` is hard coded, and not the ```${MARIADB_DATA}``` variable is used.

The fix is tested, works.


frenck (approved):
Removed the bug marker from the title, as it is a good chance, however, it does not cause issues and thus not a bug.

Thanks 👍

PR #1777, opened 2021-01-24
[Typo] [MariaDB] Repair Example Home Assistant configuration

lmagyar:
At least in OS (5.10) environment, the correct host name is ```local-mariadb``` and not ```core-mariadb```.

Though in Supervisor log this has _ (underscore). But my knowledge is way insufficient to figure out this second problem.
```text
21-01-23 20:18:02 INFO (MainThread) [supervisor.services.modules.mysql] Set local_mariadb as service provider for MySQL
```

### DIFF
--- a/mariadb/DOCS.md
+++ b/mariadb/DOCS.md
@@ -69,7 +69,7 @@ Example Home Assistant configuration:
 
 ```yaml
 recorder:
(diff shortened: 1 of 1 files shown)


frenck:
This change is incorrect. It is only local if you have your own local fork/creation. Normal pre-build and stable distribution should be `core-mariadb`.

Closing this PR.

PR #1780, opened 2021-01-25
Fix and clean zwave_js add-on

MartinHjelmare:
- Fix typos including discovery item and config file type.
- Fix driver storage by removing that config setting. It was causing an error during runtime.
- Clean white space.


pvizeli (approved):

pvizeli:
Thanks

PR #1789, opened 2021-01-29
Use hostname for deconz discovery host instead of IP

raman325:
Using an IP for discovery is temperamental because the container can pick up a new IP on restart. Using a hostname will ensure that HA is always able to connect to the add-on. I noticed this issue because I lost connectivity to my deconz add-on due to the IP change issue.

I originally opened https://github.com/home-assistant/core/pull/45674 but I think fixing it on the add-on side is better

### DIFF
--- a/deconz/CHANGELOG.md
+++ b/deconz/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.6.3
+
+- Use hostname for discovery instead of IP
+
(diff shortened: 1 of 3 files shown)


frenck (approved):
Yes!

This was actually changed quite some time ago to deal with another issue we had on DNS stuff. But that issue is long gone!

Thanks 👍

PR #1822, opened 2021-02-07
Update configurator.conf

FKSRZgaoximg:


homeassistant:
Hi @FKSRZgaoximg, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#1822).

Thanks!

configurator/rootfs/etc/configurator.conf
```diff
@@ -12,7 +12,7 @@
     "HASS_API_PASSWORD": null,
     "USERNAME": null,
     "PASSWORD": null,
-    "ALLOWED_NETWORKS": ["172.30.32.2"],
+    "ALLOWED_NETWORKS": [],
```
frenck, line 15: This change will break Ingress support.

Could you please elaborate on the reason for this PR and the contents of this change?

frenck (changes requested):

PR #1825, opened 2021-02-08
Fix config style for new Supervisor/Hardware

pvizeli:


frenck (approved):
LGTM ✅ 

Does this have to wait until the new SU has been fully rolled out?

pvizeli:
I don't think so, people can update it manual if they miss an add-on

frenck:
Alright, in that case, I'll move ahead with all other add-ons as well 😄

PR #1835, opened 2021-02-09
Update config.json to allow the use of serial over tcp socket

rbflurry:
Underlying zwave-js and zwave-js-server supports the use of tcp://ip:port to connect serial over TCP to a zwave stick. I have tested this with the home-assistant add on and it works as it should.

Tested ok with a local custom addon. Is this the correct way about allowing the TCP path to be entered into the addon configuration?

From that zwave-js changelog.

> In addition to serial ports, serial-over-tcp connections (e.g. by using ser2net ) are now supported. You can connect to such a host using a connection string of the form tcp://hostname:port . Use these ser2net settings to host a serial port: `:raw:0::115200 8DATABITS NONE 1STOPBIT

### DIFF
--- a/zwave_js/config.json
+++ b/zwave_js/config.json
@@ -21,7 +21,7 @@
     "network_key": ""
   },
   "schema": {
-    "device": "device(subsystem=tty)",
+    "device": "str",
     "network_key": "match(|[0-9a-fA-F]{32,32})",
     "emulate_hardware": "bool?"
   },


marcelveldt:
Hmmm, this change makes a very user friendly configuration option into a dumb one again.
Is there something else to do ? Like being able to choose between the two ?

danielbrunt57:
Add `Network` to the options list with a sub-menu to enter the IP address and port number?

danielbrunt57:
image
image

frenck:
@danielbrunt57 That is nice, but out of scope and out of reach.

danielbrunt57:
Then how to fix this?
image

danielbrunt57:
An earlier version of the add-on permitted it...

frenck:
@danielbrunt57 If you want to discuss issues, that is fine in an issue or bounce ideas on our community resources. Lets keep the discussion here about the code change at hand. Thanks 👍

danielbrunt57:
I think I am talking about the code change...
_Update config.json to allow the use of serial over tcp socket_

I'll bow out...

zwave_js/config.json
```diff
@@ -21,7 +21,7 @@
     "network_key": ""
   },
   "schema": {
-    "device": "device(subsystem=tty)",
+    "device": "str",
```
frenck, line 24: This change makes sense in a way, I get the reasoning. However, we do want to think to be more fool proof. Considering this platform; using the `device()` validator makes sense here.

This also is closely related to the Z-Wave JS Core integration, that sets this up. So that is something to keep in mind here, as all changes to this schema, effects the core integration as well.

As an alternative, a separate option could be made, however, that has the same challenges ( Core configures), and additionally adds that both options become "optional" in order to be able to select one or the other.

At this point, I think this is a choice between evils. And I think we need to get a decision from the core people involved with Z-Wave JS, the Supervisor & add-ons in order to get this going. Maybe this should be an architectural discussion instead?

rbflurry:
I agree with what your are saying, I also just realized the schema was "str" up until a few days/version ago, so for some it was a breaking change on a "experimental" addon.

frenck:
I've looked around, read up a bit and my opinion is that this should not be merged. 2 Reasons.

1. Add-ons are here for covering the most common use cases not all. They are aimed as easy of use, and providing a foolproof experience.
2. Running a serial over a socket is really not recommended in this environment. Experimental use at best, which is probably not the user experience we want to promote.
3. With the new API the integration can stabilize the choice/input based on the Supervisor API.
4. If one wants to run a remote z-wave stick, running the Z-Wave JS server remotely is a way better choice and experience, which we should promote.

With that in mind, I think this is to solve an advanced use case, in which we should recommend setting up your environment manually.

frenck:
I've checked up with a couple of people involved in the development of different parts of Z-Wave JS or related parts in this matter. We agree on this and are going to friendly decline this change for the reasoning set.

rbflurry:
Thank you for the discussion, In my specific use case I cannot run the Zwave JS server remotely as my USB server is just a dumb network device without enough power to run node., Also I will point out while not an addon but an integration the Home Assistant  zigbee integration allows serial over socket. 

Again, I understand your reasoning and while I would like to see edge cases or experimental setups not get pushed out of addons, I get it and will continue to try and use custom add-ons.

GaryOkie:
> "running the Z-Wave JS server remotely is a way better choice and experience, which we should promote."
This is exactly what I set out to do on my RaZberry which needs to run independently of HASS.   But when I dug into what I needed to do, I saw this:  

> Run the Z-Wave server yourself
> This is considered a very advanced use case. In this case you run the Z-Wave JS Server or zwavejs2mqtt NodeJS application directly. Installation and maintaining this is out of scope for this document.

and this:

> Z-Wave JS Server
> These instructions are for development only. These CLIs will be available as zwave-server and zwave-client after installing the NPM package.

That scared me off from running the server remotely, and found it was much easier to run ser2net and connect via tcp.  I just had to uninstall Z-Wave-JS and go with the MQTT variant to accomplish it.     

My thought was to keep the existing device pick list as-is along with device=tty validation, but simply add a new option to enter a network device with or without validation and problem solved.    If in fact installing the server independently was actually promoted through documentation that showed how that wasn't "for developers only", well, that would have been nice.

frenck:
Nobody says you need to install Node/NPM. There are Docker containers available as well.

GaryOkie:
In addition to not knowing what Node/NPM is, I have never installed or set up Docker independently either.      If I ever want to get back to using Z-Wave-JS and the capability to use a tcp device remains blocked, I'll either take the first step into the Docker learning curve or pitch the RaZberry independent Z-Wave hub approach and get something else. 

Thanks for pointing out an option I had glossed over earlier.   Was just looking for simple.

jcam:
As with most other cases where people are looking for advanced use cases, using the zwavejs2mqtt community add-on is always an option and gives much more flexibility, including setting the serial path to anything you like :)

danielbrunt57:
I don't see how this is really an advanced use case.  If my HA Blue let me plug my RaZberry daughter card into it I would...
So, I'm going to have to either...
a) continue using zwavejs2mqtt and hope and pray it does not fall by the wayside like ozw has
or 
b) resurrect HomeSeer and move my z-wave devices back to it where they've supported an ethernet connection for years with dozens if not hundreds of users using it successfully.
I prefer the RaZberry for z-wave over a dongle as I've modified it with an external antenna which improves range & reliability 10 fold.  I was looking for the ability to use just HA to manage as much of this small network as possible.

frenck:
This discussion has been made with multiple people involved.
Please note, this only removes one of the many options available for your use case.

jcam:
First, zwavejs2mqtt is zwave js. It isn't going anywhere as long as zwave js is around.
Second, if you're using a razberry for z-wave, then you're hooking it up to a raspberry pi-something. The pi can run the zwave-js server. Why wouldn't you just do that?
Third, if it is that important to you, just make your own addon that doesn't have this validation.

rbflurry:
> just make your own addon that doesn't have this validation.

https://github.com/rbflurry/zwave_js-serial-over-tcp 

but I intend to use zwavejs2mqtt without the mqtt. Thank you.

danielbrunt57:
> ... uhm, yeah, that is an advanced use case.

I beg to differ. How can it be an advanced **_use_** case?  
Your interpretation of the meaning of that term and my interpretation seem to be quite different...
It's just a tech savvy user replacing the inferior PCBA with a real antenna that works better.

frenck:
@rbflurry Maybe the name throws you off... but zwavejs2mqtt... the MQTT part is fully optional and actually disabled by default.

Nevertheless, this PR has been closed. So, let's move the rest of this discussion into our community forum 👍

PR #1836, opened 2021-02-10
Update zwave-js to 6.2.0 release

superm1:
This release fixes lock reporting status.
See https://github.com/zwave-js/node-zwave-js/pull/1641 for details.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.6
+
+- Update zwave-js to version 6.2.0
+
 ## 0.1.5
 
 - Update hardware configuration for Supervisor 2021.02.5
--- a/zwave_js/build.json
+++ b/zwave_js/build.json
@@ -8,6 +8,6 @@
   },
   "args": {
     "ZWAVEJS_SERVER_VERSION": "1.0.0-beta.5",
-    "ZWAVEJS_VERSION": "6.1.3"
+    "ZWAVEJS_VERSION": "6.2.0"
   }
 }


MartinHjelmare (approved):
Thanks!

frenck:
The add-on version wasn't bumped...
Thus this will not release.

MartinHjelmare:
Sorry, missed that.

frenck:
Opened up #1837

superm1:
> Sorry, missed that.

Whoops, sorry for missing that.  Thanks!

bekriebel:
Was the intention to also update the z-wave JS server to 6.2.0? If so, `ZWAVEJS_SERVER_VERSION` also needs to go to `1.0.0-beta.6`.

superm1:
> Was the intention to also update the z-wave JS server to 6.2.0? If so, `ZWAVEJS_SERVER_VERSION` also needs to go to `1.0.0-beta.6`.

Huh, so does bumping zwave-js version not really do anything, you need to update zwavejs-server to get it's bump (IE https://github.com/zwave-js/zwave-js-server/pull/104)?

MartinHjelmare:
No, the Z-Wave JS library bump should be usable without the server bump.

bekriebel:
> > Was the intention to also update the z-wave JS server to 6.2.0? If so, `ZWAVEJS_SERVER_VERSION` also needs to go to `1.0.0-beta.6`.
> 
> Huh, so does bumping zwave-js version not really do anything, you need to update zwavejs-server to get it's bump (IE [zwave-js/zwave-js-server#104](https://github.com/zwave-js/zwave-js-server/pull/104))?

It bumps the library that HA is using, but not the server that is running and communicating with the devices. Personally, I'm looking forward to the 6.2.0 bump on the server for the addition of barrier controls.

superm1:
That would explain why locks aren't working properly still for me too. I guess we should have another PR to bump the server.

bekriebel:
> That would explain why locks aren't working properly still for me too. I guess we should have another PR to bump the server.

Agreed. I'm trying to test one locally right now and will push it if it works.

MartinHjelmare:
The server doesn't communicate with the devices directly. It uses the driver of the Z-Wave JS library to do that.

bekriebel:
Doesn't the server also still need to be updated, though? If I connect to the instance via websockets with the zwave-js-server client library, it reports that the driverVersion is still 6.1.3. This is also what is shown in the integration configuration page:

image

My understanding is that the ZWAVEJS_VERSION is the client library version and the ZWAVEJS_SERVER_VERSION is the server. Won't both need to be updated to get the features and fixes of the newer zwave-js versions?

MartinHjelmare:
Seems like the server is still using its dependency lock files even though we moved zwave-js to peer/dev deps. I've created a PR to bump the server in the add-on.

PR #1867, opened 2021-02-17
Restart nginx for ingress on error

pvizeli:


frenck (approved):
This is OK, but it might put NGinx in a restart loop and determine why that is, becomes really hard for the end-user if that happens.

frenck (approved):

Issue #1873, opened 2021-02-18
Auth credentials inside vlc addon nginx config


olegshilov:
I think this is not a place to store credentials
https://github.com/home-assistant/addons/blob/9fd88b5e3535e5ca84ab60e734e61876e2550b6e/vlc/rootfs/usr/share/tempio/nginx.conf#L52

joelmoses:
Looks like the implementer is trying to lock VLC to the NGINX HA ingress mode for the VLC WebUI. This is clumsy, especially considering they're also generating a telnet secret elsewhere and using it for configuring "cvlc". All it would take is a base64 encoding of the secret prefixed by ":" and passing that value to be written to the resulting file.

I agree that it's not good practice to hardcode this.

joelmoses:
My stab at a fix is here: https://github.com/home-assistant/addons/pull/1878

tofuSCHNITZEL:
I'm wondering if this even matters... Since the port of nginx used for ingress is not exposed outside the docker environment, only the HA container can access it anyway... So is there even the need of a secret? Or a generated one for that matter?

tofuSCHNITZEL:
I'm wondering... Does this even matter? The vlc port is not exposed outside the container so the only way to access it is via the ingress port of the nginx which is "protected" behind the Home Assistant login...

PR #1878, opened 2021-02-23
Remove static ingress credential from nginx config for VLC add-on

joelmoses:
Use the same generated secret as the add-on uses to seed the VLC login and tie it to the NGINX ingress. Removes the static basic credential in the nginx.conf file of the original code.

EDIT: On owner request, uses separate secret for Ingress credential.

Reference to issue here: https://github.com/home-assistant/addons/issues/1873


homeassistant:
Hi @joelmoses, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#1878).

Thanks!

joelmoses:
Putting this to the attention of @pvizeli and @MartinHjelmare who merged the initial addon.

pvizeli:
This need core discovery support to work, we bump the version until a core come out that this can use

pvizeli:
Sorry read it wrong :)

vlc/rootfs/etc/cont-init.d/ingress.sh
```diff
@@ -1,11 +1,23 @@
...
+# Create VLC secret
+# ==============================================================================
+
+# Generate password
+if bashio::fs.file_exists /data/secret; then
+    bashio::exit.ok
+else
+    pwgen 64 1 > /data/secret
+fi
+
```
pvizeli, line 12: Create own secret for the http and restore the `secret.sh` file

vlc/rootfs/etc/services.d/vlc/run
```diff
@@ -1,11 +1,11 @@
...
+# Start C-VLC daemon and enable discovery
 # ==============================================================================
 readonly PASSWORD="$(cat /data/secret)"
 
 # Send out discovery information to Home Assistant
 ./discovery &
 
 # Run daemon
-exec cvlc -I http --extraintf telnet --http-host 127.0.0.1 --http-password "INGRESS" --telnet-password "${PASSWORD}" --aout=pulse
+exec cvlc -I http --extraintf telnet --http-host 127.0.0.1 --http-password "${PASSWORD}" --telnet-password "${PASSWORD}" --aout=pulse
```
pvizeli, line 11: load his secret for http

pvizeli (changes requested):

vlc/rootfs/etc/cont-init.d/ingress.sh
joelmoses: Certainly can, but is there any harm in using the same secret given to discovery (Telnet) for ingress (HTTP) as well? Simplifies things somewhat, and neither are values that the user will ever need to know.

pvizeli: the telnet secret is shared, it's not a lot better than the static secret, just a bit more feignedly secure. We can make an `/data/secret.ingress` and `/data/secret.integration`.

pvizeli: You don't need cleanup the old secret file, that is not public jet

joelmoses: I'm not sure I understand fully why it's important to use two pwgen runs for a non-user-facing secret value, but I suppose it doesn't really matter in the end.

I'll make them separate at your request.

joelmoses:
@pvizeli - All requested changes made.

vlc/rootfs/etc/services.d/vlc/run
```diff
@@ -1,11 +1,12 @@
...
 # ==============================================================================
-# Start Z-Wave JS service for Z-Wave radio
+# Start C-VLC daemon and enable discovery
 # ==============================================================================
-readonly PASSWORD="$(cat /data/secret)"
+readonly TELNET_PASSWORD="$(cat /data/telnet-secret)"
+readonly INGRESS_PASSWORD="$(cat /data/ingress-secret)"
 
 # Send out discovery information to Home Assistant
 ./discovery &
```
pvizeli, line 9: Need also update the discovery data

vlc/rootfs/etc/cont-init.d/ingress.sh
```diff
@@ -1,11 +1,23 @@
 #!/usr/bin/with-contenv bashio
+# ==============================================================================
+# Create VLC secret
+# ==============================================================================
+
+# Generate password
+if bashio::fs.file_exists /data/ingress-secret; then
+    bashio::exit.ok
```
pvizeli, line 8: This means it will never generate an ingress config on the second run. This add-on can only run once

joelmoses: I patterned it after the way you were handling the telnet-secret; is there any harm in generating an ingress secret one time and maintaining the same secret for future runs? That's how it was before when the password was a static value of ":INGRESS".

I don't mind making it generate and use a new ingress secret on every container restart, but it seems counterproductive to me.

vlc/rootfs/etc/services.d/vlc/run
joelmoses: Yep, I missed that. Thanks.

joelmoses: Resolved with https://github.com/home-assistant/addons/pull/1878/commits/94f57fc15c33cf41a0a5c212906208797cde01f0

vlc/rootfs/etc/cont-init.d/ingress.sh
pvizeli: What do you think will be happening with the code after this?

pvizeli:
Looks stale, you can anytime reopen it after the comments are address and the PR is rebased

PR #1879, opened 2021-02-24
Bump Z-Wave JS add-on to 0.1.8

marcelveldt:
Server version 1.0.0 is only compatible with HA 2021.3.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.1.8
+
+- Bump Z-Wave JS Server to 1.0.0
+- Pin Z-Wave JS to version 6.5.0
+
 ## 0.1.7
 
 - Bump Z-Wave JS Server to 1.0.0-beta.6
--- a/zwave_js/Dockerfile
+++ b/zwave_js/Dockerfile
@@ -18,7 +18,7 @@ RUN \
         python3 \
     \
     && npm config set unsafe-perm \
-    && npm install -g @zwave-js/server@${ZWAVEJS_SERVER_VERSION} zwave-js@${ZWAVEJS_VERSION} \
+    && npm install -g @zwave-js/server@${ZWAVEJS_SERVER_VERSION} zwave-js@${ZWAVEJS_VERSION} zwave-js/core@${ZWAVEJS_VERSION} \
     \
     && apk del --no-cache \
         .build-dependencies
--- a/zwave_js/Dockerfile
+++ b/zwave_js/Dockerfile
@@ -18,7 +18,10 @@ RUN \
         python3 \
     \
     && npm config set unsafe-perm \
-    && npm install -g @zwave-js/server@${ZWAVEJS_SERVER_VERSION} zwave-js@${ZWAVEJS_VERSION} zwave-js/core@${ZWAVEJS_VERSION} \
+    && npm install -g \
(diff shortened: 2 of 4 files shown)


MartinHjelmare (commented):
Please update the changelog too.

zwave_js/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN \
         python3 \
     \
     && npm config set unsafe-perm \
-    && npm install -g @zwave-js/server@${ZWAVEJS_SERVER_VERSION} zwave-js@${ZWAVEJS_VERSION} \
+    && npm install -g @zwave-js/server@${ZWAVEJS_SERVER_VERSION} zwave-js@${ZWAVEJS_VERSION} zwave-js/core@${ZWAVEJS_VERSION} \
```
MartinHjelmare, line 21: There was an `@` missing.
```suggestion
    && npm install -g \
        @zwave-js/server@${ZWAVEJS_SERVER_VERSION} \
        zwave-js@${ZWAVEJS_VERSION} \
        @zwave-js/core@${ZWAVEJS_VERSION} \
```

MartinHjelmare (changes requested):

marcelveldt:
Adding the zwave-js/core package is a workaround, without it, things will crash because that core package could not be found. Seems to be related to installing packages globally.

Info here:
https://discord.com/channels/330944238910963714/800356888827002880/814166911151243284

And here:
https://stackoverflow.com/questions/12594541/npm-global-install-cannot-find-module/14515868#14515868

MartinHjelmare (approved):
This should be merged when we cut the core beta release.

zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.1.8
+
+- Bump Z-Wave JS Server to 1.0.0
```
MartinHjelmare, line 5: Should add a header `Breaking change:` above this?

balloob: It's not a breaking change because older HA versions can't install it.

MartinHjelmare: Right, just wondered if we should point this out more in the notes.

raidenlal:
don't know if this would be the place to ask but, after this merge I am not able to update the addon in home assistant:

"You are running Home Assistant 2021.2.3, to install this add-on you need at least version 2021.2.0b0 of Home Assistant"

I have tried beta channel a week ago, then I disabled it and I was left at 2021.2.3 and now ZWaveJS addon is not happy.

MartinHjelmare:
There are breaking changes in the server 1.0.0 api which requires Home Assistant core 2021.3.0b0. We're working on a solution to allow older clients to connect to the new server but it wasn't ready for beta0. When the solution is ready we should be able to allow older versions of Home Assistant core to install the add-on again.

MartinHjelmare:
The add-on message about minimum version is incorrect, and confusing. It should say minimum versions 2021.3.0b0. That's a separate issue and we're working on that too.

raidenlal:
Thanks for the reply. What I am puzzled is that my installed version is 2.3 which SEEMS to be higher than required 2.0b0. Could you maybe shortly explain how the versioning is actually treated?

MartinHjelmare:
https://github.com/home-assistant/addons/issues/1880

raidenlal:
Sorry I didn't notice your "it should say 3.0b". Thank you for explanation.

PR #1905, opened 2021-03-06
Bump zwave-js-server to 1.1.1

raman325:
Changelog: https://github.com/zwave-js/zwave-js-server/releases/tag/1.1.1

### DIFF
--- a/zwave_js/build.json
+++ b/zwave_js/build.json
@@ -7,7 +7,7 @@
     "aarch64": "homeassistant/aarch64-base:3.13"
   },
   "args": {
-    "ZWAVEJS_SERVER_VERSION": "1.1.0",
+    "ZWAVEJS_SERVER_VERSION": "1.1.1",
     "ZWAVEJS_VERSION": "6.5.0"
   }
 }
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.11
+
+- Bump Z-Wave JS Server to 1.1.1
+
 ## 0.1.10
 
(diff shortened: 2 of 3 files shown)


MartinHjelmare (commented):
Please update the add-on changelog and version.

zwave_js/build.json
```diff
@@ -7,7 +7,7 @@
     "aarch64": "homeassistant/aarch64-base:3.13"
   },
   "args": {
-    "ZWAVEJS_SERVER_VERSION": "1.1.0",
+    "ZWAVEJS_SERVER_VERSION": "1.1.1",
     "ZWAVEJS_VERSION": "6.5.0"
```
MartinHjelmare, line 11: Should we bump zwave-js too? It that safe?

https://github.com/zwave-js/node-zwave-js/releases

raman325: The one change that's "breaking" is the switch to "buffer" for the value type if the result is returned as a buffer object. We would need to add support for that in the library, otherwise those values won't be stored.

raman325: If we are ok with that limitation then I can bump zwave-js as well

MartinHjelmare: Do we support the values today that would become buffer type?

raman325: We do: https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/zwave_js_server/model/value.py#L236

The change is to switch the type here from `string` to `buffer` for these scenarios

MartinHjelmare: Ok, then we should add support first in the library and Home Assistant before we bump.

MartinHjelmare (approved):

Issue #1690, opened 2020-11-22
Git Pull addon erased my entire config without warning


aaroncoffey:
## The problem

This is the exact same issue as https://github.com/home-assistant/addons/issues/1325, installing the git pull addon with an empty repo erased my entire config. This needs to be very clearly spelled out in the documentation, assuming this is working as intended.
Other users have experienced this in this thread: https://community.home-assistant.io/t/git-pull-addon/23435/43

## Environment

Home Assistant 0.118.2
HassOS 4.17
Running in a vm

- Add-on with the issue: git-pull
- Add-on release with the issue: 7.12
- Last working add-on release (if known): NA
- Operating environment (OS/Supervised): HassOS 4.17

## Problem-relevant configuration

```yaml
deployment_key:
  - '-----BEGIN OPENSSH PRIVATE KEY-----'
redacted
  - '-----END OPENSSH PRIVATE KEY-----'
deployment_key_protocol: rsa
deployment_user: ''
deployment_password: ''
git_branch: master
git_command: pull
git_remote: origin
git_prune: false
repository: 'git@github.com:me/myrepo.git'
auto_restart: false
restart_ignore:
  - ui-lovelace.yaml
  - .gitignore
repeat:
  active: false
  interval: 300
```

## Traceback/Error logs

```txt
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
Check SSH connection
[Warn] No valid SSH connection for git@github.com
[Info] Start adding SSH key
[Info] Setup deployment_key on id_rsa
[Warn] Git repostory doesn't exist
[Info] Backup configuration to /tmp/config-2020-11-22_09-27-53
[Info] Start git clone
Cloning into '/config'...
warning: You appear to have cloned an empty repository.
[Info] Checking if something has changed...
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
[Info] Something has changed, checking Home-Assistant config...
'git <command> [<revision>...] -- [<file>...]'
Error: Testing configuration at /config
Failed config
  General Errors: 
    - File configuration.yaml not found.
Successful config (partial)
[...]

PR #1911, opened 2021-03-09
Made git_pull more robust to wrong repo setup and added failsafe backup

ladidadida:
This patch is to solve issue #1690 and similar.
The problem is, that especially in the beginning it is very likely to run into problems, but without checking quality of remote Repo, all data is deleted.
This patch tries to solve this, but not only create backups far more often, but because it reverts configuration in case of bigger problems.

A second, small change is, to not delete the whole config folder, but to just delete yaml files from it. If additional files should be deleted, this could be achieved with git, too. (add a file to repo once and then delete it from repo). I would assume that only pulling updates from Repo is sufficient, but this is not yet implemented.

### DIFF
--- a/git_pull/CHANGELOG.md
+++ b/git_pull/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 7.12.2
+
+- Added automatic backup recovery
+- Corrected too harsh deletion of config files
+
 ## 7.12.1
 
 - Update options schema for passwords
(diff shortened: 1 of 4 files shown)


pvizeli (commented):
Please use bashio for logging and exit

ladidadida:
@pvizeli 
> Please use bashio for logging and exit
Should I do that only for my changed parts or for the whole file?
I have tried to use bashio for getting the configuration items, but it seems this does not work that easy. I would have to dig more into the bashio functionality.

Is there any documentation for bashio?

pvizeli:
https://github.com/hassio-addons/bashio

ladidadida:
> https://github.com/hassio-addons/bashio

Short answer with zero new information ;) The package lacks documentation. Ok, some might now say good code is documentation enough, but I am not one of those. I digged into it and found that it should not differ. I will have to dig deeper into my changes.

PR #1935, opened 2021-03-23
autodetect workspace name

sandercox:
This fixes issues when you do not checkout this repo into the folder `addons` but for example `ha-addons`.

But it also helps with the instructions to copy & paste the `.devcontainer` and `.vscode` folders from the addons project into your own addons folder. That might not have been named `addons`

### DIFF
--- a/.devcontainer/supervisor.sh
+++ b/.devcontainer/supervisor.sh
@@ -74,6 +74,16 @@ function cleanup_docker() {
 
 function run_supervisor() {
     mkdir -p /tmp/supervisor_data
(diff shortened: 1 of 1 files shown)


rchl:
I've just found your PR after creating mine (#1966) and I think mine is the right way to do but feel free to convince me otherwise. :)

ludeeus:
I think <https://github.com/home-assistant/addons/pull/2000> is a better way to handle this.

ludeeus:
With #2000 now merged, this is no longer needed, thanks anyway :+1:

PR #1939, opened 2021-03-24
Bump zwave-js-server to 1.3.0

raman325:
Changelog: https://github.com/zwave-js/zwave-js-server/releases/tag/1.3.0

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.1.14
+
+- Bump Z-Wave JS Server to 1.3.0
+- Pin Z-Wave JS to version 7.0.0
+
 ## 0.1.13
 
 - Bump Z-Wave JS Server to 1.2.1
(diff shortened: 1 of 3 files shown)


frenck (approved):
LGTM ✅ 

Not sure if we can merge e.g., compatibility and such?

raman325:
It's intended to be backwards compatible so it should be ok to release

balloob (approved):

balloob:
The goal of ZJS Server is to be the place where we solve compatibility issues and so support older versions of HA and newer ones (cur version HA uses schema 2, this server also supports schema 3). So this is ok to go. Leaving it up to @MartinHjelmare for final sign off because I don't know if it is tested, just know the theory.

raman325:
FWIW I am running the master branch of `zwave-js-server` in my production HA instance (2021.3), everything appears to be working as expected

MartinHjelmare:
I'll test a bit tomorrow too and then merge if everything is alright.

MartinHjelmare (approved):
Local build and test works well.

PR #1941, opened 2021-03-24
Added privilege control per user/db to MariaDB

mdegat01:
I recently discovered you can access HA's recorded data in MariaDB using the [grafana community add-on](https://github.com/hassio-addons/addon-grafana). When you do that though it says to ensure the user grafana is using only has `SELECT` access because it will not prevent you from sending destructive commands that change data or schema. I realized the add-on has no way to do that, any users listed in the `rights` section are granted `ALL PRIVILEGES` on the dbs listed.

So I added a new option for `privileges`. When provided, the user is only granted the listed privileges on the DB. If omitted then the user is granted `ALL PRIVILEGES` like normal.

Note that the option expects a string with a comma-separated list of privileges inside the string. I would've preferred to ask for an array of privileges but there doesn't seem to be a way to make a config which accepts an array optional and I didn't want to introduce a breaking change.


mdegat01:
Note that I did test this change locally. In a virtualbox instance of HA OS I pulled my MariaDB changes into the `/addons` folder and tried it out. Here was my config:

```yaml
databases:
  - testdb
logins:
  - username: admin
    password: admin
  - username: test1
    password: test1
rights:
  - username: admin
    database: testdb
  - username: test1
    database: testdb
    privileges: 'SELECT, UPDATE'
```

And here was my simple test:

```
➜  mariadb mysql -u 'test1' -p -h 'local-mariadb'
Enter password:
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 25
Server version: 10.4.17-MariaDB MariaDB Server

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> use testdb
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
MariaDB [testdb]> SELECT * FROM tutorials_tbl
    -> ;
+-------------+----------------+-----------------+-----------------+
| tutorial_id | tutorial_title | tutorial_author | submission_date |
+-------------+----------------+-----------------+-----------------+
|           1 | test           | test            | NULL            |
+-------------+----------------+-----------------+-----------------+
1 row in set (0.002 sec)

MariaDB [testdb]> create table tutorials_tbl1(PRIMARY KEY ( tutorial_id ));
ERROR 1142 (42000): CREATE command denied to user 'test1'@'172.30.32.1' for table 'tutorials_tbl1'
```

mariadb/config.json
```diff
@@ -25,7 +25,7 @@
   "schema": {
     "databases": ["str"],
     "logins": [{ "username": "str", "password": "password" }],
-    "rights": [{ "username": "str", "database": "str" }]
+    "rights": [{ "username": "str", "database": "str", "privileges": "str?" }]
```
pvizeli, line 28: Use ["list(X|Y|Z)"], we should not allow making mistakes

mariadb/DOCS.md
```diff
@@ -61,6 +61,10 @@ This should be the same user name defined in `logins` -> `username`.
 
 This should be the same database defined in `databases`.
 
+### Option: `rights.privileges` (optional)
+
+A list of privileges to grant to this user from [grant][grant] like `SELECT, CREATE`. If omitted, grants `ALL PRIVILEGES` to user. Restricting privileges of the user that homeassistant uses is not recommended but if you want to allow other applications to view recorder data should create a user limited to read-only access on the database.
```
frenck, line 66:
```suggestion
A list of privileges to grant to this user from [grant][grant] like `SELECT, CREATE`. If omitted, grants `ALL PRIVILEGES` to the user. Restricting privileges of the user that Home Assistant uses is not recommended but if you want to allow other applications to view recorder data should create a user limited to read-only access on the database.
```

mariadb/rootfs/etc/services.d/mariadb/run
```diff
@@ -84,8 +84,15 @@ for right in $(bashio::config "rights|keys"); do
     USERNAME=$(bashio::config "rights[${right}].username")
     DATABASE=$(bashio::config "rights[${right}].database")
 
-    bashio::log.info "Alter rights for ${USERNAME} to ${DATABASE}"
-    mysql -e "GRANT ALL PRIVILEGES ON ${DATABASE}.* TO '${USERNAME}'@'%';" 2> /dev/null || true
+    if bashio::config.exists "rights[${right}].privileges"; then
```
frenck, line 87: There is some tabs vs spaces making this hard to read, please use spaces.

mariadb/config.json
mdegat01: So I can't do this currently. The config is expecting the user to provide a string with one or more privileges separated by commas like this `SELECT, CREATE, UPDATE`. I could only do a `list(...)` config if I changed it to this:

```json
"rights": [{ "username": "str", "database": "str", "privileges": ["list(X|Y|Z)?"] }]
```

This is the configuration I wanted to do but it would be a breaking change. All existing configs would need to be adjusted like this to pass validation now and continue to work as expected
```diff
rights:
  - username: admin
    database: testdb
+  privileges: []
```

I suppose I can do a long `match(...)` configuration though so I'll get to work on that for now.

mdegat01: If there is a way to add an optional array config that can be omitted let me know. I would prefer users configured this like this, think it would be nicer:

```yaml
rights:
  - username: admin
    database: testdb
    privileges:
      - SELECT
      - CREATE
      - UPDATE
```

I just didn't want to introduce a breaking change

mdegat01: Added a match statement. Tested it by pulling it into `/addons` as a local addon and trying out various good and bad configs. Seems to correctly enforce the syntax we're looking for now. Let me know if that works.

mariadb/config.json
```diff
@@ -25,7 +25,13 @@
   "schema": {
     "databases": ["str"],
     "logins": [{ "username": "str", "password": "password" }],
-    "rights": [{ "username": "str", "database": "str" }]
+    "rights": [
+      {
+        "username": "str",
+        "database": "str",
+        "privileges": "match(^((ALTER|CREATE|CREATE ROUTINE|CREATE TEMPORARY TABLES|CREATE VIEW|DELETE|DELETE HISTORY|DROP|EVENT|GRANT OPTION|INDEX|INSERT|LOCK TABLES|SELECT|SHOW VIEW|TRIGGER|UPDATE)(, (?!$)|$))+$)?"
```
pvizeli, line 32:
```suggestion
        "privileges": ["match(^(ALTER|CREATE|CREATE ROUTINE|CREATE TEMPORARY TABLES|CREATE VIEW|DELETE|DELETE HISTORY|DROP|EVENT|GRANT OPTION|INDEX|INSERT|LOCK TABLES|SELECT|SHOW VIEW|TRIGGER|UPDATE)$)?"]
```

mariadb/config.json
pvizeli: `"privileges": ["list(x|y|z)?"]` should work. Otherwise, it's a bug which we should fix on the supervisor and release the add-on after the bugfix.

mdegat01: Ok makes sense. So yea looks like that configuration isn't working properly. I made a simple test addon with this for `schema` and `options`:

```json
{
	"options": {
		"log_level": "info"
	},
  "schema": {
		"log_level": "str",
		"privileges": ["list(x|y|z)?"]
	}
}
```

If I just `log_level` whatever I type won't be saved and will just revert because its invalid. But as soon as I add `privileges: []` to my config then its all good and it saves fine. Basically it treats the elements inside the array as option but the field itself becomes mandatory within that object. I can set it to `[]` but I can't leave it out.

mdegat01: Added a [PR](https://github.com/home-assistant/supervisor/pull/2752) to supervisor to add that capability. Will now update this to use it

mdegat01: Ok just updated it to use a list properly now. Also tweaked the DOCS in light of the change and modified the example to show a read-only user for completeness.

pvizeli (approved):
Can be merge after SU is released 👍 thanks

PR #1944, opened 2021-03-27
Correct grammatical error in MariaDB add-on description

chriscn:
Should be A SQL database server as S isn't a vowel.

### DIFF
--- a/mariadb/config.json
+++ b/mariadb/config.json
@@ -2,7 +2,7 @@
   "name": "MariaDB",
   "version": "2.2.2",
   "slug": "mariadb",
-  "description": "An SQL database server",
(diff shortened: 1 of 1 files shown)


frenck (approved):
Thanks, @chriscn 👍

MartinHjelmare:
It depends on how you pronounce SQL. If you spell out each letter separately, the `S` should be referenced with `An` since `s` has a leading vowel (ess) sound. If you pronounce `SQL` as `sequel` it should be `A (sequel)`.

frenck:
<https://oracle-base.com/blog/2015/01/02/a-sql-or-an-sql/>

Even Google & Microsoft style guides mess it up in both ways.

PR #1947, opened 2021-03-28
Bump zwave-js to 7.0.1

raman325:
Changelog: https://github.com/zwave-js/node-zwave-js/releases/tag/v7.0.1

A couple of notes:
1. There appears to be a change introduced in zwave-js 7.0.0 that causes different values to be mapped to different endpoints in such a way that our existing climate discovery logic doesn't capture all of the necessary values (https://github.com/zwave-js/node-zwave-js/issues/2165). It looks like this is resolved in 7.0.1 but I am waiting for confirmation.
2. Even in 7.0.1 the `Thermostat Setpoint` and `Thermostat Mode` values for my thermostat used to be on endpoint 0 and now are on endpoint 1. The end result is that the originally created climate entry is effectively abandoned, and a new one has been created. We probably need to add some logic to clean up entities that previously existed that are no longer being discovered to handle this change. I am not sure the best way to address it yet though so open to suggestions.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.15
+
+- Pin Z-Wave JS to version 7.0.1
+
 ## 0.1.14
 
 - Bump Z-Wave JS Server to 1.3.0
--- a/zwave_js/build.json
+++ b/zwave_js/build.json
@@ -8,6 +8,6 @@
   },
   "args": {
     "ZWAVEJS_SERVER_VERSION": "1.3.0",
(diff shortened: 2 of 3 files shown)


MartinHjelmare:
I've tested this PR locally and it works alright for me.

Let me know when we should merge this.

For 2. if possible we should migrate the entity registry data for the old entity to the new entity and remove the old entity entry from the entity registry.

raman325:
> I've tested this PR locally and it works alright for me.
> 
> Let me know when we should merge this.
> 
> For 2. if possible we should migrate the entity registry data for the old entity to the new entity and remove the old entity entry from the entity registry.

I think we should merge, it's not going to make the climate issue worse, just possibly better.

PR #1950, opened 2021-03-30
Expose log level as a configuration option for zwave_js

raman325:
Right now, log level is decided based on the supervisor log level. That buries the option for most users and doesn't allow for the other levels that `zwave-js` supports, so this makes it user configurable directly on the addon.

### DIFF
--- a/zwave_js/DOCS.md
+++ b/zwave_js/DOCS.md
@@ -86,7 +86,7 @@ and your controller, before rebuilding your Z-Wave network.
 ### Option `log_level` (optional)
 
 This option sets the log level of Z-Wave JS. If no `log_level` is specified, the
-log level will be set to default. Options are:
+log level will be set to `info`. Valid options are:
 - silly
 - debug
 - verbose
--- a/zwave_js/DOCS.md
+++ b/zwave_js/DOCS.md
@@ -86,7 +86,7 @@ and your controller, before rebuilding your Z-Wave network.
 ### Option `log_level` (optional)
 
 This option sets the log level of Z-Wave JS. If no `log_level` is specified, the
-log level will be set to `info`. Valid options are:
+log level will be set to the level set in the Supervisor. Valid options are:
 - silly
 - debug
 - verbose
--- a/zwave_js/DOCS.md
+++ b/zwave_js/DOCS.md
@@ -83,6 +83,18 @@ don't have a backup of this key, you won't be able to reconnect to any securely
 included devices. This may mean you have to do a factory reset on those devices
 and your controller, before rebuilding your Z-Wave network.
 
+### Option `log_level` (optional)
+
+This option sets the log level of Z-Wave JS. If no `log_level` is specified, the
+log level will be set to default. Options are:
+- silly
+- debug
+- verbose
+- http
+- info
+- warn
+- error
+
 ### Option `emulate_hardware` (optional)
 
 If you don't have a USB stick, you can use a fake stick for testing purposes.
--- a/zwave_js/DOCS.md
+++ b/zwave_js/DOCS.md
@@ -85,8 +85,7 @@ and your controller, before rebuilding your Z-Wave network.
 
 ### Option `log_level` (optional)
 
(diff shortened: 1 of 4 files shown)


zwave_js/config.json
```diff
@@ -18,11 +18,13 @@
...
   "options": {
     "device": null,
-    "network_key": ""
+    "network_key": "",
+    "log_level": "info"
   },
   "schema": {
     "device": "device(subsystem=tty)",
     "network_key": "match(|[0-9a-fA-F]{32,32})",
+    "log_level": "list(silly|debug|verbose|http|info|warn|error)",
```
frenck, line 27: Make it optional? (note: does need handling to programmatically set a default when missing)

```suggestion
    "log_level": "list(silly|debug|verbose|http|info|warn|error)?",
```

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -30,11 +30,12 @@ else
     network_key=$(bashio::config 'network_key')
 fi
 
+# log_level=$(bashio::config 'log_level')
```
frenck, line 33: Stale debug

frenck (commented):
As the configuration is managed by the integration, does this impact the integration?

raman325:
> As the configuration is managed by the integration, does this impact the integration?

No, it just sets the log level that gets dumped to console. It might be nice in the future to build a UI around exporting logs for troubleshooting and that might require more changes to the addon, but nothing planned as of now

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -30,11 +30,16 @@ else
     network_key=$(bashio::config 'network_key')
 fi
 
+if  ! bashio::config.has_value 'log_level'; then
+    bashio::log.info "No log level specified, falling back to info..."
+    bashio::addon.option log_level "info"
```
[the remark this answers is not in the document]
raman325: no, not intention, adjusted

zwave_js/config.json
raman325: I have tried setting the default `log_level` to an empty string but I can't save my configuration, even with the question mark. Do I need to have a blank option?

raman325: As far as I can tell, it's not possible to remove log_level entirely. I tested this in the zwavejs2mqtt hassio addon as well because I noticed it also makes this optional, and it didn't work there as well 🤷🏾‍♂️

zwave_js/DOCS.md
```diff
@@ -86,7 +86,7 @@ and your controller, before rebuilding your Z-Wave network.
 ### Option `log_level` (optional)
 
 This option sets the log level of Z-Wave JS. If no `log_level` is specified, the
-log level will be set to `info`. Valid options are:
+log level will be set to the level set in the Supervisor. Valid options are:
```
MartinHjelmare, line 89: This makes it sound like the valid options are for the Supervisor. Maybe move the sentence about the default log level after the list of valid options?

raman325:
I would consider this a high priority PR because we need to make it easier for users to collect debug logs for submitting upstream issues. Let me know if there's any other feedback that I need to consider

MartinHjelmare (approved):
Looks good, but I'd like Franck or Pascal to approve since I'm not so familiar with bashio.

raman325:
@frenck sorry for the tag but I am interested in seeing this released ASAP so we can make it easier for users to provide debug logs from thee addon. Can you give this another hopefully quick review?

frenck (approved):
I think this is fine to add 👍

frenck (changes requested):
Please rebase, update changelog and bump version.

raman325:
> Please rebase, update changelog and bump version.

I've rebased, but can we merge as is so I can bump zwave-js as part of this next release as well? I will open a separate PR for that and include both in the changelog

frenck (approved):

PR #1985, opened 2021-04-23
Update configurator / file editor to 0.4.1

danielperna84:
I have released a new version of the configurator which ignores the `input` keyword during yaml-linting. This PR references this new version.

I don't know how the versions here are _calculated_, so I just went with `5.3.0`. Tell me if it should be something else. 👍

### DIFF
--- a/configurator/CHANGELOG.md
+++ b/configurator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.3.0
+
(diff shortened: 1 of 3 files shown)


frenck (changes requested):
The version in the `config.json` needs to be bumped as well.

danielperna84:
Done. 👍

frenck (approved):
Thanks, @danielperna84 👍 

../Frenck 🚂

PR #1986, opened 2021-04-25
Update config.json

myhobby2:
Needed to go back to version 5.1 (downgrade) after HA is not working anymore

### DIFF
--- a/mosquitto/config.json
+++ b/mosquitto/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Mosquitto broker",
-  "version": "5.1.1",
+  "version": "5.1",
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @myhobby2, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#1986).

Thanks!

frenck:
There is an open issue for it, furthermore, the bug should be addressed. There is no direct reason to downgrade at this point.

If there is a reason for a user, they should use a snapshot.

PR #1988, opened 2021-04-28
Fix issues on WSL2

pvizeli:
Fix: #1932

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -19,7 +19,21 @@ RUN \
         xz-utils
 
 # Install docker
-RUN curl -fsSL https://get.docker.com | sh -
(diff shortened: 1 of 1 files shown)


frenck (approved):

pgross41:
I am using WSL2 and the dev container build is failing with the following: 

```
#9 0.349 update-alternatives: using /usr/sbin/iptables-legacy to provide /usr/sbin/iptables (iptables) in manual mode
#9 0.356 update-alternatives: error: alternative /usr/sbin/iptables-legacy for ip6tables not registered; not setting
```

If I comment out `RUN update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy` then it builds successfully but http://localhost:7123/ says `ERR_EMPTY_RESPONSE` and HA does not load.

PR #1989, opened 2021-04-29
check_config: Upgrades add-on base image to 2021.04.3

frenck:


frenck:
Shellcheck build error addressed in #1991

PR #1998, opened 2021-05-04
Add proxy_pass customizability

Deltamir:
Add a new config option allowing users to change the _proxy_pass_ parameter in nginx.conf.
The default setting is left as it was so most users won't see any difference. 

Changing this config option can increase the security level for users willing to do so.
A good hardening practice is to reduce the service exposure as much as possible.
This option enable users to remove the local exposure with unciphered protocol (http) on port 8123 on the LAN interface.

In configuration.yaml, you can set the _http: server_host_ to choose the binded interface.
Setting this parameter to 127.0.0.1 bind the home assistant web service on the loopback and therefore remove the LAN exposure.
This configuration needs the Nginx Reverse Proxy to send back the web feed to the loopback interface, which is not possible with the static FQDN configuration on _homeassistant.local.hass.io_ with static port.

This commit enable user to set the _proxy_pass setting_ to _http://127.0.0.1:8123_ and the _http: server_host_ configuration to _127.0.0.1_ in order to harden their system.

Discussion is open in case this is a bad idea for others considerations than hardening :)
Got inspiration from https://community.home-assistant.io/t/reverse-proxy-using-nginx/196954

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -33,6 +33,7 @@ domain: home.example.com
 certfile: fullchain.pem
 keyfile: privkey.pem
 hsts: "max-age=31536000; includeSubDomains"
+proxypass: http://homeassistant.local.hass.io:8123  
 customize:
   active: false
   default: "nginx_proxy_default*.conf"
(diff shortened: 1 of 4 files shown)


frenck:
> This commit enable user to set the proxy_pass setting to http://127.0.0.1:8123 and the http: server_host configuration to 127.0.0.1 in order to harden their system.

This makes no sense, as the used transport is a Docker virtual network. It doesn't use any exposed interface.

PR #2010, opened 2021-05-19
Update bug report form

ludeeus:
Update issue form to the new format to get better details when issues are opened.


frenck:
I really want to limit the use to prevent needing to upgrade forms everywhere when new breaking changes come in (which happens quite a bit).

ludeeus:
Resurrected in https://github.com/home-assistant/addons/pull/2262

Issue #2022, opened 2021-05-20
Mosquitto add-on verbose logging 


b3nj1:
Hi,
The mosquitto logging level has been increased (again). Back in PR https://github.com/home-assistant/addons/pull/1708, we used the system logging level into the mosquitto logging level, but this has been removed in https://github.com/home-assistant/addons/pull/2007.

I'd like to submit a PR to address this, but first, @frenck, was the unconfigurable logging level change intentional for some reason. If it was intentional, is there a preferred way to make it configurable over the way we had it below?

New template log level:  
https://github.com/home-assistant/addons/blob/54ce6a192283fa6b5e0e84b741930c6b9ca64053/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl#L7

Old code to set logging level based on configured level:
https://github.com/home-assistant/addons/blob/107585db0ce949b46270644a580f090331a02a21/mosquitto/data/run.sh#L92-L101

Thanks,

- Benjamin

frenck:
There is not need to do that or add such logic. If you want to override settings on that level, the customization feature is available. That would allow you to write about any option to mosquitto.

PR #2029, opened 2021-05-22
remove redundant log_type settings to allow customization

b3nj1:
Fix for https://github.com/home-assistant/addons/issues/2022 .
No change to existing behavior since the previously configured log_types are the defaults.
User can now customize log_types using the customize feature. Before, log_types could only be added to, not removed.

Thanks for considering this PR,

- Benjamin

### DIFF
--- a/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
+++ b/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
@@ -1,10 +1,6 @@
 protocol mqtt
 user root
 log_dest stdout
-log_type error
-log_type warning
-log_type notice
-log_type information
 persistence true
 persistence_location /data/
 


b3nj1:
Hi @frenck ,

Do you know where these '{"result": "ok", "data": {}}' (no newline) strings are coming from? When I reduce the mosquitto logging, these result strings remain. They seem to come each time a message is published.

```
2021-06-19T11:17:44: New connection from 10.0.1.62 on port 8883.
{"result": "ok", "data": {}}2021-06-19T11:17:44: New client connected from 10.0.1.62 as mosqsub|8855-raspberryp (p2, c1, k60, u'mqtt_rpi').
2021-06-19T11:18:34: New connection from 172.30.32.2 on port 1883.
```

Thanks,
Benjamin

PR #2038, opened 2021-05-25
Persist zwave_js device config files

raman325:
This is to make way for supporting the updating of the device config files without a version update through the UI and ensuring that it persists between restarts.


zwave_js/rootfs/etc/services.d/zwave_js/run
```diff
@@ -12,5 +12,8 @@ fi
 # Send out discovery information to Home Assistant
 ./discovery &
 
-# Run daemon
-exec zwave-server ${SERIAL_DEVICE} --config /etc/zwave_config.json
+# Create external config dir so configs can be persisted
+mkdir -p /data/db
```
MartinHjelmare, line 16: Is this the correct file for creating directories? We have a structure script that creates the data cache.

raman325: Ah, missed that. Will adjust

raman325:
We can, but we can also wait until the next server/zwave-js bump since this provides no utility until we add support for device config file updates in the UI

MartinHjelmare:
Ok, we can wait with releasing this then. Let's make sure we don't forget to include this in the changelog.

zwave_js/rootfs/etc/services.d/zwave_js/run
```diff
@@ -12,5 +12,5 @@ fi
 # Send out discovery information to Home Assistant
 ./discovery &
 
-# Run daemon
-exec zwave-server ${SERIAL_DEVICE} --config /etc/zwave_config.json
+# Run daemon, passing external config directory in as environment variable
+exec env ZWAVEJS_EXTERNAL_CONFIG=/data/db zwave-server ${SERIAL_DEVICE} --config /etc/zwave_config.json
```
frenck, line 16: A little more readable suggestion

```suggestion
# Set external config directory
export ZWAVEJS_EXTERNAL_CONFIG=/data/db

# Run daemon, passing external config directory in as environment variable
exec zwave-server ${SERIAL_DEVICE} --config /etc/zwave_config.json
```

raman325: I swear I tried this and it didn't work, but of course now it does 😅

PR #2042, opened 2021-05-26
Update Almond to 2.0.0

gcampax:
Also, change the version scheme of the add-on to be the Almond
version number + a monotonically increasing build number, so we
can convert back the add-on version to the Almond version, for
bug triaging.


almond/DOCS.md
```diff
@@ -27,14 +27,11 @@ Got questions?
...
 In case you've found a bug, please [open an issue on our GitHub][issue].
 
-[discord]: https://discord.gg/c5DvZ4e
-[forum]: https://community.home-assistant.io
-[issue]: https://github.com/home-assistant/hassio-addons/issues
-[reddit]: https://reddit.com/r/homeassistant
-[repository]: https://github.com/hassio-addons/repository
+[discord]: https://discord.gg/anthtR4
+[forum]: https://community.almond.stanford.edu
+[issue]: https://github.com/stanford-oval/almond-server/issues
```
frenck, line 37: The core add-ons are part of Home Assistant, and thus should be handled here.

Please revert this change. It is OK to link to upstream, but we want to handle and collect issues here. The project carries the responsibility for these add-ons and will redirect users upstream when needed. Additionally, it is very confusing to guide people to different places for different add-ons.

almond/Dockerfile
```diff
@@ -5,50 +5,77 @@ FROM ${BUILD_FROM}
...
     THINGENGINE_HOME="/data/almond-server" \
-    THINGENGINE_HOST_BASED_AUTHENTICATION="local-ip"
+    THINGENGINE_HAS_REVERSE_PROXY=true \
+    THINGENGINE_HOST_BASED_AUTHENTICATION=insecure
 
 WORKDIR /opt/almond
 ARG ALMOND_VERSION
 
+RUN adduser --disabled-password almond && \
+   chown almond:almond /opt/almond
```
frenck, line 15: We generally don't run under a different user. This might cause issues with things like restoring snapshots, as the permissions on those are not guaranteed. Have you taken care of that? If so, how?

almond/build.json
```diff
@@ -1,12 +1,11 @@
...
   "build_from": {
-    "aarch64": "homeassistant/aarch64-base:3.13",
-    "amd64": "homeassistant/amd64-base:3.13",
-    "i386": "homeassistant/i386-base:3.13",
-    "armv7": "homeassistant/armv7-base:3.13",
-    "armhf": "homeassistant/armhf-base:3.13"
+    "aarch64": "homeassistant/aarch64-base-debian:buster",
+    "amd64": "homeassistant/amd64-base-debian:buster",
+    "armhf": "homeassistant/armhf-base-debian:buster",
+    "armv7": "homeassistant/armv7-base-debian:buster"
```
frenck, line 6: Why move away from Alpine?

almond/config.json
```diff
@@ -1,17 +1,16 @@
 {
-  "name": "Almond",
-  "version": "1.1.2",
-  "slug": "almond",
-  "description": "The home server version of Almond",
+  "name": "Almond - Edge",
```
frenck, line 2: This is not the edge, we only want stable add-on in this repo.

almond/config.json
```diff
@@ -1,17 +1,16 @@
...
+  "version": "2.0.0-1",
```
frenck, line 3: Version numbers of add-ons are decoupled with the software it wraps.

The `-1` addition is not something we use.

```suggestion
  "version": "2.0.0",
```

almond/rootfs/etc/services.d/almond/run
```diff
@@ -7,6 +7,13 @@ export THINGENGINE_BASE_URL
 # Set the Ingress URL as Almond base URL for correct handling
 THINGENGINE_BASE_URL=$(bashio::addon.ingress_entry)
 
+mkdir -p /data/almond-server
+if [ ! -f /data/almond-server/prefs.db ]; then
+    # Skip authentication handling
+    echo '{"server-login":{"password":"x","salt":"x","sqliteKeySalt":"x"}}' \
+      > /data/almond-server/prefs.db
```
frenck, line 14: If this file contains fixed content, it can be added as a fixed file. It doesn't need to be "created" from a bash file.

Additionally, this is not runtime/service part, but an init part as it seems. Please put this into an init script.

almond/README.md
```diff
@@ -1,10 +1,15 @@
...
 
-[Almond] For Home Servers.
+## The Open-source, Privacy-Preserving Voice Assistant.
 
 ![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield] ![Supports armhf Architecture][armhf-shield] ![Supports armv7 Architecture][armv7-shield] ![Supports i386 Architecture][i386-shield]
 
-The Open, Privacy-Preserving Virtual Assistant.
+[Almond] provides a complete solution to turn your Home Assistant
+into a smart speaker. It supports playing music, controlling your
+IoTs, checking the weather, and other simple questions.
```
frenck, line 9: Spelling error: IoT is already multiple.

```suggestion
[Almond] provides a complete solution to turn your Home Assistant
into a smart speaker. It supports playing music, controlling your
IoT, checking the weather, and other simple questions.
```

almond/README.md
```diff
@@ -1,10 +1,15 @@
...
+
+To use Almond, you should connect a microphone and speaker to the
```
frenck, line 11: Right now, the add-on doesn't require that. Does that change?

almond/config.json
```diff
@@ -1,17 +1,16 @@
...
+  "slug": "edge",
```
frenck, line 4: We cannot change the slug

almond/config.json
```diff
@@ -1,17 +1,16 @@
...
+  "description": "The open-source, privacy-preserving voice assistant",
   "url": "https://github.com/home-assistant/hassio-addons/blob/master/almond",
-  "arch": ["armv7", "armhf", "aarch64", "amd64", "i386"],
+  "arch": ["armv7", "aarch64", "amd64"],
```
frenck, line 7: Any specific reason they are dropped?
I don't see any breaking change or information for the end-user about this in the changelog?

almond/build.json
```diff
@@ -1,12 +1,11 @@
 {
   "build_from": {
-    "aarch64": "homeassistant/aarch64-base:3.13",
-    "amd64": "homeassistant/amd64-base:3.13",
-    "i386": "homeassistant/i386-base:3.13",
-    "armv7": "homeassistant/armv7-base:3.13",
-    "armhf": "homeassistant/armhf-base:3.13"
+    "aarch64": "homeassistant/aarch64-base-debian:buster",
+    "amd64": "homeassistant/amd64-base-debian:buster",
+    "armhf": "homeassistant/armhf-base-debian:buster",
```
frenck, line 5: `armhf` was dropped according to the add-on configuration, but still in here?

almond/README.md
```diff
@@ -1,10 +1,15 @@
 # Home Assistant Add-on: Almond
 
-[Almond] For Home Servers.
+## The Open-source, Privacy-Preserving Voice Assistant.
```
frenck, line 3: There is no need to make this an heading.

almond/Dockerfile
```diff
@@ -5,50 +5,77 @@ FROM ${BUILD_FROM}
...
+        "https://github.com/stanford-oval/almond-server" . && \
+    sudo -u almond npm ci && \
+    sudo -u almond npm install --no-package-lock --ignore-scripts snowboy && \
+    cd node_modules/snowboy/ && \
+    sudo -u almond PYTHON=python2 npx node-pre-gyp clean configure && \
+    sudo -u almond make -C build/ && \
+    rm -fr /home/almond/.cache && \
+    rm -fr /home/almond/.npm && \
+    mkdir /home/almond/.cache \
+    && apt-get purge -y --auto-remove \
```
frenck, line 69: Please make the format consistent with the rest of our Dockerfile.

almond/Dockerfile
```diff
@@ -5,50 +5,77 @@ FROM ${BUILD_FROM}
...
-        --run \
-    && yarn global remove modclean \
+    && npm config set unsafe-perm \
     \
-    && yarn cache clean \
-    && apk del --no-cache \
-        .build-dependencies \
+    && sudo -u almond git clone -b "${ALMOND_VERSION}" --depth 1 \
+        "https://github.com/stanford-oval/almond-server" . && \
+    sudo -u almond npm ci && \
```
frenck, line 61: Sudo?

frenck (changes requested):

almond/CHANGELOG.md
```diff
@@ -1,62 +1,70 @@
...
-- Initial version
+# Changelog
+
+## 2.0.0-1
+
+- Almond was updated to 2.0.0: this is a major release that
+  brings significant changes to the supported skills. Additional
+  details are in the release notes: https://wiki.almond.stanford.edu/en/release-planning/two-point-oh
+- Voice support is now included in this addon, and the Ada add-on
+  is not required (in fact, it should not be enabled at the same time).
```
frenck, line 9: If that is critical, please ensure that during startup and provide a message to the end-user in the logs.

almond/Dockerfile
gcampax: It's unsafe to run as root, especially cause you're not using rootless docker. It seems pretty obvious to me.

almond/build.json
gcampax: Some of the prebuilt software we use depends on glibc.

almond/config.json
gcampax: It's a mistake to decouple version numbers from upstream software, and one that I hope to avoid with this addon. You and us both need to know what version of Almond is associated with the specific add-on build, otherwise triaging bugs becomes a nightmare.
Even if you consider yourself a "distro" (bad, outdated model) you'll want to track the upstream version and the build number separately.

almond/rootfs/etc/services.d/almond/run
gcampax: IIRC there was a race condition during initialization if this was done in the init script. Also, it cannot be a fixed file because it lives in the data partition.

almond/README.md
gcampax: Yes, the new addon is quite different than the old one, and uses the audio directly.

almond/config.json
gcampax: They are dropped because they do not build correctly, due to missing packages or binary blobs for those architectures. For example, the version of node was updated to 12, and there are no builds of node 12 for armv6.

almond/Dockerfile
gcampax: What does that mean?

almond/Dockerfile
gcampax: Yes. npm is very picky about running as root, and will generally break. So the source is downloaded and compiled as the regular user instead of root. We could use a USER directive but that would split the image into multiple layers and probably make it larger.

almond/CHANGELOG.md
gcampax: How do I check if a different add-on is enabled?

almond/DOCS.md
frenck: > We want users to reach out to upstream directly, to report issues and suggestions.

We don't. This was not a request, but a requirement of core integrations. Please adjust as requested.

almond/Dockerfile
frenck: That didn't answer my questions.

almond/build.json
frenck: So why not rebuild it?

almond/config.json
frenck: > It's a mistake to decouple version numbers from upstream software

It is not. please adjust.

almond/rootfs/etc/services.d/almond/run
frenck: Alternatively, store a file outside of the data folder and move it into place.

almond/README.md
frenck: So it cannot be used without? That would be a big breaking change.

almond/config.json
frenck: > node 12 for armv6

There is. Also available on Alpine.

almond/Dockerfile
frenck: Code styling is completely different. Please adjust it to match the code styling we use in our Dockerfile in our project.

almond/Dockerfile
frenck: The linter fails at this as well. There is no reason to use sudo here.

almond/CHANGELOG.md
frenck: Using the Supervisor API capabilities, this API is documented on our developers website.

almond/Dockerfile
frenck: > npm is very picky about running as root, and will generally break.

That is odd, we have quite a lot of add-ons that need NPM that don't show that problem. Any specific reason why this became a problem in this case?

almond/build.json
gcampax: I think the issue was snowboy (which uses glibc symbols on arm64) - that's a binary blob, there is nothing we can do about it.

almond/Dockerfile
gcampax: I don't know if snapshots work with a separate user (although I expect they will if the user ID doesn't change from one image to the next, which it shouldn't).

Now, reading the docker more carefully, the user is not actually used as runtime. It is only used at build time. I didn't write this part of the docker file originally, but I expect it has to do with npm which is broken when run as root. In that case, snapshots should be fine because files in the data file system will be owned by root. Is it ok to leave the non-root user here then?

almond/README.md
gcampax: It can be used without audio, it's just not very useful. I can clarify this in the readme. Do you have a suggested wording?

almond/config.json
gcampax: On debian, we need use the nodesource repository for node 12, and that doesn't have node for armv6. I mean, you saw the build failure, right?

almond/Dockerfile
frenck: If the user isn't used ad runtime, this stuff all isn't really needed and obsolete to begin with.
Please remove it.

almond/build.json
frenck: Why are we adding a closed source in here? Are we allowed to distribute it even?
How does this work for the future? Snowboy has been shutdown?

almond/Dockerfile
gcampax: Again, I think it *is* needed because npm does not run as root, and the build will not run correctly without this. I can test if you insist - but please keep in mind we went through a lot of iterations to make this work, so we have learned what it takes.

almond/build.json
gcampax: Yes snowboy is distributable. Snowboy is currently discontinued but it works. There are no open-source wake-word capabilities (Picovoice, used by Ada, is not open-source either, and it has no good general wake word).
In the future, we'll likely change the wakeword and the implementation, but it probably won't be open-source either.

almond/Dockerfile
frenck: As said, we use it with Z-Wave JS without issue. For my personal add-ons; Z-Wave 2 MQTT, NGinx Proxy Manager, Jupyter Lab (extensions), Node-RED, Home Panel.

So I'm not sure why it shouldn't work here?

almond/config.json
frenck: Yep, but now learned about Snowboy 👍 
So I guess no choice at this point.

almond/Dockerfile
gcampax: Is it npm 6? And do those packages require "prepare" steps? In my experience, npm 6 will not run "prepare" steps when it is run as root, which results in a botched build.

gcampax: Now, reading through upstream node issues (https://github.com/npm/cli/issues/2062 https://github.com/npm/cli/issues/1287 https://github.com/npm/cli/issues/1214 and others) it seems the bug only occurs with git dependencies. The released version of Almond has no git dependencies, so it's possible that we can remove the non-root user here. Development versions of Almond do have git dependencies though, and I'd rather keep the docker build as close as possible between development and stable versions, so we don't introduce bugs when we build the stable release.

frenck: Yes NPM 6.14+ on them all, including native extensions build with gyp

almond/README.md
frenck:
```suggestion
To use Almond as a voice assistant, you should connect a microphone and speaker to the
```

almond/rootfs/etc/services.d/almond/run
frenck: This comment has not been addressed yet.

pvizeli:
Only for clarification, Ada was never a replacement for Almond's voice assist. It was a concept to show how you can build a voice assistant based on tts/stt API of Home Assistant which allows to plug&play different provider of such service.

gcampax:
Yeah I don't mind that Ada continues to exist with that purpose. I know people have custom intents built with Home Assistant's conversation integration and those don't work through Almond. But the common case is for people to combine Ada and Almond (in fact, it is often conflated in social media and support forums) so we should document the change.

frenck (approved):
Tested it locally, including an upgrade path for existing users, which seems to work fine.

As for audio, I did get output, so that works. Input didn't work on my end, however, I blame my $1.50 USB mini mic for that, I don't see a reason for it being a general issue (@balloob and some others also tested the upstream version, which also showed it working).

So from my end, I think it is OK 👍

PR #2057, opened 2021-06-03
git_pull: Use ff-only when pulling

apeeters:
Fixes hint showing up in the logs on every pull. We can use ff-only since no local commits are to be expected.


ludeeus (changes requested):
This needs a version bump in config.json and an update in CHANGELOG.md

apeeters:
@ludeeus Added the requested changes: version bump & changelog.

michalchecinski (approved):

Issue #1353, opened 2020-05-25
MariaDB does not recover from HASS.io backup


Spirituss:
**Home Assistant release with the issue:**

Home Assistant 0.110.2

**Last working Home Assistant release (if known):**
Unknown (but it definitely worked earlier).

**Operating environment (Hass.io/Docker/Windows/etc.):**

Hass.io/Docker/Supervised

**Component/platform:**

MariaDB add-on used as record/history component.

**Description of problem:**
After recovery from Hass.io snapshot MariaDB doesn't work any more. The only way to force it to work is to uninstall the add-on and install it again with the same settings. Of course, in this case you lose your DB and history.

**Problem-relevant configuration.yaml entries and (fill out even if it seems unimportant):**

```
recorder: 
  db_url: mysql://hass:password@192.168.xxx.xxx/homeassistant?charset=utf8
  purge_interval: 7
  purge_keep_days: 7
```

**Traceback (if applicable):**

**Additional information:**
It worked fine on one of earlier version of hass.io. There are other issues containing relevant information that were closed without a solution:
https://github.com/home-assistant/core/issues/26152
https://github.com/home-assistant/hassio-addons/issues/667

frenck:
Anything in the logs?

Spirituss:
> Anything in the logs?

Different cases on log messages are inside #667 . I had similar one:
```
(MySQLdb._exceptions.OperationalError) (2002, "Can't connect to MySQL server on 'core-mariadb' (115)")
```.
So the only working solution was to remove MariaDB addon and install it again in order it to re-create DB loosing all previous history.

frenck:
A lot has changed in the meantime, I would really need recent add-on logs in order to get clue on where to start.

agners:
I mentioned it in the other thread, taking a snapshot from a database while the service is running just like that is **not** a proper backup operation.

Either the service must be shut down during snapshoting, or the steps documented in the "Filesystem Snapshots" need to be followed: https://mariadb.com/kb/en/backup-and-restore-overview/#filesystem-snapshots. Everything else is a hackaround.

frenck:
That is correct, unfortunately, the add-on is not aware of such a thing, and cannot handle it as suggested.

michaelarz:
[...]

PR #2063, opened 2021-06-08
Add lock capabilities during snapshot

agners:
This adds a new service `mariadb-lock` to the root file system. The service locks all tables, making it possible to copy the complete database while still running the database daemon. This is the documented procedure for file system snapshots (see https://mariadb.com/kb/en/backup-and-restore-overview/#filesystem-snapshots).

Note that the `mariadb` client must stay open.

Fixes: #1353


ludeeus (dismissed):
The version in config.json needs to be bumped, and the changelog.md file needs to be adjusted

agners:
@pvizeli thanks for addressing this, was about to do it too :)

pvizeli (approved):

PR #2071, opened 2021-06-12
Added options to configure DHCP failover with a seconds DHCP server

fschaeck:
This pull request just changes the way the options are processed in run.sh.
I added options to allow run.sh to generate a dhcpd.conf that contains all necessary definitions to have the DHCP server run in failover mode.

The way the options are added makes sure, that updating an already running addon of version 1.2 will change the user's configuration in a way, that the addon will work as before without needing any manual changes to the configuration, even though the default configuration will be added to the existing configuration automatically. Only if the user changes the option 'enable_failover' from its default value of 'false' to 'true' will the failover-options be taken into consideration by runs.sh when generating the dhcpd.conf. 

The addon documentation is updated accordingly and contains links to further information about DHCP failover configuration.

### DIFF
--- a/dhcp_server/DOCS.md
+++ b/dhcp_server/DOCS.md
@@ -28,7 +28,6 @@ dns:
   - 8.8.4.4
 default_lease: 86400
 max_lease: 172800
-enable_failover: true
 networks:
   - subnet: 192.168.1.0
     netmask: 255.255.255.0
@@ -40,7 +39,9 @@ networks:
     failover_peer: failover_partner
 failover_peers:
   - name: failover_partner
-    role: primary
+    # specifying mclt and split will make
+    # this server take the primary role
+    # ommiting both will make it secondary
     mclt: 60
     split: 128
     peer_address: 192.168.1.2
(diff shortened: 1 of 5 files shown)


homeassistant:
Hi @fschaeck, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2071).

Thanks!

dhcp_server/DOCS.md
```diff
@@ -28,6 +28,7 @@ dns:
   - 8.8.4.4
 default_lease: 86400
 max_lease: 172800
+enable_failover: true
```
pvizeli, line 31: Is that option needed? If there is no failover_peers, we can assume that is disabled

fschaeck: Hi @pvizeli!

Well... without that option I can't provide a default configuration in config.json for failover without screwing up any already existing user configuration in case of an update.

But I guess I could just provide an empty list for 'failover_peers' and don't add 'networks.failover_peer' at all in config.json. That would make the users configuration valid (schema fails if 'failover_peers' does not show up at all - there is no way to make a structured option element optional in terms of "no need to even specify an empty list", is there?), and thus avoid the need for 'enable_failover'.

After all the user can copy an example failover configuration from the documentation if need be...

So I will change the code accordingly and amend this pull request...

dhcp_server/config.json
```diff
@@ -1,13 +1,19 @@
...
   "arch": ["armhf", "armv7", "aarch64", "amd64", "i386"],
   "advanced": true,
   "startup": "system",
   "host_network": true,
+  "ports": {
+    "647/tcp": 647
+  },
+  "ports_description": {
+    "647/tcp": "DHCP failover handshake"
+  },
```
pvizeli, line 16: This makes no sense if you not read this data later with bashio and adjust the port based on the user settings

PR #2085, opened 2021-06-19
Add Lets Encrypt Auto Renewal to Docs

appleguru:


letsencrypt/DOCS.md
```diff
@@ -89,6 +89,30 @@ transip_username: ''
 transip_api_key: ''
 ```
 
+
+
+## Auto Renewal
+
+By default, this add-on will shut down once it runs and will **not** automatically renew your certificate. To automatically renew your certificate, it is recommended to run a home assistant automation that restarts this add-on at a convenient time daily or weekly so that your certificate is automatically renewed when it is time.
```
ludeeus, line 96: This is not enough.
This will update the certificates on disc (assuming it does not fail), but a restart of Home Assistant is required to start using the new certificates.

appleguru: While that’s true, most of us pretty regularly reboot HA (eg to update it). Is there a reasonable way to automate a restart or otherwise force a reload of the certs? (Eg, is restarting the nginx add on sufficient?)

ludeeus: _if_ you use that add-on, restarting it after you create new certs are enough (and needed), which brings in another point. Not everyone uses these certs the same way.

PR #2103, opened 2021-07-07
dnsmasq: enable dnssec, increase cache

RubenKelevra:
- the dnssec validation in dnsmasq makes sure signed domains are validated before handed to the requesting clients.
- the dnssec-check-unsigned makes sure that responses that are unsigned are validated from the root to make sure that there's no signature available
- the change in cache size from 150 to 100K leverages the large amount of memory a HA computer usually has, to reduce the response times and reduce the overhead of dnssec by caching much more requests
- skip the log of queries
- filters out queries for the non-routed address space (e.g. private networks)


RubenKelevra:
This PR requires dnsmasq to be built with dnssec support (haven't checked if this is already the case)

PR #2104, opened 2021-07-07
dnsmasq: extend the default dns servers to more than ipv4-only google

RubenKelevra:
Currently, the default DNS servers are only two ipv4 addresses from Google.

This PR extends this list to more providers which all support dnssec and do no filtering.

[Source](https://en.wikipedia.org/wiki/Public_recursive_name_server)

Providers:
- (unfiltered) AdGuard
- Cloudflare
- Google
- Neustar
- OpenNIC


pvizeli:
IPv6 is not working currently with docker on the internal network

RubenKelevra:
@pvizeli Ah allright, but this shouldn't be an issue, as dnsmasq just skips the non-working servers.

pvizeli:
right, but why we should include some things which is not working?

ludeeus (commented):
I do not think we should merge this.
The current defaults are fine for most installations.
For those that want more/other they can add/change that themselves, we should not do that for them.

PR #2111, opened 2021-07-08
Wrap provider configurations in details blocks

ludeeus:
Wrap the example configurations in details blocks, IMO this makes it more readable

image


frenck (approved):
Nice!

PR #2116, opened 2021-07-12
Add documentation for configuring RFC2136 provider

derekatkins:
Fix for Issue #1529 for LetsEncrypt Add-On; Replacement for PR#1763

This PR adds documentation for how to properly configure RFC2136 DNS-based certbot using a DNS server you control. The documentation is necessary because there are a few "gotchas" in the implementation, namely:

    The port specification is required (even if you're using the default port)
    The port specification schema is expecting a string, not a number, so it must be quoted
    In some cases some users have reported that using the FQDN for the DNS server does not work properly.

Improving the documentation will improve the user experience and save users time when trying to configure this mode of operation.

NB: changing the schema could be a breaking change to all existing users, so documenting the current behavior is IMHO safer.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -377,24 +377,31 @@ transip_api_key: ''
   <summary>RFC2136</summary>
 
   You will need to set up a server with RFC2136 (Dynamic Update) support with a TKEY (to authenticate the updates).  How to do this will vary depending on the DNS server software in use.  For Bind9, you first need to first generate an authentication key by running
+  
   ```
-  # dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST letsencrypt
+  $ dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST letsencrypt
   Kletsencrypt.+165+20675
   ```
+  
   The key file (Kletsencrypt.+165+20675.key in this example) looks like the following:
+  
   ```
-  # cat Kletsencrypt.+165+20675.key
+  $ cat Kletsencrypt.+165+20675.key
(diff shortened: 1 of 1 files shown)


letsencrypt/DOCS.md
```diff
@@ -373,6 +373,55 @@ transip_api_key: ''
...
 
+<details>
+  <summary>RFC2136</summary>
+
+  You will need to set up a server with RFC2136 (Dynamic Update) support with a TKEY (to authenticate the updates).  How to do this will vary depending on the DNS server software in use.  For Bind9, you first need to first generate an authentication key by running
+  ```
+  # dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST letsencrypt
+  Kletsencrypt.+165+20675
+  ```
+  The key file (Kletsencrypt.+165+20675.key in this example) looks like the following:
```
ludeeus, line 384: There should be a blank line around code blocks
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines

ludeeus: Code that are expected to be executed should generally not have a prefix <https://developers.home-assistant.io/docs/documenting/create-page#embedding-code>
Pasting it with `#` would not do anything, if separation is needed for command/output, use `$`

ludeeus (changes requested):

letsencrypt/DOCS.md
derekatkins: Okay, pushed a change that fixes both of these issues.
Anything else you see?
Thanks!

ludeeus (approved):
Thanks @derekatkins 👍

PR #2134, opened 2021-07-29
Bump zwave-js-server to 1.9.2

raman325:
Changelog: https://github.com/zwave-js/zwave-js-server/releases/tag/1.9.2

Main change here is to fix the ability to update log levels on the fly in the HA Configuration UI


MartinHjelmare (approved):
Thanks!

PR #2176, opened 2021-09-09
Bump zwave-js to 8.2.3

raman325:
Changelog:
- https://github.com/zwave-js/node-zwave-js/releases/tag/v8.2.2
- https://github.com/zwave-js/node-zwave-js/releases/tag/v8.2.3

Main reason for this bump is because there was a bug in 8.2.1 that affected insecure inclusion for devices that support S2


MartinHjelmare (approved):
Thanks!

PR #1689, opened 2020-11-21
SSH: Add persistent TMUX settings.

rigoorozco:
Updated profile.sh to copy the default .tmux.conf to the /data directory and then create a symbolic link to it at /root. This allows changes to TMUX bindings and other settings to be persistent across reboots.

PR #2191, opened 2021-09-23
SSH persistent tmux config

ddemidov:
This is basically #1689 with the updated changelog, as requested by @pvizeli.

### DIFF
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.2.1
+
+- Make tmux configuration persistent
+
 ## 9.2.0
 
 - Upgrade to Alpine 3.14
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 9.2.1
+## 9.3.0
 
 - Make tmux configuration persistent
 
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 9.3.0
 
-- Make tmux configuration persistent
+- Persist tmux configuration 
 
 ## 9.2.0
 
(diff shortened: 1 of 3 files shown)


homeassistant:
Hi @ddemidov, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2191).

Thanks!

ssh/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.2.1
```
ludeeus, line 3:
```suggestion
## 9.3.0
```

ssh/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+- Make tmux configuration persistent
```
ludeeus, line 5:
```suggestion
- Persist tmux configuration 
```

ludeeus (changes requested):
This needs a version bump in config.json

ddemidov:
> This needs a version bump in config.json

done (714ca80a0e8a630fccb45d9008ff843da35bcabc)

ssh/rootfs/etc/cont-init.d/profile.sh
```diff
@@ -10,6 +10,14 @@ if ! bashio::fs.file_exists /data/.bash_profile; then
     chmod 600 /data/.bash_history
 fi
 
+# Persist tmux configuration by redirecting .tmux.conf to /data
```
ludeeus, line 13: Thinking about this change a bit more...
If we add this, we can no longer adjust the default configuration for existing installations.
Which will result in new installations behaving differently from existing ones if we need to do that.

@pvizeli what do you think?

ddemidov: The reason I picked up the PR is that I did not like the default tmux configuration, and there was no way to update it in a way that would survive a reboot. I guess another option is to put the contents of `tmux.conf` into the addon config?

ludeeus: I think it's best to wait with anything like this until <https://github.com/home-assistant/supervisor/issues/1842> is implemented.

ddemidov: That indeed looks promising. Should I close this?

PR #2199, opened 2021-09-27
SSH: Upgrade CLI to 4.14.0

pvizeli:


frenck (approved):

ssh/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.2.1
+
+- Upgrade Home Assistant CLI to 4.14.0
```
ludeeus, line 5: This includes the datadisk commands which require a supervisor version that is currently not on stable

ludeeus: I can't suggest for some reason :thinking: 
Should we make the version a link?
`[4.14.0](https://github.com/home-assistant/cli/releases/tag/4.14.0)`

ludeeus: ah, it's merged already :+1:

pvizeli: right, but also not will break anything

PR #2203, opened 2021-09-28
Adjust documentation for lets_encrypt.algo

### LINKED ISSUE #2202: Addon: DuckDNS - Algo Option
After Updating to 1.14.0 i noticed that the new optional Option "lets_encrypt.algo" is not correctly documented.
The Documentaiton say that this is optional, but after updating i can`t start the Addon till i add the line "algo: secp384r1" to my DuckDNS (Lets Encrypt) configuration.

Maybe someone can add this additional Line to the Documentaion-examples for other users, so they dont need to try and search for a working config and check if its really an optional Setting.

DuckDNS 1.14.0
OS core-2021.9.7

- Add-on with the issue: DuckDNS
- Add-on release with the issue: 1.14.0
- Last working add-on release (if known): <
- Operating environment (OS/Supervised): OS core-2021.9.7

Error after updating: "[supervisor.addons.addon] Add-on core_duckdns has invalid options: Missing option 'algo' in lets_encrypt in Duck DNS (core_duckdns). Got {'lets_encrypt': {'accept_terms': True, 'certfile': 'XXXXXX', 'keyfile': 'XXXXXXX'}, 'token': 'XXXXXXXXXXXX', 'domains': ['XXXXXXXXXXX'], 'aliases': [], 'seconds': 300}"

ludeeus:
Closes #2202


driehle:
@ludeeus That doesn't look like a good fix to me. Why would you require all users to specify an algorithm if most users probably have no idea what that means? I think the add-on should use a default algorithm.

Hence, I suggest in [run.sh](https://github.com/home-assistant/addons/blob/master/duckdns/data/run.sh) to replace

```
ALGO=$(bashio::config 'lets_encrypt.algo')
```

with

```
if bashio::config.has_value "lets_encrypt.algo"; then ALGO=$(bashio::config 'lets_encrypt.algo'); else ALGO="secp384r1"; fi
```

and in [config.json](https://github.com/home-assistant/addons/blob/master/duckdns/config.json) to replace

```
      "algo": "list(rsa|prime256v1|secp384r1)"
```

with 

```
      "algo": "list(rsa|prime256v1|secp384r1)?"
```

to actually mark the parameter optional.

ludeeus:
The merging of the config has been fixed in the supervisor.
This PR _only_ adjusts the documentation.
#2202 is about documentation hence this PR fixes that.

driehle:
Thanks for the clarification. If the config merging for nested settings has been fixed in the supervisor, then why is this PR needed? Shouldn't the parameter be optional if the config merging has been fixed?

ludeeus:
>  then why is this PR needed?

The current documentation is wrong

> Shouldn't the parameter be optional if the config merging has been fixed?

No

driehle:
Why shouldn't the parameter be optional? We probably have a different understanding of "optional" though. To me, if the parameter has a default value and the user does not need to specify the parameter, I'd call it optional. Obviously, in the code the parameter will always be present, as it has a default value.

I think in the docs that should be called "optional", because the docs is what HA displays me in the tab "Configuration" of the DuckDNS add-on.

ludeeus:
It should exist. The supervisor now makes sure it exists by properly merging.

driehle:
> It should exist. The supervisor now makes sure it exists by properly merging.

Agreed! But exactly this makes the parameter optional for me, as an end user. And whom is the docs for if not for end users? (as said, they're displayed in HA, right beside the add-on configuration tab...)

ludeeus:
No, it does not, it still has to be in the configuration that you as the user see/manage/control.

frenck (approved):

duckdns/DOCS.md
```diff
@@ -62,14 +62,11 @@ The name of the private key file generated by Let's Encrypt. The private key fil
...
 **Note**: _The file is stored in `/ssl/`, which is the default for Home Assistant_
 
-#### Option `lets_encrypt.algo` (optional)
+#### Option `lets_encrypt.algo`
 
 Public key algorithm that will be used.
 
-Supported values: `rsa`, `prime256v1` and `secp384r1`. 
-
-The default is `secp384r1`
```
inverse, line 71: Can't this be configured as the default in the config of the addon?

PR #2207, opened 2021-09-30
Bump zwave-js-server to 1.10.6

raman325:
https://github.com/zwave-js/zwave-js-server/releases/tag/1.10.6


MartinHjelmare (approved):
Thanks!

PR #2208, opened 2021-09-30
Properly handle security key casing in zwave_js

raman325:
Fixes https://github.com/home-assistant/core/issues/56810#issuecomment-931399748

Basically we were transforming the key from lower case to upper case but that caused a mismatch. Now we don't transform the config, but we do a case insensitive match between keys (`s0_legacy_key` and `network_key` as well as all keys against the documented examples)

### DIFF
--- a/zwave_js/rootfs/etc/cont-init.d/config.sh
+++ b/zwave_js/rootfs/etc/cont-init.d/config.sh
@@ -45,7 +45,7 @@ fi
 # Validate that no keys are using the example from the docs and generate new random
 # keys for any missing keys.
 for key in "s0_legacy_key" "s2_access_control_key" "s2_authenticated_key" "s2_unauthenticated_key"; do
-    network_key="$(bashio::config "${key}")"
+    network_key=$(bashio::config "${key}")
     network_key_upper=$(bashio::string.upper "${network_key}")
     if [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_1}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_2}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_3}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_4}" ]; then
         bashio::log.fatal
--- a/zwave_js/rootfs/etc/cont-init.d/config.sh
+++ b/zwave_js/rootfs/etc/cont-init.d/config.sh
@@ -75,6 +75,12 @@ if [[ ${flush_to_disk:+x} ]]; then
     bashio::addon.options > "/data/options.json"
 fi
 
+# We have to reflush the config to disk if we have set the network_key
+if ! bashio::config.has_value 'network_key'; then
+    bashio::addon.option 'network_key' "$(bashio::config 's0_legacy_key')"
+    bashio::addon.options > "/data/options.json"
+fi
+
 s0_legacy=$(bashio::config "s0_legacy_key")
 s2_access_control=$(bashio::config "s2_access_control_key")
 s2_authenticated=$(bashio::config "s2_authenticated_key")
--- a/zwave_js/rootfs/etc/cont-init.d/config.sh
+++ b/zwave_js/rootfs/etc/cont-init.d/config.sh
@@ -3,6 +3,12 @@
 # Generate Z-Wave JS config file
 # ==============================================================================
 declare network_key
+declare s0_legacy_key
+declare s0_legacy
+declare s2_access_control
+declare s2_unauthenticated
+declare log_level
+declare flush_to_disk
 
 readonly DOCS_EXAMPLE_KEY_1="2232666D100F795E5BB17F0A1BB7A146"
 readonly DOCS_EXAMPLE_KEY_2="A97D2A51A6D4022998BEFC7B5DAE8EA1"
--- a/zwave_js/rootfs/etc/cont-init.d/config.sh
+++ b/zwave_js/rootfs/etc/cont-init.d/config.sh
@@ -6,6 +6,7 @@ declare network_key
 declare s0_legacy_key
 declare s0_legacy
 declare s2_access_control
+declare s2_authenticated
 declare s2_unauthenticated
 declare log_level
 declare flush_to_disk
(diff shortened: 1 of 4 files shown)


zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -14,7 +14,9 @@ if bashio::config.has_value 'network_key'; then
     # we don't know which one to pick so we have to exit. If they are both set
     # and do match, we don't need to do anything
     if bashio::config.has_value 's0_legacy_key'; then
-        if bashio::config.equals 's0_legacy_key' "$(bashio::config \"network_key\")"; then
+        network_key=$(bashio::string.upper "$(bashio::config 'network_key')")
+        s0_legacy_key=$(bashio::string.upper "$(bashio::config 's0_legacy_key')")
```
MartinHjelmare, line 18: We declare `network_key` at the top of the file. Is that something we need to do for `s0_legacy_key` too? I only know minimal bash, so just asking about things I see.

raman325: I am not actually sure what the `declare` does. In Googling around it seems optional, but I agree that either we should do it for all or do it for none. It seems to be a pattern to do this in other addons, so I will add the missing declarations

ludeeus (commented):
Since the network_key is not optional, it's schema should change.
`"match(|[0-9a-fA-F]{32,32})?"` -> `"match(|[0-9a-fA-F]{32,32})"` (remove `?`)

raman325:
`network_key` *is* optional though - it was an old option that we migrate to a new option if needed. All keys can be left blank and the addon will generate keys for each and then run.

ludeeus:
It breaks all past core installations if not present, since the addon support core down to 2021.2.0b0 it's not optional

MartinHjelmare:
The init script generates a key if the user hasn't set one, so for the user it's optional but in the end we guarantee that it's present.

ludeeus:
In that case, the default should be removed.

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -67,6 +75,12 @@ if [[ ${flush_to_disk:+x} ]]; then
     bashio::addon.options > "/data/options.json"
 fi
 
+# We have to reflush the config to disk if we have set the network_key
+if ! bashio::config.has_value 'network_key'; then
+    bashio::addon.option 'network_key' "$(bashio::config 's0_legacy_key')"
+    bashio::addon.options > "/data/options.json"
+fi
```
ludeeus, line 82: The if check should be moved up, and should set `flush_to_disk=1` instead of writing directly here

raman325: the problem was that I couldn't access the newly set option until I flushed to disk the first time to handle the case where `s0_legacy_key` is unset initially

raman325: With that being said, I can move this into the auto generation logic so we only flush once

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -44,8 +45,9 @@ fi
 # Validate that no keys are using the example from the docs and generate new random
 # keys for any missing keys.
 for key in "s0_legacy_key" "s2_access_control_key" "s2_authenticated_key" "s2_unauthenticated_key"; do
-    network_key=$(bashio::string.upper "$(bashio::config \"${key}\")")
-    if [ "${network_key}" == "${DOCS_EXAMPLE_KEY_1}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_2}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_3}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_4}" ]; then
+    network_key="$(bashio::config "${key}")"
+    network_key_upper=$(bashio::string.upper "${network_key}")
+    if [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_1}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_2}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_3}" ] || [ "${network_key_upper}" == "${DOCS_EXAMPLE_KEY_4}" ]; then
```
ludeeus, line 50:
```suggestion
if [[ " ${DOCS_EXAMPLE_KEYS[*]} " =~ " ${network_key_upper} " ]]; then
```
and:
```
readonly -a DOCS_EXAMPLE_KEYS=(
        "2232666D100F795E5BB17F0A1BB7A146",
        "A97D2A51A6D4022998BEFC7B5DAE8EA1",
        "309D4AAEF63EFD85967D76ECA014D1DF",
        "CF338FE0CB99549F7C0EA96308E5A403"
)
```

For another PR

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -36,8 +45,9 @@ fi
 # Validate that no keys are using the example from the docs and generate new random
 # keys for any missing keys.
 for key in "s0_legacy_key" "s2_access_control_key" "s2_authenticated_key" "s2_unauthenticated_key"; do
-    network_key=$(bashio::config "${key}")
-    if [ "${network_key}" == "${DOCS_EXAMPLE_KEY_1}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_2}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_3}" ] || [ "${network_key}" == "${DOCS_EXAMPLE_KEY_4}" ]; then
+    network_key="$(bashio::config "${key}")"
```
ludeeus, line 48:
```suggestion
    network_key=$(bashio::config "${key}")
```

ludeeus (approved):

MartinHjelmare (approved):
Looks good!

Have we tested running this locally?

raman325:
Failure scenarios I tested:
- Providing a lower case version of one of the keys from the docs -> this causes the addon to exit with the right log message
- Providing different keys for `network_key` and `s0_legacy_key` -> this causes the addon to exit with the right log message

Success scenarios I tested:
- Providing a key for `s0_legacy_key` and verifying that it copies to `network_key` (`s2_authenticated_key` was also provided, the rest were autogenerated)
- Providing a key for `network_key` and verifying that it copies to `s0_legacy_key` (`s2_authenticated_key` was also provided, the rest were autogenerated)
- Providing the same key for `network_key` and `s0_legacy_key` but with different casing -> the key check worked successfully

MartinHjelmare:
Thanks!

PR #2215, opened 2021-10-03
Update CHANGELOG.md

mkaatman:
Added release date context


ludeeus:
Why is this added? 
Why only for this add-on?

frenck:
I think the current changelog is fine.
Adding this to just a single add-on makes no sense at this point.

I suggest to close up this PR.

PR #2217, opened 2021-10-03
Bump zwave-js-server to 1.10.7

raman325:
https://github.com/zwave-js/zwave-js-server/releases/tag/1.10.7

Note that we are bumping the schema one more time to force users onto the latest to use the new compression option

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-# 0.1.45
+## 0.1.45
 
 - Bump Z-Wave JS Server to 1.10.7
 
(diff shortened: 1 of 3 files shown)


zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+# 0.1.45
```
MartinHjelmare, line 3:
```suggestion
## 0.1.45
```

MartinHjelmare (commented):
Note that users will be forced to upgrade once we bump the minimum server version in the client.

MartinHjelmare (approved):

zwave_js/CHANGELOG.md
raman325: I do this everytime! 🤦🏾

raman325:
> Note that users will be forced to upgrade once we bump the minimum server version in the client.

Yes - we were already forcing users to update as part of this next HA release so it shouldn't be an issue

PR #2232, opened 2021-10-11
Correct port to 4212 for VLC discovery service

mib1185:
The default port for telnet interface of VLC is 4212, those this should be used for the discovery.
btw. I'm working on the config flow for the [`vlc_telnet`](https://www.home-assistant.io/integrations/vlc_telnet) integration incl. the discovery step via hassio, so that the discovery via this addon will work ... will link the PR soon here

### DIFF
--- a/vlc/CHANGELOG.md
+++ b/vlc/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.3
+
+- Correct port on discovery
+
 ## 0.1.2
 
 - Revert restart nginx service on error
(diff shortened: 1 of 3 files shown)


mib1185:
uhhh ... I did not know that ... I hope it is OK, that I would like to contribute to your branch, those I have already created home-assistant/core/pull/57501 🙂

MartinHjelmare:
I think this PR is ok. The port is a leftover from the zwave_js add-on this add-on is based on. There are some more leftovers that we can clean up.

PR #2259, opened 2021-11-08
Reset Authorization header on proxied request

### LINKED ISSUE #2014: Username with colon is NG in Mosquitto broker 6.0.0
## The problem

I'm using a locally configured Meross device. Meross devices use a MAC address with a colon in the username.

Until 5.1.1, a username with a colon was allowed, but from 6.0.0 it became unavailable and the Meross device became unusable.

## Environment

- Add-on with the issue: Mosquitto broker
- Add-on release with the issue: 6.0.0
- Last working add-on release (if known): 5.1.1
- Operating environment (OS/Supervised): supervisor-2021.04.3

```yaml
logins:
  - username: '48:e1:e9:xx:xx:xx'
    password: meross_eb61de8166d394ecba0bc84cab7axxxx
anonymous: false
customize:
  active: false
  folder: mosquitto
certfile: fullchain.pem
keyfile: privkey.pem
require_certificate: false
```

## Traceback/Error logs

```txt

```

## Additional information
The username is fixed in the device firmware and cannot be changed by us.

bytespider:
Home Assistant allows a much richer set of characters for username, however the Mosquitto auth plugin sends both Basic Auth and a URL encoded parameter string.

HA will parse the Basic Authentication header first and fail, while the later URL encoded string authorization will not.

In order to skip the more restrictive Basic Authentication check, I've updated the Nginx proxying to remove that header.

This fixes #2014

### DIFF
--- a/mosquitto/rootfs/usr/share/tempio/nginx.gtpl
+++ b/mosquitto/rootfs/usr/share/tempio/nginx.gtpl
@@ -35,6 +35,7 @@ http {
 
     location /authentication {
       proxy_set_header        X-Supervisor-Token "{{ env "SUPERVISOR_TOKEN" }}";
+      proxy_set_header        Authorization "";
       proxy_pass              http://supervisor/auth;
     }
 
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.0.2
+
+- Fix authorization of MQTT devices using colon in username (issue: #2014)
+
 ## 6.0.1
 
 - Fix loading custom mosquitto configuration
--- a/mosquitto/config.yaml
+++ b/mosquitto/config.yaml
@@ -1,4 +1,4 @@
-version: 6.0.1
+version: 6.0.2
 slug: mosquitto
 name: Mosquitto broker
 description: An Open Source MQTT broker


bytespider:
This fixes https://github.com/home-assistant/addons/issues/2014

bytespider:
Is there anything more I need to do to get this reviewed?

prosper86:
Hi please explain for Me on the picture

bytespider:
@MartinHjelmare is anyone interested in this? I believe it's the correct solution as it brings the known behaviour from 5.x back into 6.x devices could no longer authenticate if they contained non-alphanumeric characters.

mosquitto/rootfs/usr/share/tempio/nginx.gtpl
```diff
@@ -35,6 +35,7 @@ http {
 
     location /authentication {
       proxy_set_header        X-Supervisor-Token "{{ env "SUPERVISOR_TOKEN" }}";
+      proxy_set_header        Authorization "";
```
frenck, line 38: This is incorrect, this will break HA authentication.

frenck (changes requested):

mosquitto/rootfs/usr/share/tempio/nginx.gtpl
bytespider: @frenck can you please elaborate how?

As far as the code in code when this was developed, the code checked first the Basic Auth, then the parameters that were passed in the URL.

HA uses Mosquitto Password and ACL files and doesn't authenticate via HTTP from this addon afaik.

bytespider: @frenck wondering if you saw my reply to you last response?

frenck: @bytespider I missed that indeed. Sorry about that.

> and doesn't authenticate via HTTP from this addon afaik.

It does :) It authenticates against Home Assistant with Home Assistant user accounts.

bytespider: Okay, thank you.

I'll go through the code again as I clearly missed something. I thought I understood it.

bytespider: Okay,
So perhaps I'm wrong, but...

The HA mosquito user is set in the PW file
https://github.com/home-assistant/addons/blob/master/mosquitto/rootfs/etc/cont-init.d/mosquitto.sh#L33-L36 and doesn't authenticate from Mosquitto to the HA API.

And even if it did, Skipping Basic Auth would be fine as the Mosquitto Auth Plugin sends along a URL encoded string with the credentials that will get checked as long as the Authorization header is not present.
https://github.com/home-assistant/supervisor/blob/4416b6524eb120fc1456a305cf6dda49557c32cd/supervisor/api/auth.py#L59-L83

frenck: It does: 

https://github.com/home-assistant/addons/blob/master/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl#L11-L30

bytespider: > It does:
> 
> https://github.com/home-assistant/addons/blob/master/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl#L11-L30

All this says is that it uses the HA HTTP API.
As far as I understand, That API can authenticate a user via Basic Auth, URL encoded form variables, or JSON body.

The Mosquitto Auth Plugin sends both Basic Auth and URL encoded form variables, when only one is required by HA Auth API
https://github.com/pvizeli/mosquitto-auth-plug/blob/master/src/be-http.c#L178-L197.

My change causes NGinx to remove the Authorization header from the request, thus bypassing the Basic Auth check and move onto authenticating the user via the  URL encoded form variables.

frenck: But that would mean this would work around a Supervisor API issue? Or an issue in the Auth plugin? 

As in, Basic Auth versus URL encoded form variables should result in the exact same behavior?

bytespider: This is true. However Basic Auth only allows for a limited set of characters, whilst Mosquitto usernames and passwords are more diverse.

What would the impact of retiring Basic Auth from the Supervisor Auth API be?

I wonder if it's worth having the Auth Plugin control which method is sent to the backend HTTP server? @pvizeli thoughts?

frenck: > What would the impact of retiring Basic Auth from the Supervisor Auth API be?

It is not about impact, it is about where this should be fixed. I feel like we are now working around a problem that shouldn't be "fixed" here, but on another place.

bytespider: Sorry perhaps I was unclear, I was suggesting that one option was perhapst that the Supervisory API no longer supports Basic Auth as a way or authorizing user credentials since it cannot handle the full set of characters supported by HA usernames and passwords.

frenck: > I was suggesting that one option was perhapst that the Supervisory API no longer supports Basic Auth

It was introduced as the last of all authentication handling and actually used by almost all add-ons.

So thinking about the specifications and limitations, this is about the colon `:` I guess.
I suggest ensuring we don't accept colons in usernames. Reason why: This is making it work for this special case, but that doesn't solve the case that Home Assistant currently allows for it (apparently) and the Supervisor can't handle it (apparently).

TL;DR; this PR works around an issue, and is not address or fixing the greater problem behind it. Therefore, I don't think this PR should be merged.

bytespider: Disappointing conclusion, but thank you.

Issue #2252, opened 2021-11-05
Update Zwave-JS to 1.46 crashing HA heavely


canedje:
## The problem
Zwave-js update to 1.46 did heavely crash the system.
Even a restore of a backup is not working
Updating the VM running HA gives an error the zwave USB is not recognized anymore.
I had to run a working backup at HA and restart the NUC for getting up and running

## Environment

- Add-on with the issue: Zwave-js
- Add-on release with the issue: 1.46
- Last working add-on release (if known):  1.45
- Operating environment (OS/Supervised): 
## System Health

version | core-2021.11.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.9.7
os_name | Linux
os_version | 5.10.75
arch | x86_64
timezone | Europe/Amsterdam

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
Github API Calls Remaining | 4268
Installed Version | 1.16.0
Stage | running
Available Repositories | 889
Installed Repositories | 29

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | failed to load: timeout

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 6.6
-- | --
update_channel | stable
supervisor_version | supervisor-2021.10.8
docker_version | 20.10.8
disk_total | 48.5 GB
disk_used | 7.6 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
installed_addons | File editor (5.3.3), NGINX Home Assistant SSL proxy (3.0.2), Node-RED (10.0.1), Samba share (9.5.1), Terminal & SSH (9.2.1), Z-Wave JS (0.1.45), Samba Backup (5.0.0), ESPHome (2021.10.3), Duck DNS (1.14.0)

</details>

<details><summary>Lovelace</summary>

dashboards | 3
-- | --
resources | 21
views | 97
mode | storage

</details>
## Problem-relevant configuration

```yaml

```

## Traceback/Error logs

![image](https://user-images.githubusercontent.com/23150947/140574785-585b4789-e78c-4b56-92d0-8e29b0ad4c30.png)

```txt
   manufacturer ID:     0x86
                                    product type:        0x01
                                    product ID:          0x5a
                                    supported functions: 
[...]

PR #2260, opened 2021-11-08
Add Zwave JS disable_soft_reset

aximusnl:
Fixes: https://github.com/home-assistant/addons/issues/2252

It adds a configuration option to be able to disable the Zwave-JS soft reset. I am unable to test it, it's based on the documentation:

```
	/**
	 * Soft Reset is required after some commands like changing the RF region or restoring an NVM backup.
	 * Because it may be problematic in certain environments, we provide the user with an option to opt out.
	 * Default: `true,` except when ZWAVEJS_DISABLE_SOFT_RESET env variable is set.
	 */
	enableSoftReset?: boolean;
```

My first contribution to an open source project, please be gentle :)

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.47
+
+- Add disable_soft_reset option
+
 ## 0.1.46
 
 - Bump Z-Wave JS to 8.7.5
--- a/zwave_js/config.json
+++ b/zwave_js/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Z-Wave JS",
-  "version": "0.1.46",
+  "version": "0.1.47",
   "slug": "zwave_js",
   "description": "Control a ZWave network with Home Assistant Z-Wave JS",
   "arch": ["amd64", "i386", "armhf", "armv7", "aarch64"],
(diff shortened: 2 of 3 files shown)


homeassistant:
Hi @aximusnl, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2260).

Thanks!

MartinHjelmare:
Thanks for the PR. We'd like to allow for an automatic approach where the user doesn't need to enable/disable options. PR for that here:
https://github.com/home-assistant/addons/pull/2261

I think it's ok to add an option for advanced users but it must then be compatible with the automatic approach. Ie we use the auto-disable on VM unless the user has chosen to enable/disable soft-reset explicitly.

aximusnl:
> Thanks for the PR. We'd like to allow for an automatic approach where the user doesn't need to enable/disable options. PR for that here: #2261
> 
> I think it's ok to add an option for advanced users but it must then be compatible with the automatic approach. Ie we use the auto-disable on VM unless the user has chosen to enable/disable soft-reset explicitly.

I like your approach. But there may still be edge cases where you may want to manually override this setting. For example; in some VM environments it’s not needed to disable soft reset (auto re-attach).

PR #2261, opened 2021-11-08
Disable zwave-js soft-reset on vm

MartinHjelmare:
- Check host chassis if vm is detected and in that case disable soft-reset in zwave-js.
- We want to avoid forcing users to configure options as far as possible. An option to allow advanced users running on VM to enable soft-reset may be added in the future but should then be compatible with the automatic disable mechanism. Ie we still want to disable soft-reset automatically unless the user has enabled the option specifically.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.47
+
+- Disable soft-reset if VM is detected
+
 ## 0.1.46
 
 - Bump Z-Wave JS to 8.7.5
--- a/zwave_js/config.json
+++ b/zwave_js/config.json
@@ -8,6 +8,7 @@
   "startup": "services",
   "init": false,
   "stage": "stable",
+  "hassio_api": true,
   "homeassistant": "2021.2.0b0",
   "ports": {
     "3000/tcp": null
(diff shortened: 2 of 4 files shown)


MartinHjelmare:
I have tested this locally both in a VM install and in an Odroid 2 install. It behaves as expected, ie soft-reset is disabled on VM and enabled on Odroid 2.

MartinHjelmare:
I haven't bumped version yet. I can do that when we've decided what PRs should go in the next release.

frenck (approved):
LGTM, as per comment in the other PR, I would opt to bump the version on this one. After the release of this detection, we can see if there is still need for a manual override or not.

KermitOfDarkness:
I have just upgraded from 1.45 to 1.47 on my DSM 7.1 VM. It works great, again ! Thanks a lot.

jaj75074:
When I upgrade the community add-on, even if it's set to not start, it breaks my usb connection which means I have to restore a backup of my VM.

How can I upgrade the add-on w/o the driver resetting the USB before I can disable it?

MartinHjelmare:
We don't support the community add-on here. Please ask questions in our help channels.

https://www.home-assistant.io/help/#communication-channels

Thanks!

PR #2268, opened 2021-11-14
Apple Device Support

ananthb:
Load vfs_fruit modules for better Apple device support.

### DIFF
--- a/samba/rootfs/usr/share/tempio/smb.gtpl
+++ b/samba/rootfs/usr/share/tempio/smb.gtpl
@@ -6,9 +6,6 @@
    security = user
    ntlm auth = yes
 
-   local master = yes
-   preferred master = yes
-
    load printers = no
    disable spoolss = yes
 
--- a/samba/rootfs/usr/share/tempio/smb.gtpl
+++ b/samba/rootfs/usr/share/tempio/smb.gtpl
@@ -6,6 +6,9 @@
    security = user
    ntlm auth = yes
 
+   local master = yes
+   preferred master = yes
+
    load printers = no
    disable spoolss = yes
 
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @ananthb, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2268).

Thanks!

tomkv:
Sorry to enter into this conversation with this, but Local Master should not be used.

Local Master is part of SMB1, which is obsolete and security risk. Microsoft has recommended not to use SMB1 for years (https://techcommunity.microsoft.com/t5/storage-at-microsoft/stop-using-smb1/ba-p/425858), and Windows 10 disables SMB1 support, if not used in few days after installation.

For modern discovery, other protocols are being used: WSD for Windows and Zeroconf/Bonjour for macOS and Linux. Samba itself doesn't implement WSD, but there is a third-party project that does (https://github.com/christgau/wsdd). Zeroconf/Bonjour is implemented by Avahi.

For a supported, future-proof discovery methods, I would suggest to implement these protocols, instead of enabling legacy one.

ananthb:
@tomkv I had no idea that was deprecated. I'll remove that and leave the fruit stuff in because it looks nicer on Macs. Would you be interested in me adding Zeroconf and WSD to samba? I'd like to tackle that.

tomkv:
@ananthb implementing WSD for Samba is another can of worms... see https://bugzilla.samba.org/show_bug.cgi?id=11473. I would suggest integrating the above mentioned wsdd instead. It would be also much easier ;). Synology also went this route (for wsd, they use their own solution, `synowsdiscoveryd`, but the same principle; for Zeroconf, they use Avahi with a simple config file - it basically advertises _smb._tcp service at port 445).

ananthb:
@tomkv I've left just the vfs_fruit changes here. I'll raise another PR for the discovery stuff once I dig into WSD a bit more.

PR #2281, opened 2021-11-23
Update-deployment

pvizeli:


.github/workflows/builder.yml
```diff
@@ -83,14 +83,14 @@ jobs:
...
-           fi
+          if [[ "${{ steps.info.outputs.architectures }}" =~ ${{ matrix.arch }} ]]; then
+             echo "::set-output name=build_arch::true";
+           else
+             echo "${{ matrix.arch }} is not a valid arch for ${{ matrix.addon }}, skipping build";
+          fi
 
-           version=$(jq -r '.version' "./${{ matrix.addon }}/config.json")
-           echo "::set-output name=version::${version}";
+          version=$(jq -r '.version' "./${{ matrix.addon }}/config.json")
```
ludeeus, line 92:
```suggestion
          version=$(jq -r '.version' "./${{ matrix.addon }}/config.yaml")
```

ludeeus: Line 33 also needs the same update, but can't suggest it.

ludeeus (changes requested):

.github/workflows/builder.yml
ludeeus: We are not using `steps.check.outputs.version` so those 2 version related lines can be removed

.github/workflows/builder.yml
```diff
@@ -83,14 +83,14 @@ jobs:
...
+          echo "::set-output name=version::${version}";
```
ludeeus, line 93:
```suggestion
```

.github/workflows/builder.yml
ludeeus: Line 33 still needs to be changed from  `.json` -> `.yaml`

ludeeus (approved):

PR #2307, opened 2021-12-15
Allow use of ports other than 8123 in HA

felipecrs:
I'm setting up the nginx-proxy in my HA installation just to leverage HTTP/2 because of [LL-HLS](https://www.home-assistant.io/integrations/stream/#ll-hls).

I, however, use port 8123 for external HTTPS access because my ISP blocks me from using port 443.

While I could map the nginx-proxy 443 port to something like 8124, I would prefer to maintain the 8123 port for HTTPS access, so that I don't need to re-configure all my devices.

In order to achieve that, I'm removing the hardcoded 8123 port from nginx.conf and fetching the correct port instead. So that, I would set:

HA's `http.server_port` to `8124`, and map nginx-proxy 443 port to 8123.

### DIFF
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -1,4 +1,4 @@
-version: 3.0.2
+version: 3.1.0
 slug: nginx_proxy
 name: NGINX Home Assistant SSL proxy
 description: An SSL/TLS proxy
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -1,4 +1,5 @@
 version: 3.1.0
+hassio_api: true
 slug: nginx_proxy
 name: NGINX Home Assistant SSL proxy
 description: An SSL/TLS proxy
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.1.0
+
+- Allow use of ports other than 8123 in Home Assistant Core
+
 ## 3.0.2
 
 - Update Alpine to 3.14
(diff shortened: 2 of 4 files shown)


homeassistant:
Hi @felipecrs, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2307).

Thanks!

ludeeus (commented):
The supervisor already knows which port to use <https://developers.home-assistant.io/docs/api/supervisor/endpoints#core> That can be fetched with bashio instead of adding a new config option.

felipecrs:
Awesome. I'll rework it.

felipecrs:
@ludeeus done.

ludeeus:
That was fast 👍 
You would need to rebase from master to resolve the CI failure.

felipecrs:
@ludeeus done.

felipecrs:
Ops sorry, I messed up. One sec.

felipecrs:
@ludeeus done for real now.

ludeeus:
Please don't squash and force push after a review has started.

ludeeus (changes requested):
The change looks OK to me.
Before we can merge it, you need to update the CHANGELOG.md file and bump the version in config.yaml

felipecrs:
@ludeeus done.

frenck (approved):
Agreed, looks good 👍

ludeeus (approved):
Thanks @felipecrs 👍

felipecrs:
> Please don't squash and force push after a review has started.

Oh, my bad. Some projects prefer this way so that they can use the rebase strategy for merging, but now I know what's the preferred option here.

nginx_proxy/config.yaml
```diff
@@ -1,4 +1,4 @@
-version: 3.0.2
+version: 3.1.0
```
ludeeus, line 1:
```suggestion
version: 3.1.0
hassio_api: true
```
Needed to use the supervisor API

Issue #2020, opened 2021-05-20
MQTT seems to have broken with latest 6.0.1 update


Rusti-gotrage:
## The problem

Since updating the MQTT broker in Supervisor, I've noticed connection issues with multiple applications talking to MQTT, especially the PAI add-on (Paradox Alarm Interface), but that might be because it's the most 'visible' to me...

I'm not seeing the log update with the connections like I used to. Demonstrated when I connect to MQTT using MQTT-Explorer, as well as when I restart Node-Red or PAI addons in Supervisor.
When PAI finally DOES deign to connect, I'm not getting the statuses of all the devices.

## Environment

Running v.6.0.1 of MQTT on HASS core-2021.5.5 on a Raspberry Pi 4

## Traceback/Error logs

And I keep seeing this in my HomeAssistant Core logs:
2021-05-20 16:49:59 ERROR (Thread-4) [homeassistant.components.mqtt] Unable to connect to the MQTT broker: Connection Refused: not authorised.

## Additional information

Stupidly I didn't do a snapshot backup before upgrading MQTT - another user in Discord had the same issues, but he DID a snapshot and was able to rollback and get everything working again...

frenck:
Please provide add-on configuration details, client configuration details, what versions are used and OS details.

I need a reproduction scenario in order to troubleshoot this report.

Rusti-gotrage:
PAI 2.5.2
MQTT 6.0.1
Core:   core-2021.5.5
Supervisor:   supervisor2021.04.3  Stable
Host:   OS  Home Assistant OS 5.13 Production

Redacted configs @  https://pastebin.com/Sp7HCvYv

ivankarpov:
Same here:
PAI 2.5.2
MQTT 6.0.1
Core: core-2021.5.5

ivankarpov:
After enabled verbose mode of mosquito found this:

1621585592: Denied PUBLISH from paradox_mqtt/3edb0d6834396faf (d0, q0, r1, m0, 'paradox/interface/run_status', ... (7 bytes))
1621585592: Denied PUBLISH from paradox_mqtt/3edb0d6834396faf (d0, q0, r1, m0, 'paradox/interface/availability', ... (7 bytes))
1621585592: Denied PUBLISH from paradox_mqtt/3edb0d6834396faf (d0, q0, r1, m0, 'paradox/interface/run_status', ... (7 bytes))
[...]

PR #2137, opened 2021-08-03
Let Mosquitto use SQLite over password file for auth

kr0ner:
The SQLite approach allows more flexibility when it comes to special characters, e.g. colon in username. The additional libs are minimal and changes to the existing code are just a few lines.

Possible fix for https://github.com/home-assistant/addons/issues/2020

PR #2308, opened 2021-12-17
Let Mosquitto use SQLite over password file for auth

kr0ner:
The SQLite approach allows more flexibility when it comes
to special characters, e.g. colon in username. The additional libs
are minimal and changes to the existing code are just a few lines.

Fixes https://github.com/home-assistant/addons/issues/2020
Replaces https://github.com/home-assistant/addons/pull/2137

### DIFF
--- a/mosquitto/Dockerfile
+++ b/mosquitto/Dockerfile
@@ -8,18 +8,22 @@ RUN apk add --no-cache \
         mosquitto \
         nginx \
         pwgen \
+        sqlite \
+        sqlite-libs \
     && apk add --no-cache --virtual .build-dependencies \
         build-base \
         curl-dev \
         git \
         mosquitto-dev \
         openssl-dev \
+        sqlite-dev \
     \
     && git clone --depth 1 -b "${MOSQUITTO_AUTH_VERSION}" \
         https://github.com/pvizeli/mosquitto-auth-plug \
     \
     && cd mosquitto-auth-plug \
     && cp config.mk.in config.mk \
+    && sed -i 's/BACKEND_SQLITE ?= no/BACKEND_SQLITE ?= yes/g' config.mk \
     && make \
     && mkdir -p /usr/share/mosquitto \
     && cp -f auth-plug.so /usr/share/mosquitto \
--- a/mosquitto/rootfs/etc/cont-init.d/mosquitto.sh
+++ b/mosquitto/rootfs/etc/cont-init.d/mosquitto.sh
@@ -2,8 +2,7 @@
 # ==============================================================================
 # Configures mosquitto
 # ==============================================================================
-readonly ACL="/etc/mosquitto/acl"
-readonly PW="/etc/mosquitto/pw"
+readonly DB="/etc/mosquitto/mosquitto.sqlite"
 readonly SYSTEM_USER="/data/system_user.json"
 declare cafile
 declare certfile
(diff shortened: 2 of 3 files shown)


mdegat01 (changes requested):
I tested it locally and confirmed that yes I could not use the username `test:test` before and now I can. However a couple things come to mind:

First - small change: Please set the value for `username` in `config.json` to `str(,100)` since you set the column size to 100. Can't set a max length for a `password` type field yet unfortunately but we should mention it in the docs. Let's try to do what we can to ensure we can't get in the situation where users enter usernames/passwords that validate but break later again.

Second - more significant: You're actually replacing one invalid character with another. `:` works now but `'` does not since the value is put directly into the SQL and `'` messes with the statement. Which is really another way of saying you're exposed to SQL injection here, what can we do to fix that?

mdegat01:
It's crude and ugly but I think if you ran the values through something like this before putting them in the SQL that should work:
```
echo "${value_for_sql}" | sed -e 's/\([\\'"'"']\)/\\\1/g'
```

frenck:
I honestly believe usernames should never contain colons in Linux world. I'm more in favor of properly adding a regular expression on the add-on configuration that prevents that from being used and document it as such as well.

kr0ner:
> I tested it locally and confirmed that yes I could not use the username `test:test` before and now I can. However a couple things come to mind:
> 
> First - small change: Please set the value for `username` in `config.json` to `str(,100)` since you set the column size to 100. Can't set a max length for a `password` type field yet unfortunately but we should mention it in the docs. Let's try to do what we can to ensure we can't get in the situation where users enter usernames/passwords that validate but break later again.
> 
> Second - more significant: You're actually replacing one invalid character with another. `:` works now but `'` does not since the value is put directly into the SQL and `'` messes with the statement. Which is really another way of saying you're exposed to SQL injection here, what can we do to fix that?

Thanks for the review. You are right, I've missed that. Will fix it, if we decide to continue or go with @frenck proposal.

mdegat01:
> I honestly believe usernames should never contain colons in Linux world (or any other special characters for that reason). I'm more in favor of properly adding a regular expression on the add-on configuration that prevents that from being used and document it as such as well.
> 
> Using SQL for this as a workaround is just odd (with other side-effect that now are introduced as well).

@frenck it's a fair point. And notably Mosquitto does mention this restriction in the description for `mosquitto_passwd` [here](https://mosquitto.org/man/mosquitto_passwd-1.html). But that is specifically for password files because of the parsing problem. The broker itself doesn't seem to have a restriction and it defers that decision to the auth plugin provider of your choice.

Adding a regex is obviously a very simple change but it is apparently breaking. Looking back in time prior to the `6.0.0` update it looks like on a connection attempt it would read in the local users from config and use `jq` to check against it ([here](https://github.com/home-assistant/addons/blob/d42e52b7b4e899419d7a29e1cf5acf96d87c2cb3/mosquitto/data/auth_srv.sh)). Since config is JSON all characters worked fine.

kr0ner:
> I honestly believe usernames should never contain colons in Linux world (or any other special characters for that reason). I'm more in favor of properly adding a regular expression on the add-on configuration that prevents that from being used and document it as such as well.
> 
> Using SQL for this as a workaround is just odd (with other side-effect that now are introduced as well).
> 
> 

I've chosen SQL cause it does not have any limitations on characters, if used correctly ^^ ... which we can achieve.

I totally agree that a colon should not be part of the username, but there are a lot of devices out there that just don't care. IMHO we should not punish users for the devs fault. Means I'm fine with dropping SQL, but then we need an alternative.

How about dropping ":" from the username and password when storing and querying for them?

mdegat01:
> How about dropping ":" from the username and password when storing and querying for them?

Dropping isn't an option. That creates unexpected conflicts like `test:test` actually equals `testtest`.

We could base64 encode or hash the username when putting it in the file and checking. But what makes me uncomfortable about that is we'd have to change [mosquitto-auth-plug](https://github.com/pvizeli/mosquitto-auth-plug) to be incompatible with `mosquitto_passwd` (which is supposed to be used to generate these password files). It would also be a breaking change in that library although I have no idea if anyone else is using that fork.

When you say there are devices that "just don't care" can you give an example? I read the issue about mac addresses being used as usernames but I thought people were entering those to keep track of their devices. Are you saying you have a device that will only talk to an MQTT broker using a specific username that contains colons and you can't change it? Can you share details on this deivce?

kr0ner:
> Dropping isn't an option. That creates unexpected conflicts like test:test actually equals testtest

I don't think that would be an issue, since we echo $USER:$PASSWORD to the file. So if we drop ":" from both username and password we still end up with "test:test". Although you are right with reading it back from the file ... so yep ... not an option

Changing the auth_plugin seems the wrong approach to me anyways. So we still have the possibility to roll back to jq as you mentioned, right?

E.g. the Meross devices (I have the MSS310) use the mac as username and are used by many in combination with https://github.com/krahabb/meross_lan
>  username: 48:e1:e9:60:ad:a9
    password: 0_b33496ea448d3d53a0c69214cecc2972

PR #2310, opened 2021-12-18
Fixed bash PS1 escaping

SteveEasley:
Fixes the bash prompt so it ignores non-printable characters (the color escape sequence in the PS1), which allows it to correctly calculates the size of the prompt. Without this fix, on longer lines pressing the home key, and cycling through the bash history is buggy.

Replicating: Type a long line like `echo abcdefghijklmnopqrstuvwxyz`, press enter, arrow up, then the home key. Note the cursor does not go all the way back to the start of the line.

ref: https://stackoverflow.com/questions/19092488/custom-bash-prompt-is-overwriting-itself

### DIFF
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.2.2
+
+- Fix escape codes in color bash prompt
+
 ## 9.2.1
 
 - Upgrade Home Assistant CLI to 4.14.0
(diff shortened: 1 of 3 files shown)


pvizeli:
Please update the changelog + update version + 0.0.1

frenck:
@SteveEasley You also need to bump the actual version in the add-on configuration file.

SteveEasley:
Fixed as well.

ludeeus (approved):
Thanks @SteveEasley 👍

Issue #2054, opened 2021-06-02
deCONZ+Phoscon:  The web page at http://hassio.local:8123/api/hassio_ingress/<long ident string>/pwa/index.html might be temporarily down or it may have moved permanently to a new web address.


dongursli:
## The problem
From inside http://hassio.local:8123/core_deconz/dashboard

Works: http://hassio.local:8123/api/hassio_ingress/\<long ident string\>/novnc/vnc_lite.html?scale=yes

Don't work with FQDN:
Can't open The web page at http://hassio.local:8123/api/hassio_ingress/\<long ident string\>/pwa/index.html might be temporarily down or it may have moved permanently to a new web address.

Works with IP: http://<ip-addr>:8123/api/hassio_ingress/\<long ident string\>/pwa/index.html

## Environment

- Add-on with the issue: deCONZ
- Add-on release with the issue: 6.9.0
- Last working add-on release (if known): 
- Operating environment (OS/Supervised): HassOS 5.13, Core Version core-2021.5.5, Supervisor Version supervisor-2021.05.4

## Problem-relevant configuration
N/A

Issue #1521, opened 2020-07-30
Phoscon Web App stays empty


Br3b:
hi there,
i just moved my deconz from a pi 2b raspian to my homeassistant pi 4b. 
i managed to get everything working including my phoscon backup i did before.

today i updated  to the new HA version and its not working anymore.
i can access Deconz and the devices are there but if i open Phoscon Web the site keeps empty.

i already tried to reinstall the addon serveral times and restart the pi/HA.. 

like this i cant control any lights :/
Please help!

![1](https://user-images.githubusercontent.com/49213470/88907702-3b7a1400-d259-11ea-8b08-a234c019a9a3.PNG)
![2](https://user-images.githubusercontent.com/49213470/88907833-63697780-d259-11ea-97cf-5f76a55393ea.PNG)

Btullis11:
If you are still having issues, here is what I did
Click on the phoscon "P" icon
It should give you a screen that is partially functional.
Click on the 3 horizontal bars icon in the upper left hand corner and select logout
Go back to any other menu in home assistant such as overview
Click on the deconz app. 
Mine then prompted me to enter a new password twice 
Once it logged in everything was functional, however the entity names had changed so I had to rename everything to get Hass entities working again.
It seems to have been changes to the conbee firmware that necessitated the reset.
I hope this helps.

Sirezul:
I'm facing the same issue, I think it's due a lack of **https support**  on the Deconz/Phoscon side.
Try to access it over http, it'll work.

Mimiix:
@Sirezul It is party that, but it also is partly to the fact you can't access it over the external IP if your port isn't open.

Br3b:
thank you very much for your answers.
i just did everything new and changed to the conbee stick.

PR #2325, opened 2021-12-26
Fix deCONZ/Phoscon Ingress access when using FQDN

### LINKED ISSUE #947: Deconz through ingress has an unselectable system
I have updated to Deconz 5.0 and HA 103.4 HassOS 3.7 and when I use the Deconz addin through ingress I get a page where I can see my Conbee stick but not select it (it is grayed out):
image

Everything else that was configured before upgrading to version 4 of Deconz still works, I just cannot use the deconz web interface to add new things or change config.

### LINKED ISSUE #1677: After upgrade to deCONZ 6.5 the phoscon web app does not work any more
I click on deconz / phoscon. Normally it would launch the login dialog and after login the web app.
Now there is just the phoscon logo at the top. When I click onto it, a loading screen starts: "Bitte warten, während die Suche nach Gateways durchgeführt wird." --> Wait while looking for gateways. That's it. No more.

The other deconz-VNC-App seems to work.

### LINKED ISSUE #2172: deCONZ: Phoscon Web App Login only visible when using IP-Addr as an URL via Ingress
## The problem

My vanilla hass.io instance ran smothly from 192.168.2.10:8123 (your address my vary). 
To make it it nicer, I set up a dns entry that would link ha.mydomain.tld to the address and then use http://ha.mydomain.tld:8123 just because.

Now when I use the domain url (does NOT matter if http or https), the login field of deCONZ does NOT show up / you can't choose the stick.

Since I would enforce SSL and the domain in the next step, deCONZ can't be configured any more properly.

## Environment
- Raspberry PIE 4 / 8GB / 64 Bit Image
- Home Assistant 2021.8.8 , but Problem exists at least since 2021.6 or 5 (not tried that before)
- ConBeeUSB II Stick
- Fresh Install is the same

There is no config that causes that except when I use a domain name instead of an IP Address

## Traceback/Error logs
I can't find log errors on the server. It might fail in the browser but looking at the Web-Console reveals nothing to my knowledge (so no error in the JS-Console for Example)

## Additional information
Don't know anything else.
I've included a screenshot side by side of the Browser display with IP-Address and with Domain-Name. 
Screenshot - no Login Side by Side

tannyl:
This fixes the problem with not being able to use the Phoscon App interface via Ingress when using a fully qualified domain name.

The Phoscon Apps javascript code is, in lack of better words, a bit wonky and contains errors, but with this fix, Ingress should at least work when using both IP and domain name to access Home Assistant.

The fix helps the App find the correct gateway and also stops it from trying to find other gateways.

It also improves the way "subnet probing" is stopped.

I have tested this by applying the fixes directly to the nginx.conf file in the docker container and everything seems to work as it should.

Resolves #2172
Resolves  #947
Resolves #1677

Most likely related to  #1521 and #2054


pvizeli:
Change log + bump version with 0.0.1

thanks

tannyl:
@pvizeli updated version and changelog. :smiley:

tannyl:
@pvizeli anything else that needs to be done?

tannyl:
@klaasnicolaas @frenck
Maybe one or both of you would take a look at this? Thanks.

frenck:
@tannyl Please don't ping people at random for reviews. That is not something we want in this org, to prevent people from being flooded with pings. Additionally, I cannot review it now, as I don't want such ping behavior to work/be rewarded with a review. Hope you'll understand (nothing personal). There are 800+  PRs open across the org, this isn't the oldest either.

tannyl:
Sorry about that @frenck 

It was not "at random" though. I saw that you and klassnicolaas where some of the last ones to add changes to this addon, thus I expected you might be the ones "in charge of it". I will remember not to ping like this again and I have not taken anything personal.

mdegat01 (approved):
Wow, great work. I can merge just need you to rebase and resolve the conflict with `config.yaml`. Believe you just need to add `---` as a formatter did that in a later PR.

tannyl:
@mdegat01 sounds great. I have fixed `config.yaml` and it should be ready to go no.

mdegat01:
Thanks @tannyl 👍

PR #2329, opened 2021-12-29
Update to the latest Almond/Genie release

gcampax:
This PR updates the Almond add-on to the latest version, and rebrands it as Genie.

I'm sending this PR early, before Genie itself is fully released, to get early feedback on the new dockerfile, because there are some significant changes due to how the client is installed. I will bump the version numbers in build.yaml to the stable release before merging.

I tested the add-on using our third-party repository and the same dockerfile.


gcampax:
Hello, is there any update on this?

mdegat01 (approved):
FYI you should be able to rename the folder `almond/rootfs/etc/services.d/almond` to `almond/rootfs/etc/services.d/genie` without issue. S6 just runs everything in the `/etc/services.d` folder. The only thing that we can't change from almond to genie is the almond folder in the root of this repo and the slug, can only do that by making a new addon called genie and deprecating this one.

@gcampax can you do a rebase though? Then I can merge it, looks good otherwise.

almond/build.yaml
```diff
@@ -2,8 +2,6 @@ build_from:
   aarch64: ghcr.io/home-assistant/aarch64-base-debian:buster
   amd64: ghcr.io/home-assistant/amd64-base-debian:buster
   armv7: ghcr.io/home-assistant/armv7-base-debian:buster
-codenotary:
-  signer: notary@home-assistant.io
-  base_image: notary@home-assistant.io
```
mdegat01, line 7: Actually wait, I thought this was a rebase issue but it looks like this section was specifically removed from the `build.yaml` file. Is there a reason this information was removed?

PR #2330, opened 2021-12-30
Hint to not use @ in password

gartenbauer:
A hint to not use @ in the mysql password, otherwise the connection does not work

### DIFF
--- a/mariadb/DOCS.md
+++ b/mariadb/DOCS.md
@@ -10,7 +10,7 @@ Follow these steps to get the add-on installed on your system:
 
 ## How to use
 
-1. Set the `logins` -> `password` field to something strong and unique.
+1. Set the `logins` -> `password` field to something strong and unique without `@` in it.
 2. Start the add-on.
 3. Check the add-on log output to see the result.
 4. Add `recorder` component to your Home Assistant configuration.


homeassistant:
Hi @gartenbauer, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2330).

Thanks!

frenck (commented):
Why wouldn't that work?

gartenbauer:
The format to add recoder integration [db_url](https://www.home-assistant.io/integrations/recorder/#db_url)  for [mysql](https://www.home-assistant.io/integrations/recorder/#mysql) of is the fallowing: 
`mysql://user:password@SERVER_IP/DB_NAME?charset=utf8mb4`

When I chose a password with an @, the the logs of the homeassistant core gave me an error message.
I suspect the recorder is now trying to connect to the part in front of the @ with Mariadb. Which does not work, because on the one hand the password is wrong and on the other hand the IP makes no sense
**Example:**
password set in the mariadb addon: _CDhF@fp4%@j25&&#_
the recoder would try to connect to the server_ip: _j25&&#_

frenck:
That is not because you can't, it's because you are using it incorrectly.

The database connection string is an URI; just like an URL,  which cannot contain `@` plain,  but they need to URL encoded. So, an `@` becomes: `%40`.

Closing this PR; as it is not correct.

PR #2331, opened 2022-01-01
letsencrypt: add IONOS DNS provider support

morphis:
IONOS provides API support to change DNS configuration which is
supported by certbot. See https://developer.hosting.ionos.de/docs/dns
and https://github.com/helgeerbe/certbot-dns-ionos


jandelgado:
Since IONOS is a major provider in germany, it would be great to see this PR accepted

pokemon81:
When comes Release 4.13.0 ?

LukasMayr:
push!

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -96,6 +96,13 @@ elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-njalla" ]; then
     bashio::config.require 'dns.njalla_token'
     PROVIDER_ARGUMENTS+=("--authenticator" "certbot-dns-njalla:dns-njalla" "--certbot-dns-njalla:dns-njalla-credentials" /data/dnsapikey "--certbot-dns-njalla:dns-njalla-propagation-seconds" "${PROPAGATION_SECONDS}")
 
+# IONOS
+elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-ionos" ]; then
+    bashio::config.require 'dns.ionos_endpoint'
+    bashio::config.require 'dns.ionos_prefix'
+    bashio::config.require 'dns.ionos_secret'
+    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /data/dnsapikey)
+
```
mdegat01, line 105:
```suggestion
```
Don't do schema validation here. If you want to do validation like this you have to do it in the `cont-init.d` file. The problem with doing validation in the run script is that if this script fails then s6 will try to restart it causing the same validation errors over and over again.

mdegat01 (changes requested):
Looks good! Small change and please let me know about my question on `ionos_endpoint`

opacey:
Hi, Did IONOS ever get merged? I don't see them in the list on the addon, despite this issue being closed and there being a recent series of minor updates released. Thanks.

agners:
This is not an issue, this is a pull request. The author would need to address the review comments. With the latest updates to newer Alpine version it would also need a rebase.

opacey:
Oops, so it is! @morphis would you consider refreshing your work on the IONOS module (and the review comments)?

exp1338:
Is there no support any more? Ionos is not listed inside the supported DNS provides. :(

PR #2353, opened 2022-01-20
Enable Samba on all supported interfaces

### LINKED ISSUE #2074: Samba share 9.5.1 only work in eth0
Hi

With actual version do not posible to access via SAMBA over wlan0's IP

In configutarion, 
interface: ''

Also probe it:
interface: wlan0, eth0

but same case,

Any idea?

With previous version also eth0 and wlan0's IP was accesible via SAMBA  automatically

lmagyar:
fixes #2074
problems introduced in #2031

Samba v9.5.x removed the possibility to configure `interfaces` parameter of samba and defaulted to the primary interface of HA (previously samba was running on all interfaces by default).

This breaking change broke a lot of non-standard configuration with vpn, ssh, wlan, etc.

This PR enables Samba on all supported interfaces (note: *all supported interfaces* is a shorter list than *all interfaces*).

This PR is already tested on some real machines.

~This PR restores the `interfaces` configuration:~
~- it is backward compatible with 9.5.1, ie. it's optional, if it is not configured, behaves exactly the same as 9.5.1, samba runs on the primary interface of HA~
~- if you specify `interfaces`, it works as expected, runs on the specified interface, or interfaces~
~- it is renamed to `interfaces` (note the plural), a string-list, instead of a string, because this is de facto a list in samba config, like `allow_hosts`~

~Sample config, that can be added to the original configuration:~


frenck:
We should not re-introduce this, as this should not be a problem on supported systems. The only reason I see to re-introduce this is to support setups we actually do not support?

I guess the main reason to do this, is because you want multiple interfaces (that are supported) to work? If so, I suggest we adjust the code to iterate over all available interfaces configured @ the Supervisor and use that instead.

For example, by getting all known interfaces.

```bash
declare -a interfaces

interfaces+=$(bashio::network.interfaces))
for interface in "${interfaces[@]}"; do
  # Do something
done
```

The above code is obviously not a cut & paste solution but is merely to illustrate a different approach.

frenck (changes requested):
Hit the wrong button, meant to hit the request changes button. Second try :)

lmagyar:
> I guess the main reason to do this, is because you want multiple interfaces (that are supported) to work?

If using multiple interfaces are supported, i think problem solved. Currently Samba runs only on the primary interface, causing the complaints.

I will use & test & PR your change request soon.

**Question**: I'm not familiar how supervisor filters down all the `ip link show` "long-list" interfaces, temporarily unavailable interfaces, eg. mobile stick + vpn will be in the supervisor's "short-list"?

lmagyar:
Now Samba runs on all supported interfaces, no user config allowed/needed.

Tested with only eth0, and with eth0 + unconfigured wlan0. Logs look good.

**Question:** some other addons can also use loopback, like ssh or mariadb, and they can communicate through it, entering only "lo" and "localhost" values in config files, no direct IP values scattered around. What do you think about adding `lo` to the interface list of Samba?

frenck:
> What do you think about adding lo to the interface list of Samba?

I don't see the use case for that? As in, you would need an actual samba-client that would connect to the loopback interface. When would that happen?

lmagyar:
**On loopback corner case:** I use ssh to forward my local 139 and 445 ports (and 3306 mysql, etc), and ssh connects to samba through loopback, so samba is available only through ssh tunnel. OK, maybe I'm paranoiac to access everything through ssh. Port forwarding works through eth0 address also, but in that case I had to use my HA eth0 address in ssh client config (L x.x.x.x:139 **192.168.x.x**:139, with samba on loopback it is only L x.x.x.x:139 **loopback**:139). Note: I'm a newbie in this ecosystem, I'm not surprised if you have a better idea.

frenck:
Right, well since the add-on runs on the host network, I think that isn't an issue and (arguably) a valid case. I don't see a harm in binding it to `lo`.

lmagyar:
OK, I will modify the PR this afternoon/evening.

lmagyar:
I've added loopback to the interfaces list. Tested on my machines.

frenck:
@lmagyar Can you make sure CI passes?

lmagyar:
@frenck CI passes now. Sorry for my mess.

daoudeddy:
Tested on both wlan0 and eno1 working fine without any problems

lmagyar:
Resolved conflicts in config.yaml

mdegat01 (approved):

mdegat01:
Thanks @lmagyar 👍

markus99:
Appears it may have caused some unintended issue(s) @lmagyar / @mdegat01 - Issue created here: https://github.com/home-assistant/addons/issues/2447 and forum post here: https://community.home-assistant.io/t/samba-9-6-upgrade-killed-samba/416362

PR #2364, opened 2022-01-27
Improve GitHub action

agners:
- Add YAML linter
- Drop GitHub registry login


.github/move.yml
```diff
@@ -1,6 +1,7 @@
 # Configuration for move-issues - https://github.com/dessant/move-issues
```
agners, line 1: I don't think this is active is it? So I can drop the whole file.

ludeeus: Drop it, we use native github move or ask the user to recreate issues in the correct repo.

.github/dependabot.yml
```diff
@@ -1,8 +1,9 @@
...
-  directory: "/"
-  schedule:
-    interval: daily
-    time: "06:00"
-  open-pull-requests-limit: 10
\ No newline at end of file
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
```
ludeeus, line 7:
```suggestion
      interval: weekly
```
Let's do weekly for these

deconz/config.yaml
```diff
@@ -1,7 +1,9 @@
+---
 version: 6.12.0
 slug: deconz
 name: deCONZ
-description: Control a Zigbee network with ConBee or RaspBee by Dresden Elektronik
+description: >
```
ludeeus, line 5:
```suggestion
description: >-
```
Use `>-` to strip whitespace for these (same goes for all others like it)

letsencrypt/config.yaml
```diff
@@ -66,7 +67,11 @@ schema:
     ovh_consumer_key: str?
     ovh_endpoint: str?
     propagation_seconds: int(60,3600)?
-    provider: list(dns-cloudflare|dns-cloudxns|dns-digitalocean|dns-directadmin|dns-dnsimple|dns-dnsmadeeasy|dns-gehirn|dns-google|dns-linode|dns-luadns|dns-njalla|dns-nsone|dns-ovh|dns-rfc2136|dns-route53|dns-sakuracloud|dns-netcup|dns-gandi|dns-transip)?
+    provider: "list(dns-cloudflare|dns-cloudxns|dns-digitalocean|\
```
ludeeus, line 70: Can you use `>-` with multiline here to make it more readable?

agners: I don't think this works, `>-` only strips newlines at the end (according to https://yaml-multiline.info/)

ludeeus: `|-` should work

ludeeus (approved):

PR #2372, opened 2022-02-01
homematic: use cp instead of mv for crRFD files to avoid mv subdir error

sdirkwinkel:
mv "${TMPDIR}/usr/local/etc/config/crRFD/*" does not expand the glob and fails.
Due to the previous change we also ignored the globbing error.
Let's clean the directory up, copy from the source and avoid globbing errors.
By doing it this way we can also drop the ugly "|| true".

related: https://github.com/jens-maus/RaspberryMatic/issues/1681
cc @jens-maus

### DIFF
--- a/homematic/rootfs/opt/hm/bin/createBackup.sh
+++ b/homematic/rootfs/opt/hm/bin/createBackup.sh
@@ -68,8 +68,9 @@ if [[ -d "${TMPDIR}" ]]; then
   cp -a /data/* "${TMPDIR}/usr/local/etc/config/"
 
   # move crRFD files to data sub-dir to be compatible
-  mkdir -p "${TMPDIR}/usr/local/etc/config/crRFD/data"
-  mv "${TMPDIR}/usr/local/etc/config/crRFD/*" "${TMPDIR}/usr/local/etc/config/crRFD/data/" 2>/dev/null || true
+  rm -rf "${TMPDIR}/usr/local/etc/config/crRFD"
+  mkdir -p "${TMPDIR}/usr/local/etc/config/crRFD"
+  cp -a /data/crRFD "${TMPDIR}/usr/local/etc/config/crRFD/data"
 
   # cleanup
   rm -f "${TMPDIR}/usr/local/etc/config/options.json"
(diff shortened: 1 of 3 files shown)


sdirkwinkel:
Sorry for all the noise. I've tried this by logging in with docker exec and the backups are now usable

homematic/rootfs/opt/hm/bin/createBackup.sh
```diff
@@ -68,8 +68,9 @@ if [[ -d "${TMPDIR}" ]]; then
   cp -a /data/* "${TMPDIR}/usr/local/etc/config/"
 
   # move crRFD files to data sub-dir to be compatible
+  rm -rf "${TMPDIR}/usr/local/etc/config/crRFD"
   mkdir -p "${TMPDIR}/usr/local/etc/config/crRFD/data"
-  mv "${TMPDIR}/usr/local/etc/config/crRFD/*" "${TMPDIR}/usr/local/etc/config/crRFD/data/" 2>/dev/null || true
+  cp -r /data/crRFD/. "${TMPDIR}/usr/local/etc/config/crRFD/data/"
```
jens-maus, line 73: Why not simply use the following instead:

```
cp -a /data/crRFD "${TMPDIR}/usr/local/etc/config/crRFD/data"
```

This should work aswell and looks cleaner to me.

jens-maus: And then you have to alter the "mkdir -p" call one line above to

```
mkdir -p "${TMPDIR}/usr/local/etc/config/crRFD"
```

sdirkwinkel: ok, I've updated my branch

jens-maus: Ok, thanks. Looks better now. So please retest with your environment and then let use know if the backup+restore procedure still works as expected.

sdirkwinkel: ok, I've now tested:
- update backup script using docker exec
- create backup (no error in ui)
- import backup into new install of raspberrymatic
=> i can see and control homematic ip devices

jens-maus: perfect. and what about old/traditional BidCos/homematic devices? do you see them?

sdirkwinkel: I don't have any of those, sorry

jens-maus: ok.

jens-maus:
@pvizeli Sorry to bother once again with this old homematic addon. But hopefully @sdirkwinkel fixed the last bit of the partly broken backup procedure. So please merge 99.0.5 so that homematic add-on users can migrate more easily.

pvizeli (approved):

PR #2409, opened 2022-03-16
Clarify DuckDNS alias instructions

roziscoding:
Since we place DuckDNS names on the `domains` key, users who don't pay attention to the domain names used in the example (like me) are lead to think the should put the DuckDNS domain in the `domain` key of the alias object.
This PR adds an observation noting that the `domain` key should receive the custom domain name, and not the DuckDNS domain

### DIFF
--- a/duckdns/DOCS.md
+++ b/duckdns/DOCS.md
@@ -115,6 +115,8 @@ aliases:
     alias: my-domain.duckdns.org
 ```
 
+> Note that the `domain` key should receive **your custom domain name** and the `alias` key should rececive the DuckDNS domain name
+
 Don't add your custom domain name to the `domains` array. For certificate creation, all unique domains and aliases are used.
 
 Also, don't forget to make sure the dns-01 challenge can reach Duckdns. It might be required to add a specific CNAME for that:


homeassistant:
Hi @roziscoding, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2409).

Thanks!

duckdns/DOCS.md
```diff
@@ -115,6 +115,8 @@ aliases:
     alias: my-domain.duckdns.org
 ```
 
+> Note that the `domain` key should receive **your custom domain name** and the `alias` key should rececive the DuckDNS domain name
```
mdegat01, line 118:
```suggestion
Notice that here `domain` is set to a a custom domain name, not a duckdns one. The duckdns domain this custom domain is an alias for goes in `alias`.
```
This isn't quoting anything so it shouldn't use that formatting. There are other examples of `Note: ` formatting in this doc but I actually don't think this needs it. I think we can just rephrase it as part of the example's explanation like so.

mdegat01 (changes requested):

PR #2410, opened 2022-03-18
 Support CNAME records

ocalvo:
This change adds support for CNAME records in the DNSMasq addon.


ocalvo:
@pvizeli Can you review this PR again? The old one got closed while I was on vacation.

ocalvo:
Keep alive

ocalvo:
Keep alive

mdegat01 (approved):
@ocalvo Looks good. One piece of feedback, DNSMasq documentation includes [this note](https://github.com/imp/dnsmasq/blob/9560658c5ba27a370b97d30850db474ce67912b7/dnsmasq.conf.example#L664-L667):
> # Provide an alias for a "local" DNS name. Note that this _only_ works
> # for targets which are names from DHCP or /etc/hosts. Give host
> # "bert" another name, bertrand
> #cname=bertand,bert

I must admit I don't have a lot of familiarity with DNSMasq as I use Adguard myself. But the emphasis on "only" there suggests to me it is something which has caused DNSMasq users confusion. Therefore I would suggest we include the same information as a note in the doc on the `cnames` option. What do you think?

ocalvo:
Keep alive

PR #2415, opened 2022-03-21
ada: Update to Debian bullseye

agners:


ada/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update \
     && rm -rf .git \
     && pip3 install --no-cache-dir -U setuptools wheel \
     && pip3 install --no-cache-dir -r requirements.txt \
-    && pip3 install --no-cache-dir -e . \
+    && pip3 install --no-cache-dir . \
```
pvizeli, line 26: If we do this, we do not need compileall and also can clean up the folder

pvizeli (approved):

Issue #2222, opened 2021-10-04
Mosquitto Broker Add-On Documentation Issue


rpfinneran:
## The problem

The documentation for the Mosquitto Add-on (docs.md) has a section on leveraging an ACL to restrict user access to topics. If a user implements the ACL file, they must include the 'homeassistant' user with read/write priv to all topics, otherwise HA won't be able to discover new entities or even determine entity availability for manually added entities in the configuration.yaml.

## Environment

- Add-on with the issue: Mosquitto broker
- Add-on release with the issue: 6.0.1
- Last working add-on release (if known): N/A, documentation issue
- Operating environment (OS/Supervised): OS 6.3

## Problem-relevant configuration

Local documentation page is: /hassio/addon/core_mosquitto/documentation

Bottom code block should be changed from:

```
user [YOUR_MQTT_USER]
topic readwrite #
```
To:

```
user homeassistant
topic readwrite #

user [YOUR_MQTT_USER]
topic readwrite #
```

## Traceback/Error logs

``` N/A

```

PR #2421, opened 2022-03-28
Updated mosquitto ACL docs with note about homeassistant user

jabelone:
I ran into the same issue as reported in #2222, but looks like it never got fixed.

The summary is that once a user enables the MQTT ACL, home assistant integrations that rely on MQTT will stop functioning as they won't have permissions to publish/subscribe to their topics. I've added a note and example to the documentation to stop this from happening as it's a very confusing state to get yourself in if you don't have a good understanding of how MQTT ACLs work.

### DIFF
--- a/mosquitto/DOCS.md
+++ b/mosquitto/DOCS.md
@@ -97,7 +97,7 @@ See the following links for more information:
 - [Mosquitto topic restrictions](http://www.steves-internet-guide.com/topic-restriction-mosquitto-configuration/)
 - [Mosquitto.conf man page](https://mosquitto.org/man/mosquitto-conf-5.html)
 
-Add the following configuration to enable **unrestricted** access to all topics.
+Add the following configuration to enable **unrestricted** access to all topics for `[YOUR_MQTT_USER]`. Once you enable MQTT ACLs, home assistant will no longer have permission to publish or subscribe to the topics that it needs to. It is suggested that you add the `homeassistant` and `addons` user to the ACLs as demonstrated below.
 
 1. Enable the customize flag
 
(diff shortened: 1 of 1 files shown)


mosquitto/DOCS.md
```diff
@@ -97,7 +97,7 @@ See the following links for more information:
 - [Mosquitto topic restrictions](http://www.steves-internet-guide.com/topic-restriction-mosquitto-configuration/)
 - [Mosquitto.conf man page](https://mosquitto.org/man/mosquitto-conf-5.html)
 
-Add the following configuration to enable **unrestricted** access to all topics.
+Add the following configuration to enable **unrestricted** access to all topics for `[YOUR_MQTT_USER]`. Once you enable MQTT ACLs, home assistant will no longer have permission to publish or subscribe to the topics that it needs to. It is suggested that you add the `homeassistant` and `addons` user to the ACLs as demonstrated below.
```
mdegat01, line 100:
```suggestion
Add the following configuration to enable **unrestricted** access to all topics for `[YOUR_MQTT_USER]`.

**Note:** Home Assistant expects the users `homeassistant` and `addons` to have unrestricted readwrite access to all topics. If you choose to enable ACLs, you should grant these users these permissions as demonstrated below. Otherwise you will run into issues.
```

mdegat01 (approved):
Good catch, small suggested tweak to the language but I agree this should be noted.

jabelone:
Perfect, updated! I don't have a super good understanding of how the MQTT integrations work, so whatever phrase you think is best. :)

mdegat01:
Ah shoot. Missing a version bump and changelog. Gotta get better at remembering that 🤦 will fix

PR #2449, opened 2022-04-28
Use HA wheels during build

mdegat01:
Use HA wheels when possible at build time to improve performance.


letsencrypt/Dockerfile
```diff
@@ -22,7 +23,9 @@ RUN \
         musl-dev \
         openssl-dev \
         cargo \
-    && pip3 install --no-cache-dir \
+    && pip3 install --no-cache-dir --find-links \
+        "https://wheels.home-assistant.io/alpine-$(cut -d '.' -f 1-2 < /etc/alpine-release)/${BUILD_ARCH}/" \ 
+        cryptography==36.0.2 \
```
ludeeus, line 28: Add the version to https://github.com/home-assistant/addons/blob/master/letsencrypt/build.yaml#L17 and reference it like the others

PR #2453, opened 2022-04-28
Update mosquitto to 2.0.11

mdegat01:
Update mosquitto from `1.6.12` to `2.0.11`. In support of this I also made the following changes:
- Change from [mosquitto-auth-plug](https://github.com/pvizeli/mosquitto-auth-plug) to [mosquitto-go-auth](https://github.com/iegomez/mosquitto-go-auth) as that latter appears to be actively maintained, working with Mosquitto v2.0.0+ and recommended by Eclipse (see [here](https://mosquitto.org/documentation/authentication-methods) under Authentication Plugins)
- Base image changed from alpine3.13 to debian:bullseye as mosquitto-go-auth uses a c-shared build and those do not work on musl

I pulled this branch locally onto my arm64 dev system and confirmed it works. Was able to authenticate with the broker using both local and HA users successfully.

### DIFF
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -6,6 +6,10 @@
 - Remove `mosquitto-auth-plug` and replace with `mosquitto-go-auth` 1.8.2
 - Change base image from Alpine to Debian (required for `mosquitto-go-auth`)
 
+Note: Mosquitto 2.0.0 did contain some breaking changes. We don't anticipate
+most users to be affected by them but if you have a heavily customized mosquitto
+config we would advise reviewing [their changelog](https://mosquitto.org/ChangeLog.txt). 
+
 ## 6.0.2
 
 - Mention homeassistant and addon users in ACL doc
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Update base image to Alpine Linux 3.14
 - Update mosquitto to 2.0.11
 - Update nginx to 1.20.2
+- Remove mosquitto-auth-plug and replace with mosquitto-go-auth 1.8.2
 
 ## 6.0.2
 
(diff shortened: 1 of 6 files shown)


mosquitto/Dockerfile
```diff
@@ -4,36 +4,48 @@ FROM $BUILD_FROM
...
     \
-    && apk del --no-cache .build-dependencies \
+    && apt-get purge -y --auto-remove \
+        build-essential \
+        git \
+        mosquitto-dev \
+        libmosquitto-dev \
+        openssl \
+        libssl-dev \
+        golang-go \
```
pvizeli, line 38: We have to cleanup the go compiler cache as well on the /root home

mdegat01: So just `rm -rf /root/go` basically right? makes sense

pvizeli: I think there is another dotCache folder on dotConfig or plain at home

mdegat01: Found it, `/root/.cache`. Cleaned both.

mdegat01:
@pvizeli I read the mosquitto changelog and based on the particular breaking changes in 2.0.0 I don't personally feel they warrant a major version bump for us. I don't think most users will notice them. I did add a note though and a link to the changelog to give users with advanced and heavily customized config a heads up. Let me know what you think.

mosquitto/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+## 6.1.0
```
frenck, line 3: This should have been an major bump. This PR contains major chsnges.

mosquitto/Dockerfile
```diff
@@ -4,36 +4,50 @@ FROM $BUILD_FROM
...
         /etc/nginx/* \
         /usr/share/nginx \
-        /usr/src/mosquitto-auth-plug \
+        /usr/src/mosquitto-go-auth \
         /var/lib/nginx/html \
-        /var/www
+        /var/www \
+        /var/lib/apt/lists/* \
+	/root/.cache \
+	/root/go
```
frenck, line 50: Indentation got lost here?

Issue #2474, opened 2022-05-04
Mosquitto 6.1.x with mosquitto-go-auth spams "error: received null username, ..."


igogold:
### Describe the issue you are experiencing

After upgrade Mosquitto add-on to 6.1.1 I've got a spam in add-on's log with a line:
```text
error: received null username, clientid or topic, or access is equal or less than 0 for acl check
error: received null username, clientid or topic, or access is equal or less than 0 for acl check
error: received null username, clientid or topic, or access is equal or less than 0 for acl check
error: received null username, clientid or topic, or access is equal or less than 0 for acl check
```
I enabled debug output for mosquitto and found after a message from one of external devices I always got this error. Second devices with the same bridge configuration works well. There are two lines in the debug mode:
```text
1651663393: Received PUBLISH from wirenboard-A4JBKLBD.wb2hass (d0, q1, r1, m26535, 'wb1/devices/power_status/controls/Vin', ... (5 bytes))
error: received null username, clientid or topic, or access is equal or less than 0 for acl check
1651663393: Sending PUBACK to wirenboard-A4JBKLBD.wb2hass (m26535, rc0)
1651663393: Received PUBLISH from wirenboard-AHL3GFZO.wb2hass (d0, q1, r1, m54105, 'wb2/devices/wb-msw-v3_115/controls/Current Motion', ... (2 bytes))
1651663393: Sending PUBACK to wirenboard-AHL3GFZO.wb2hass (m54105, rc0)
```
It's interesting I anyway got the data to 'wb1/#' topic, but also have a huge spam as I have continuous flow of data especially from wirenboard-A4JBKLBD.wb2hass ('wb1/#').

I found this error text in the mosquitto-go-auth plugin sources but see no reason to get this message as have no difference between device with 'wb1/#' and one with 'wb2/#' messages.

I've tried to add usernames for mqtt clients in the add-on config, before I used HA accounts. mosquitto spams the error anyway.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Mosquitto broker

### What is the version of the add-on?

6.1.1

### Steps to reproduce the issue

1. Upgrade add-on to 6.1.1
2. Wait client connected.
3. Read spam in the log.

mdegat01:
Do you use the tasmota integration by any chance?

igogold:
[...]

PR #2477, opened 2022-05-04
Add debug option to mosquitto

mdegat01:
When true, turns on debug logging for mosquitto and the auth plugin. May help us get to the bottom of issues like #2474

### DIFF
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.1.2
+
+- Add debug option to help with issues
+
 ## 6.1.1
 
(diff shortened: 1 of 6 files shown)


frenck (approved):

frenck:
This looks good, however, I think we should consider handling this debugging (logging) more globally in general.

pvizeli:
Yeah, we have also logging information on the `/info` call for such proposal

mdegat01:
I agree with the idea of centralizing but I'm not sure if I actually like the idea of tying all debug logging to supervisor's log level if that's what you mean @pvizeli . Often you want to debug one thing and turning on debug logging everywhere creates a lot of unnecessary noise.

Issue #1715, opened 2020-12-15
IPv6 Config URL causing duckdns script to exit with status 7


mattclar:
## The problem
DuckDNS addon is failing and exiting the script when a URL value is added for ipv6 only.
I have tested with just adding a static IPv6 address as the config value and this works fine
Also only adding a URL for the ipv4 value works
It seems to be only when I use a URL value for the ipv6 config that it fails?!

## Environment

- Add-on with the issue: DuckDNS
- Add-on release with the issue: 1.12.4
- Last working add-on release (if known): NA
- Operating environment (OS/Supervised): Hassos 5.8

## Problem-relevant configuration

```lets_encrypt:
  accept_terms: true
  certfile: fullchain.pem
  keyfile: privkey.pem
token: 6697a35c-7b8f-48de-8c4e-##########
domains:
  - arthurst.duckdns.org
aliases: []
seconds: 300
ipv6: 'https://api6.ipify.org/'

```

## Traceback/Error logs

```[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
# INFO: Using main config file /data/workdir/config
+ Account already registered!
[cmd] /run.sh exited 7
[cont-finish.d] executing container finish scripts...
[cont-finish.d] done.
[s6-finish] waiting for services.
[s6-finish] sending all processes the TERM signal.

```

make-all:
I hadn't figured out the reason, but I have been seeing this error from duckdns since the last update at the beginning of November.
My configuration is the same: 
    ipv6: 'https://api6.ipify.org/'
and nothing for ipv4.

My ISP allocates IPv6 prefixes dynamically, and the IPv4 addresses are CG-NAT, so no route back in from the internet.  The DuckDNS plugin has never really worked correctly for this setup, but at least it used to keep running for the letsencrypt cert renewal.

nicogramm:
Same problem here.
I use the Nginx Proxy Manager for my certificate so I have Let's Encrypt disabled in this Add-on (`lets_encrypt.accept_terms`=`false`).
But still, when I set the config option for IPv6 the Add-on exists with status 7.

## Add-on log output
```text
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[...]

PR #2490, opened 2022-05-20
duckdns: improve IPv6 support

sbyx:
As I found the IPv6 support for duckdns a bit unusable, here are some proposed changes:

* Use Hassio API to detect IPv6 addresses of the host, selectable by interface
* Separate IPv4 and IPv6 updates as DuckDNS skips IPv4 autodetection with an IPv6 address present
* Fix misleading help text for IPv6 address autodetection (DuckDNS states it is not supported)

You can now use the IPv6 field also to either input "default" or the name of an interface from the hassio network info to use that interface's IPv6 address in the DNS update. There is also a basic check to skip link-locals or ULA addresses.

Now when an IPv6 address is present in the request the IPv4 autodetection was disabled (btw. unless stated in the parameter documentation there is no IPV6 autodetection). Therefore I have separated IPv4 and IPv6 updates so both work in parallel. You can manually disable IPv4 by inputting something like "none" in the IPv4 address parameter, i.e. something that is not an IP-address and not a URL (in principle anything that doesn't contain a "." or ":/").

As for the second commit, this is mainly housekeeping, since I was having some s6 related trouble when building / testing locally as mentioned in the blog here: https://developers.home-assistant.io/blog/2022/05/12/s6-overlay-base-images/ Please skip or verify it is sound.


sbyx:
Fixes #1715

sbyx:
> Please update the changelog

Done, thanks for the review.

pvizeli:
Can you please rebase it for making CI working on the PR?

sbyx:
Rebased to current HEAD

PR #2499, opened 2022-05-25
Add executable flag to all found scripts

raman325:
This PR was created by running the following four commands:
- `find . -type f -name "*.sh" -exec chmod +x {} \;`
- `find . -type f -name "discovery" -exec chmod +x {} \;`
- `find . -type f -name "finish" -exec chmod +x {} \;`
- `find . -type f -name "run" -exec chmod +x {} \;`

It may not cover every non executable script, these were just the obvious search patterns


mdegat01:
I see and the idea was to intentionally not update the changelog and version. That way we don't build a new image right now and still get another chance to ensure we got them all next time each of these addons is updated. That makes sense, man there are a lot of these. 

I would guess we also have to add `init: false` to a lot of addons as detailed [here](https://developers.home-assistant.io/blog/2022/05/12/s6-overlay-base-images/#minimum) but that can be a separate PR.

mdegat01:
@raman325 just to clarify my comment above, as is this won't build any new images for any addons. Is there an addon broken right now that you wanted to fix with this PR? If so then we should comb through that addon in particular to ensure none are missed and bump its version/changelog. If not then I'm fine with merging this as there's no harm and it gets us one step closer to s6 v3 readiness.

raman325:
> @raman325 just to clarify my comment above, as is this won't build any new images for any addons. Is there an addon broken right now that you wanted to fix with this PR? If so then we should comb through that addon in particular to ensure none are missed and bump its version/changelog. If not then I'm fine with merging this as there's no harm and it gets us one step closer to s6 v3 readiness.

Not that I am aware of. A release of `zwave_js` was broken yesterday because this fix wasn't applied. This is really a preventative step to help the next person.

ludeeus (approved):

PR #2518, opened 2022-06-08
Bump config check base to `2022.05.0`

### LINKED ISSUE #2512: Upgrade from 2022.5.5 to 2022.6: check_config add-on reports error with python

I'm following up on https://github.com/home-assistant/core/issues/72868, where allenporter asked that I create a separate issue.

When running the Check Home Assistant configuration add-on for the latest version (2022.6), the following error occured and displayed in the log:

`Testing configuration at /tmp/config Failed config General Errors: - Component error: nest - Error relocating /usr/local/lib/python3.9/site-packages/grpc/_cython/cygrpc.cpython-39-arm-linux-gnueabihf.so: _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE7reserveEv: symbol not found`

Please note that after reviewing documentation and checking my backups, I decided to upgrade to 2022.6, and that worked without any issues.

Home Assistant OS

Check Home Assistant configuration

3.10.0

1. Start the Check Home Assistant configuration add-on
2. Wait for it to install a new version
3. Check the log for details. It fails with the error I reported above.

```txt
N/A
```

```txt
N/A
```

### Additional information

_No response_

mdegat01:
Bump config check base from `2022.02.0` to `2022.05.0` to match core

Fixes #2512

### DIFF
--- a/check_config/CHANGELOG.md
+++ b/check_config/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.10.1
+
+- Update base image to 2022.05.0
+
 ## 3.10.0
 
 - Update base image to 2022.02.0
--- a/check_config/build.yaml
+++ b/check_config/build.yaml
@@ -1,10 +1,10 @@
 ---
 build_from:
-  aarch64: ghcr.io/home-assistant/aarch64-homeassistant-base:2022.02.0
-  amd64: ghcr.io/home-assistant/amd64-homeassistant-base:2022.02.0
-  armhf: ghcr.io/home-assistant/armhf-homeassistant-base:2022.02.0
-  armv7: ghcr.io/home-assistant/armv7-homeassistant-base:2022.02.0
-  i386: ghcr.io/home-assistant/i386-homeassistant-base:2022.02.0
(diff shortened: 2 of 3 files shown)


frenck (approved):
✅ 

We should consider deprecating this add-on IMHO. It doesn't hold its original value anymore (it originates from the time that upgrading with a configuration breaking change would cause your system not to come up again, which is no longer the case).

mdegat01:
@frenck idk I mean I still find the idea of it pretty useful personally? It would be nice to know before you updated and wound up in emergency mode that something is wrong about your config. Although its useful does depend on how many yaml-configurable integrations you have since it does nothing for UI-configurable ones

frenck:
> It would be nice to know before you updated and wound up in emergency mode that something is wrong about your config.

That is the whole point, updating won't trigger that.

ludeeus:
Since we are not testing the configuration against the same container image as what you get when you are upgrading, this add-on has really never been that much help and have always caused confusion and false results.
I'm in favor of deprecating it.

If things go bad, the supervisor will do a rollback.

PR #2528, opened 2022-06-13
zwave: Remove old OpenZWave add-on

frenck:
This PR suggests removing the OpenZWave add-on. 

The OpenZWave integration has been deprecated for a long time and removed from Home Assistant a couple of releases ago.


MartinHjelmare:
How does this work for users running an old core version still using the ozw integration with this add-on?

frenck:
The current installation will just keep working (the add-ons do not stop or get un-installed or something).

MartinHjelmare:
Will the user be able to install the add-on on a new host?

MartinHjelmare:
I think we should wait a year after we removed the ozw integration before removing the add-on.

PR #2530, opened 2022-06-14
Fix TLSv1.3

### LINKED ISSUE #2527: nginx_proxy TLS 1.3

As TLS 1.3 is enabled via config id didn’t work. Only tls 1.2 works. This can be checked via did tools

In ngix.conf line 51 variable for hsts is used and I think the variable didn’t has the ; ad end so the config for ssl_protocols wollen be found. 

Home Assistant OS

NGINX Home Assistant SSL proxy

Stable

zeehio:
Closes #2527.

According to https://serverfault.com/questions/1097716/tls-1-3-not-working-on-nginx-1-21-with-openssl-1-1-1n and to my own testing (using `curl --tlsv1.3 <URL>`), tlsv1.3 was not working until I moved those lines up (from the server block to the http block)

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.1.2
+
+- Fix TLSv1.3 support
+
 ## 3.1.1
 
 - Hide server version banner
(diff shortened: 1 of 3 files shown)


mdegat01 (changes requested):
Good catch! Tested and confirmed it doesn't do TLS 1.3 before and does after. Also that the port 80 redirect still works.

@zeehio Please bump the version and update the changelog though.

zeehio:
Done

mdegat01:
Thanks @zeehio !

PR #2532, opened 2022-06-17
nginx_proxy: reject SSL handshake with incorrect domain

pyos:
This is slightly faster (terminates after client hello) and removes the need for a self-signed certificate.


mdegat01 (changes requested):
Looks good to me, tested and confirmed working. Since an update was just merged to this addon though please rebase and bump the version to a new one @pyos

pyos:
Done

mdegat01 (approved):

mdegat01:
Thanks @pyos !

PR #2538, opened 2022-06-20
Update DOCS.md

james28909:
windows users will likely need to eventully add the home assistant ip and hostname to the host file, especially if the have had multiple home assistant instances in the past, or the windows ip table gets misconfigured somehow.

### DIFF
--- a/samba/DOCS.md
+++ b/samba/DOCS.md
@@ -18,7 +18,7 @@ Follow these steps to get the add-on installed on your system:
 
 ## Connection
 
-If you are on Windows you use `\\<IP_ADDRESS>\`, if you are on MacOS you use `smb://<IP_ADDRESS>` to connect to the shares.
+If you are on Windows you use `\\<IP_ADDRESS>\` (Or make sure your host file has an entry `<homeassistant ip address>    HOMEASSISTANT`), if you are on MacOS you use `smb://<IP_ADDRESS>` to connect to the shares.
 
 This addon exposes the following directories over smb (samba):
 


homeassistant:
Hi @james28909, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2538).

Thanks!

samba/DOCS.md
```diff
@@ -18,7 +18,7 @@ Follow these steps to get the add-on installed on your system:
 
 ## Connection
 
-If you are on Windows you use `\\<IP_ADDRESS>\`, if you are on MacOS you use `smb://<IP_ADDRESS>` to connect to the shares.
+If you are on Windows you use `\\<IP_ADDRESS>\` (Or make sure your host file has an entry `<homeassistant ip address>    HOMEASSISTANT`), if you are on MacOS you use `smb://<IP_ADDRESS>` to connect to the shares.
```
frenck, line 21: Your PR description says:

> windows users will likely need to eventully add the home assistant ip and hostname to the host file

However, our documentation lists everything by IP. I'm confused, why the hosts file edit is even needed for this example? As a matter of fact, since Home Assistant uses LLMNR and mDNS, one should not even have to use an IP address and should be able to use the name directly (without editing the hosts file).

For that reason, I think it's better not to suggest edits like these (which people should not need).

frenck:
Closing the PR for the above-listed reason. Nevertheless, thanks for willing to contribute 👍

samba/DOCS.md
james28909: i will also add it is a simple line in a read me. if you dont want to add it in the read me then make another file with possible fixes for problems people are having. dont just chalk it up to "people are idiots". thanks

james28909: look. the host name breaks for some reason. there is nothing in the log files about why the hostname stops working, and noone has even tried to provide a fix or any work around, this will help people whos hostname has broken. and it does not mess with any other domains or sub domains ni home assistant, i am behind a reverse proxy with a good handfull of domains and sub dmains, and HOMEASSISTANT from samba is the only one who stops working. it is most likely a windows issue, but this is a quick fix to get back into samba shares in case of emergency

james28909: > (which people should not need).

dude, use google, and also i am here i your ear about it. i am sharing a cure for it. take it or leave it i simply couldnt care any less. my own machine is working as it should now though. have a good one

frenck: > , you can blow it off as a non issue but the people who use this add on, and it is many, do suffer from this issue.

The documentation shows it by IP. This will always work, regardless of anything else.
You may call that blowing it off, that is fine.

In the end, the issue you are experiencing is caused by network misconfiguration on the user end, which is not something we can ever solve.

james28909: > > , you can blow it off as a non issue but the people who use this add on, and it is many, do suffer from this issue.
> 
> The documentation shows it by IP. This will always work, regardless of anything else. You may call that blowing it off, that is fine.
> 
> In the end, the issue you are experiencing is caused by network misconfiguration on the user end, which is not something we can ever solve.

there is nothing about my machine that is misconfigured. it is working great. the samba share add on in this repo is the only problem i had. as a matter of fact, i chased down the issue and shared it with you. the addon <-> windows commuication breaks only for this add on. no other entity in my file explorer ever gives me these problems unless the endpoint is unreachable.

frenck: Congratulations, you got yourself a ban.

PR #2542, opened 2022-06-23
Bump zwave-js-server to 1.20.0

raman325:
https://github.com/zwave-js/zwave-js-server/releases/tag/1.20.0


raman325:
tested inclusion on 2022.5.5 and was able to pair the device successfully. Going to go ahead and merge so we can let people move forward with setting up their networks

PR #2544, opened 2022-06-23
Deprecate config check addon

mdegat01:
This addon does not work. It shows false positive errors every time which makes it not useful for its defined purpose (telling you whether your config will work after the update).

Upon review it is clear recent updates to core have made this addon irrelevant. Core will always start after an update now and if something is wrong with your config it will tell you what broke. It does this for YAML and config entries (which this addon never supported) automatically and with a better UI (persistent notification instead of checking addon logs).

Therefore this addon should be deprecated.

### DIFF
--- a/check_config/CHANGELOG.md
+++ b/check_config/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 3.11.0
+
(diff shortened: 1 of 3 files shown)


frenck (approved):
LGTM 👍

PR #2545, opened 2022-06-23
Samba: Don't mangle names by default

### LINKED ISSUE #2541: Samba add-on not passing long filenames to MAC SMB client - ":" in filename causes issues

I have a bunch of files (images) who's names are of the format "Garden-2022-06-20_11:05:00.jpg".

When I make a SMB connection from my MAC (MacOS 12.4), I get filenames like: "GRJUYC~U.JPG"

image

Looks like the : in the filename is stopping the expansion...I can SFTP the files to my mac without issue.

Home Assistant OS

Samba share

9.7.0

1. On HA host create a file in /config/www called "Some:File.txt"
2. Open SMB connection on MAC and open config/www
3. Observe mangled filename
...

```txt
Nope
```

```txt
Nothing
```

### Additional information

I suspect it's because Windoze can't deal with a : in the filename...

This - https://serverfault.com/questions/124611/special-characters-in-samba-filenames - might be a useful pointer.  Perhaps more SMB options need to be exposed in the AddOn.

mdegat01:
Samba defaults to mangling names with special characters in order to be compatible with [really old versions of windows](https://www.oreilly.com/openbook/samba/book/ch05_04.html). Applied the fix suggested [here](https://serverfault.com/a/765951) by default with a new compatibility option if people need to revert this behavior.

This is a breaking change since it changes the default with a new revert option rather then the other way around. Although I imagine its one users will be happy with. 

Fixes #2541 

Also bump base image to Alpine 3.16

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -2,14 +2,15 @@
 
 ## 10.0.0
 
-BREAKING CHANGE: Don't mangle names by default
+BREAKING CHANGE: Don't mangle filenames
 
 By default Samba mangles filenames with special characters to ensure
-compatibility with all possible clients (including really old versions
-of Windows). The default is changed to not do this anymore. If you
-need this behavior add `mangled_names: true` to the config.
+compatibility with really old versions of Windows which have a very limited
+charset for filenames. The addon no longer does this as modern operating
+systems do not have these restrictions.
 
-- Don't mangle names by default (fixes #2541)
+- Don't mangle filenames (fixes #2541)
+- Upgrade Alpine Linux to 3.16
 
 ## 9.7.0
 
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 BREAKING CHANGE: Don't mangle filenames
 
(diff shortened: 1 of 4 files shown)


samba/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 9.7.0
+version: 10.0.0
```
frenck, line 2:
```suggestion
version: 10.0.0
```

samba/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 10.0.0
+
+BREAKING CHANGE: Don't mangle filenames
+
+By default Samba mangles filenames with special characters to ensure
+compatibility with really old versions of Windows which have a very limited
+charset for filenames. The addon no longer does this as modern operating
+systems do not have these restrictions.
```
frenck, line 10:
```suggestion
By default, Samba mangles filenames with special characters to ensure
compatibility with really old versions of Windows which have a very limited
charset for filenames. The add-on no longer does this as modern operating
systems do not have these restrictions.
```

frenck (approved):

PR #2549, opened 2022-06-23
Ensure all addons have init: false

mdegat01:
Not bumping the version and changelog, just want to make sure this doesn't break on next update of these addons.

Also setting `with-contenv` now that s6 is running the scripts for some of these.


frenck (approved):

frenck (approved):

mdegat01:
I also have to fix the finish scripts but that's going to be a change to basically all of the addons. Think I'll merge this then do that.

PR #2513, opened 2022-06-03
Add propagation delay for letsencrypt with RFC2136

mateuszdrab:
Reopening a new PR to replace https://github.com/home-assistant/addons/pull/1729

Fixes #1424

PR #2557, opened 2022-06-24
Fix letsencrypt typo again

mateuszdrab:
Fixes typo in https://github.com/home-assistant/addons/pull/2513#issuecomment-1165821876

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.12.5
+
+- Fix another syntax error in runs script for rfc2136
+
 ## 4.12.4
 
 - Fix syntax error in runs script for rfc2136
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 4.12.4
+version: 4.12.5
 slug: letsencrypt
 name: Let's Encrypt
 description: Manage certificate from Let's Encrypt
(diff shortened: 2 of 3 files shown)


mdegat01:
Wait that's strange, `--certbot-dns-rfc2136:dns-rfc2136-propagation-seconds` doesn't work? That's the same thing we use for all the other ones...

I know the certbot prefix is old now but I thought it still worked. Maybe rfc2136 is newer so it was never supported for it?

mdegat01 (approved):
Huh yea looks like the old certbot prefix is not supported for rfc2136 plugin arguments but still supported for the other plugins. Should probably remove it from them too since its not even documented anymore. Separate PR though.

mateuszdrab:
Yeah, the good news is that it works now:
```
[23:40:28] INFO: Selected DNS Provider: dns-rfc2136
[23:40:28] INFO: Use propagation seconds: 120
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Certificate not yet due for renewal
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Certificate not yet due for renewal; no action taken.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```
Finally don't have to rely on 'luck' to get my certs renewed (DNS transfer to slaves takes about a minute).

You're right though, looks like the addon needs a review - the run script is a mix of different options some referring to the variable and some hard-coding the provider name.

PR #2568, opened 2022-07-03
Update config.yaml

DominoDev:
Added rw flag to share mapping.  This will allow logging to the share directory.

### DIFF
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -15,7 +15,7 @@ image: homeassistant/{arch}-addon-nginx_proxy
 init: false
 map:
   - ssl
-  - share
+  - share:rw
 options:
   domain: null
   hsts: max-age=31536000; includeSubDomains


frenck:
It should not have write access to that folder.

Logging can be done towards the stdout, which gets recorded via Docker in journald (which can be accessed in several ways).

journald would handle that much better compared to a static log path of nginx (e.g., rotates and protects against running out of diskspace).

mdegat01:
Agree with Frenck. Logging is not a good reason to add rw access here as docker containers should not be logging to files. They should log to stdout so docker can manage it. We already have journal logging set up which captures everything and manages it.

DominoDev:
I have 3 static sites hosted on my server using the nginx proxy addon.  I am trying to setup local statistics reporting using awstats.  To facilitate that I would like to have the option to write my access and error log files for the sites to something other than stdout.

frenck:
Hi there @DominoDev 👋 

Don't get me wrong, but that is a pretty uncommon use case. And I'm sorry to say, we do not have the intention to support and build options for every possible use case out there. Mainly to keep things maintainable (across the whole project, not this add-on).

../Frenck

DominoDev:
@frenck I understand that edge cases should not and can not be a development priority.  However, I can't see how this would be a maintainability issue as there are already several addons that do have write access to the share folder.  I can't see how this would introduce any risk of being a breaking change to the project.  If we can't move forward with this I would love to have an offline discussion with you to pick your brain on how I can pull usable statistical data from the journald logs.

frenck:
> edge cases should not and can not be a development priority

At no point, do I talk about priority? This is about maintainability.

> However, I can't see how this would be a maintainability

Adding support for edge cases means you carry "options" forward for a long time. Accepting all options in our project for each edge case will end up with something that is hard to keep in control, change or improve without breaking things. Thus, in general, if we add something, there has to be a bigger use case.

Besides, this PR adds write access that is not needed in general. When keeping things neat and secure, we should keep permission minimal.

> If we can't move forward with this I would love to have an offline discussion with you to pick your brain on how I can pull usable statistical data from the journald logs.

This add-on is not made with that goal in mind. For more information on journald, please consult their documentation. For enabling add-on access to journald, see our add-on configuration developer documentation: <https://developers.home-assistant.io/docs/add-ons/configuration>

PR #2582, opened 2022-07-07
Update mariadb base to 3.16

mdegat01:
Update mariadb to alpine 3.16 and s6 v3

### DIFF
--- a/mariadb/config.yaml
+++ b/mariadb/config.yaml
@@ -10,8 +10,8 @@ arch:
   - aarch64
   - amd64
   - i386
-backup_post: s6-svc -wd -d /run/s6/services/mariadb-lock/
-backup_pre: s6-svc -wU -u /run/s6/services/mariadb-lock/
+backup_post: s6-svc -wd -d /run/s6/legacy-services/mariadb-lock/
+backup_pre: s6-svc -wU -u /run/s6/legacy-services/mariadb-lock/
 image: homeassistant/{arch}-addon-mariadb
 init: false
 options:


pvizeli:
Not sure if that work with the PATH, I guess exec commands not going over the ENTRYPOINT and it will miss the PATH adjustments?

pvizeli:
nvm, it seems to work somehow:
image

pvizeli (approved):

mariadb/config.yaml
```diff
@@ -10,8 +10,8 @@ arch:
   - aarch64
   - amd64
   - i386
-backup_post: /bin/s6-svc -wd -d /run/s6/services/mariadb-lock/
-backup_pre: /bin/s6-svc -wU -u /run/s6/services/mariadb-lock/
+backup_post: s6-svc -wd -d /run/s6/services/mariadb-lock/
+backup_pre: s6-svc -wU -u /run/s6/services/mariadb-lock/
```
pvizeli, line 14:
```suggestion
backup_post: s6-svc -wd -d /run/services/mariadb-lock/
backup_pre: s6-svc -wU -u /run/services/mariadb-lock/
```

mdegat01: See my comment below. `/run/s6/services/mariadb-lock/` and `/run/services/mariadb-lock/` both resulted in error. Only `/run/s6/legacy-services/mariadb-lock/` seemed to work.

pvizeli:
Nice work!

frenck:
> So I don't have explanation for this since I read the same doc you both did but I can assure you it actually needs to be /run/s6/legacy-services/mariadb-lock/. I just tested it locally. 

Very odd, but I take it 😄 

nice work indeed 👍

PR #2595, opened 2022-07-16
letsencrypt: Add DNS Hetzner provider

cosandr:
Tested on an x86_64 appliance

```
Home Assistant 2022.7.5
Supervisor 2022.07.0
Operating System 8.2
Frontend 20220707.1 - latest 
```


homeassistant:
Hi @cosandr, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2595).

Thanks!

mdegat01 (changes requested):
Looks good. Bump the version in `config.yaml` and add an entry to the changelog and this is ready.

cosandr:
Rebased and bumped versions as requested.

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -91,6 +91,11 @@ elif [ "${DNS_PROVIDER}" == "dns-gandi" ]; then
     fi
     PROVIDER_ARGUMENTS+=("--authenticator" "certbot-plugin-gandi:dns" "--certbot-plugin-gandi:dns-credentials" /data/dnsapikey)
 
+# Hetzner
+elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-hetzner" ]; then
+    bashio::config.require 'dns.hetzner_api_token'
```
mdegat01, line 96:
```suggestion
```
Shoot sorry I missed this the first time I guess but don't do config validation in here. Config validation like has to be in `cont-init.d` if it's necessary. The problem with doing it in here is that will cause the run script to fail and when it fails then s6 just keeps restarting it.

mdegat01 (changes requested):

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
cosandr: I put it there because other plugins already do it in there (directadmin, njalla, azure to name a few). I don't see anything similar in `cont-init.d`. If you're OK with removing it entirely I can do that.

mdegat01: Hm yea I see others are doing it. Ok fair enough we can leave it. It may put the addon in a restart loop but I suppose at least the user will clearly see what they did wrong.

mdegat01 (approved):

mdegat01:
Thanks @cosandr ! 👍

PR #2347, opened 2022-01-13
Add Azure DNS challenge support

backerman:
As it says on the tin. Tested on my RPi3 running stable.

PR #2605, opened 2022-07-30
Fix Let's Encrypt documentation for Azure DNS.

backerman:
The documentation for Azure DNS support in the Let's Encrypt plugin (#2347) misstated a parameter; this patch corrects my error.


mdegat01 (approved):
@backerman can you rebase? That will fix that "head commit is behind error in the CI". But yea looks good otherwise

PR #2610, opened 2022-08-03
Add option `use_identity_as_username` to Mosquitto

mdegat01:
Apparently some devices do not allow you to provide a username and password and only allow you to use a client certificate for authentication. Enabling this option to help those users:
https://community.home-assistant.io/t/publishing-data-to-mqtt-broker-using-only-with-certificate/446470

### DIFF
--- a/mosquitto/config.yaml
+++ b/mosquitto/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 6.1.3
+version: 6.2.0
 slug: mosquitto
 name: Mosquitto broker
 description: An Open Source MQTT broker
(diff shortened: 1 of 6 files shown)


mdegat01:
As an aside this option appears to be noticeably faster on connect when testing from MQTT explorer. Which won't matter most of the time since clients only authenticate once but if you have a device/service which doesn't maintain a persistent connection it might matter. At least for some super advanced users willing to mess around with client certificates.

mosquitto/config.yaml
```diff
@@ -39,6 +39,7 @@ schema:
   certfile: str
   cafile: str?
   keyfile: str
+  use_identity_as_username: bool?
```
pvizeli, line 42: I guess we need an default for that?

rforro:
I don't understand why this never got merged. I wanted to move from debian + docker to HASSOS but now I'm stuck because all my MQTT devices are using certificates instead of user/pass combo.

PR #2611, opened 2022-08-09
Add/Fix Azure section

icyfire0573:
I went through this and couldn't get this to work without looking at error messages
first problem was it didn't say that i need the azure_config and what that did. I corrected that (row 165)
second problem was a style issue? or me trying to fix this based on experience.  I'd have tweaked more but ran out of time.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -162,18 +162,21 @@ challenge: dns
 dns:
   provider: dns-azure
   azure_creds: azure.txt
+  azure_config: azure.txt
 ```
 
-Please copy your credentials file "azure.txt" into the "share" shared folder
+Please copy your configuratino file "azure.txt" into the "share" shared folder
 on the Home Assistant host before starting the service. One way is to use the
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @icyfire0573, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;addons#2611).

Thanks!

backerman:
I've got #2605 in the queue to fix the issue with the nonexistent `azure_creds` parameter (correct: `azure_config`).

mdegat01:
Yea I think we just need to fix the doc bug in #2605. I get that you're trying to make the steps more explicit but to be honest I don't want that in this doc. I think the doc for this addon should basically just link to the certbot plugin doc. The problem is that we (developers in this repo) don't own this setup process. It can change at any time because Microsoft or the developers of that plugin made a change. And then now the addon needs to be updated because the doc it out of date.

Better to simply link to instructions maintained by the folks who actually maintain the plugin.

PR #2624, opened 2022-08-21
Fix my.home-assistant.io link for Supervisor logs in Issue form

DCSBL:
The logs for supervisor were moved, but the issue form was not updated.


frenck (approved):
Thanks, @DCSBL 👍

PR #2626, opened 2022-08-21
letsencrypt: add ionos dns support

jandelgado:
IONOS provides API support to change DNS configuration which is
supported by certbot. See https://developer.hosting.ionos.de/docs/dns
and https://github.com/helgeerbe/certbot-dns-ionos

This is more or less what was already tried with #2331, but with the review comment incorporated. 

Tested on a Raspberry 4 (arm64)


letsencrypt/build.yaml
```diff
@@ -17,3 +17,4 @@ args:
   CERTBOT_NETCUP_VERSION: 1.1.1
   CERTBOT_NJALLA_VERSION: 1.0.0
   CERTBOT_VERSION: 1.21.0
+  CERTBOT_DNS_IONOS_VERSION: 2022.05.15
```
mdegat01, line 20:
```suggestion
  CERTBOT_DNS_IONOS_VERSION: 2022.5.15
```
It looks like pip figures it out if you use `2022.05.15` but the version is actually `2022.5.15` according to pip:
Screen Shot 2022-08-29 at 11 34 14 AM

mdegat01 (changes requested):
You need to bump the version in `config.yaml` to `4.13.0` as well. You only added it to the changelog.

jandelgado:
all fixed & rebased

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -119,6 +119,10 @@ elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-azure" ]; then
     fi
     PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-config" "/data/azure_creds")
 
+# IONOS
+elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-ionos" ]; then
+    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /data/dnsapikey)
```
mdegat01, line 124:
```suggestion
    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" /data/dnsapikey "--${DNS_PROVIDER}-propagation-seconds" "${PROPAGATION_SECONDS}")
```
Need to support the propagation seconds option for this provider.

mdegat01 (changes requested):

jandelgado:
@mdegat01 anything I can do to get this PR merged?

PR #2637, opened 2022-08-29
DNSMasq - default cnames to empty array

### LINKED ISSUE #2636: DNSMASQ v1.5.0 just fails to start

After upgrade 1.4.4.->1.5.0. it fails to start:
image

By docs - cname is optional config, as well Configuration page does not provide possibility to define CNAME's.

Home Assistant OS

Dnsmasq

1.5.0

Do upgrade 1.4.4. -> 1.5.0 and tried to start it.

```txt
Nothing in Supervisor logs about add-on.
```

```txt
No add-on log available.
```

### Additional information

v1.4.4. add-on config is very simple:
```
defaults:
  - 192.168.1.6
forwards: []
hosts:
  - host: ha.local
    ip: 192.168.1.11
<couple hosts defined>
services: []
```

mdegat01:
Default cnames setting to empty array to prevent errors with existing configs.

Fixes #2636

### DIFF
--- a/dnsmasq/config.yaml
+++ b/dnsmasq/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.5.0
+version: 1.5.1
 slug: dnsmasq
 name: Dnsmasq
 description: A simple DNS server
@@ -20,6 +20,7 @@ options:
   forwards: []
   hosts: []
   services: []
+  cnames: []
 ports:
   53/tcp: 53
   53/udp: 53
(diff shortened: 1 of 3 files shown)


dnsmasq/config.yaml
```diff
@@ -20,6 +20,7 @@ options:
   forwards: []
   hosts: []
   services: []
+  cnames: []
```
ludeeus, line 23: I don't think we should do this.
It should be marked as optional in the schema instead.

ludeeus: The same goes for the others, options we do not expect the majority of users to use, should be optional.

PR #2647, opened 2022-09-03
Update DOCS.md - Azure example azure_creds -> azure_config

samyun:
Seems at some point, certbot was updated. Option list shows `azure_config`, but the example seems to have been missed.


mdegat01:
Thanks @samyun 👍

PR #2650, opened 2022-09-05
Use edge version of NodeJS in zwave_js

ludeeus:


zwave_js/Dockerfile
```diff
@@ -8,8 +8,9 @@ ARG ZWAVEJS_VERSION
 WORKDIR /usr/src
 RUN \
     set -x \
+    && echo "@edge http://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
```
ludeeus, line 11:
```suggestion
    && echo "@edge https://dl-cdn.alpinelinux.org/alpine/edge/main" >> /etc/apk/repositories \
```

zwave_js/Dockerfile
```diff
@@ -8,8 +8,9 @@ ARG ZWAVEJS_VERSION
...
     && apk add --no-cache \
-        nodejs \
+        nodejs@edge \
         npm \
```
ludeeus, line 14:
```suggestion
        npm@edge \
```

balloob (approved):

MartinHjelmare:
I've built this locally and tested it. It works, but I can't verify if it solves the memory leak.

PR #2678, opened 2022-09-23
Use node-zwave-js with `xstate` pinned to an older version in Z-Wave JS

AlCalzone:
That version is a [test build](https://github.com/zwave-js/node-zwave-js/pull/5108), which is basically 10.2.0, but with xstate pinned to 4.29.0 - the version node-zwave-js had in its own lockfile. This seems to be the most obvious difference in dependencies between the addons.

With this change, the core addon will use an older version of the dependency than the community addon (4.32.1), but one that was previously in use and did not leak. With 4.32.1, the driver does not compile - they are using some funky types.

@raman325 @MartinHjelmare I'm not sure if anything else needs to be changed (server versions etc.)

Closes https://github.com/home-assistant/addons/issues/2683

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.73
+
+- [Bump Z-Wave JS Server to 1.23.0](https://github.com/zwave-js/zwave-js-server/releases/tag/1.23.0)
+
 ## 0.1.72
 
 - Use same base image as community add-on zwave-js-ui
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## 0.1.73
 
 - [Bump Z-Wave JS Server to 1.23.0](https://github.com/zwave-js/zwave-js-server/releases/tag/1.23.0)
+- [Bump Z-Wave JS to 10.2.0](https://github.com/zwave-js/node-zwave-js/releases/tag/v10.2.0)
+- Bump Z-Wave JS to test build to try to address memory leak
 
 ## 0.1.72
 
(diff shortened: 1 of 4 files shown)


homeassistant:
Hi @AlCalzone

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant/addons#2678).

Thanks!

MartinHjelmare:
I started the CI build.

MartinHjelmare:
Looks like we need to bump zwave-js in the server first.

raman325:
I don't get why, there were no breaking changes in 10.2.0

raman325:
I just published a new release, @AlCalzone can you bump the server version as well?

raman325:
we should also update the changelog

kpine:
You'll have to use `npm install --force` to bypass the peer dependency versioning error. The PR build is considered a pre-release version, so it won't ever match the server's pinned version.

Or you can install `xstate@4.29.0` manually at the same time as `zwave-js@10.2.0` (PR version not required).

AlCalzone:
Not sure if the latter will work, npm sometimes does strange stuff and could cause two different versions to be installed.

--force should work.

raman325:
I added the force flag and everything built properly. Not home this weekend so someone else will need to validate that the addon works before we merge

AlCalzone:
How can I do that again? Just copy  the directory in my local addons folder on the yellow, right?

MartinHjelmare:
Yes, see this section for the details:
https://developers.home-assistant.io/docs/add-ons/testing#remote-development

AlCalzone:
Still getting confused by the process of switching addons. This is all but easy.

Anyhow, the addon starts, I can include and exclude and control devices.

pvizeli (commented):
We should also jump back to our base image

pvizeli (approved):

pvizeli (approved):

PR #2707, opened 2022-10-15
Adding support for DuckDNS challenge

miguelrjim:
Installing plugin for DuckDNS certbot challenge validation, tested it against my homeassistant instance by installing the addon locally.


Zehir:
Any updates ?

mdegat01 (changes requested):
So the PR looks good except you need to add an entry to the [changelog](https://github.com/home-assistant/addons/blob/master/letsencrypt/CHANGELOG.md).

Also in this particular case, can you explain why we need this? Because the [DuckDNS addon](https://github.com/home-assistant/addons/tree/master/duckdns) already exists. So this seems redundant?

Zehir:
> So the PR looks good except you need to add an entry to the [changelog](https://github.com/home-assistant/addons/blob/master/letsencrypt/CHANGELOG.md).
> 
> Also in this particular case, can you explain why we need this? Because the [DuckDNS addon](https://github.com/home-assistant/addons/tree/master/duckdns) already exists. So this seems redundant?

Good point, I using the duckdns addon too

miguelrjim:
> So the PR looks good except you need to add an entry to the [changelog](https://github.com/home-assistant/addons/blob/master/letsencrypt/CHANGELOG.md).
> 
> Also in this particular case, can you explain why we need this? Because the [DuckDNS addon](https://github.com/home-assistant/addons/tree/master/duckdns) already exists. So this seems redundant?

Mostly to give users a different way to manage their certificates; for reference, I'm using this [blueprint]( https://gist.github.com/mcnovy/caa97512ed7aedbe3f407bc7bbabe737) to manage renewal & restarting homeassistant to pick up the renewed certificate.

PR #2712, opened 2022-10-19
added WSDD daemon to Samba addon

L2jLiga:
Added WebServiceDiscovery daemon to Samba addon, this makes share discoverable in Windows 10/11 by default.


pvizeli:
Unsure if that end up in a conflict with new Core WSD service?

L2jLiga:
Can you share some details about new service?

L2jLiga:
It will not conflict, but just will not work if someone else already took ports used by WSD

L2jLiga:
Hi @pvizeli 
Was this service added in 2022.11? Now I can see 2 devices in network (with this PR applied): 
image

And without this PR I also see same 2 devices:
image

For both cases:
Double click on "Bender (Home Assistant)" opens web application
Double click on "Bender" opens Samba share

Definitely this PR does not break anything but brings ability to see Samba share in Network tab for earlier Home Assistant releases, so I would not say new update makes this PR unnecessary

madbrain76:
The discovery is not working for me in Win11. "Network" does not show the home assistant system. If I use "Net use x: \\homeassistant.localdomain\config", it does work.

On other Linux-based system, I solved this by adding the wsdd-server service. It used to be called just "wsdd". I am wondering if the renaming is what's the causing the issue, or something else. Is there a way to check if this service is running ?

PR #2714, opened 2022-10-20
Allow nginx ssl proxy to have an empty "hsts" setting

LordMike:
The docs say that "if empty the header is not sent", but the field is required.

image


LordMike:
Was this never accepted? I just realized. Wow.
@frenck could you open+merge this

PR #2723, opened 2022-10-24
Use root as the user inside the devcontainer

ludeeus:
VSCode has changed the default user to `vscode`, this PR set that back to `root`


ludeeus:
https://github.com/home-assistant/supervisor/pull/3969#discussion_r1003527220

agners (approved):
For addons, let's continue use root at least for now.

PR #2746, opened 2022-11-15
Add matter server discovery

MartinHjelmare:
- Requires a Supervisor with support for matter discovery service.
- See https://github.com/home-assistant/supervisor/pull/4002


MartinHjelmare:
I need to test this when we have a Supervisor with the discovery service for matter included published (dev).

MartinHjelmare:
I've tested this PR locally with my add-on flow branch for Core that I'm working on.

agners (approved):

PR #2763, opened 2022-11-25
silabs-multiprotocol: Set valid default for autoflash option

agners:


agners:
Also add amd64 batch do the readme and hint about flashing ability.

frenck (approved):

PR #2766, opened 2022-11-27
Update DOCS.md

dlangille:
dnssec-keygen no longer creates 

% dnssec-keygen -a HMAC-SHA512 -b 512 -n HOST letsencrypt          21:31:02
dnssec-keygen: fatal: unknown algorithm HMAC-SHA512

From the man page:

In prior releases, HMAC algorithms could be generated for use as TSIG keys, but that feature was removed in BIND 9.13.0. Use tsig-keygen to generate TSIG keys.


mdegat01 (approved):

mdegat01:
Thanks @dlangille 👍

PR #2769, opened 2022-11-29
Add feature to set MariaDB variables on start

fapfaff:
Adding the possibility to set global variables upon the start of the addon.
Currently there is no possibility to persistently modify variables or the config.

### DIFF
--- a/mariadb/config.yaml
+++ b/mariadb/config.yaml
@@ -23,6 +23,7 @@ options:
   rights:
     - database: homeassistant
       username: homeassistant
+  variables: []
 ports:
   3306/tcp: null
 schema:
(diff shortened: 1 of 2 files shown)


fapfaff:
Oh, I accidentally deleted my fork, but it should still work, right?

mdegat01:
What is the use case for this? HA wouldn't make any use of these in normal execution since it wouldn't even know they existed. Is this useful for any reason other then user writing their own custom queries on the DB?

fapfaff:
I'm using a nextcloud addon within homeassistant. However it doesnt work sometimes because I get `Out of sort buffer memory error (sort_buffer_size)` sometimes, but it isn't possible to change these settings persistently currently.
[Here's](https://community.home-assistant.io/t/how-can-i-tweak-hass-io-mariadb-addon-settings/184494/18) also a thread discussing this issue.

DTidey:
I would support this enhancement. 
I have lots of sql sensors that query a custom database hosted on Mariadb. I now run out of concurrent connections and have to manually increase max_connections manually at startup.

PR #2780, opened 2022-12-03
DHCP server - option 121 (Classless Static Route)

k-wojcik:
I added the ability to configure option 121 (Classless Static Route) for selected networks.

### DIFF
--- a/dhcp_server/DOCS.md
+++ b/dhcp_server/DOCS.md
@@ -43,6 +43,11 @@ hosts:
   - name: webcam_xy
     mac: aa:bb:ee:cc
     ip: 192.168.1.40
+static_routes:
+  - subnet: 192.168.1.0
+    network: 192.168.2
+    mask: 24
+    gateway: 192.168.1.2
 ```
 
 ### Option: `domain` (required)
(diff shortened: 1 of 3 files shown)


Blakout007:
Hey
Great job with these dhcp-server.
There is more options than just 121... why not make it possible to define numbered options.
E.g. i would love to have option 43, that is Wireless-Server, for dhcp.
Hope for a change so this will be possible too.
/Jesper

dhcp_server/DOCS.md
```diff
@@ -132,6 +137,31 @@ The MAC address of the client device.
...
+
+This option defines settings for DHCP option 121 (Classless Static Route).
+It allows you to inject static network route with DHCP configuration.
+
+By default, non are configured.
+
+#### Option: `static_routes.subnet`
+
+The network schema/subnet (`networks.subnet`) for which the rule will apply.
+
```
mdegat01, line 150: With this config a user would do something like this:
```yaml
networks:
  - subnet: 192.168.1.0
    ...
static_routes:
  - subnet: 192.168.1.0
    network: 192.168.2.0
    mask: 24
    gateway: 192.168.1.3
```
Instead of making them repeat `subnet`, lets just embed this in the network items. Like this:
```yaml
networks:
  - subnet: 192.168.1.0
    ...
    static_routes:
      - network: 192.168.2.0
        mask: 24
        gateway: 192.168.1.3
```

mdegat01 (changes requested):
In addition to the config change requested, you also need to bump the version of the addon in `config.yaml` and add an entry to the changelog detailing your change.

wolfwander:
> Hey Great job with these dhcp-server. There is more options than just 121... why not make it possible to define numbered options. E.g. i would love to have option 43, that is Wireless-Server, for dhcp. Hope for a change so this will be possible too. /Jesper

Instead of numbering options in configuration file, it would be a lot better adding the possibility of including a customized DHCP configuration.

Issue #2565, opened 2022-06-29
Lets Encrypt : option to choose the chain, certficate is not valid anymore on older devices


pergolafabio:
### Describe the issue you are experiencing

Hi, seems the Lets Encrypt addon is hardcoded to use this chain : 

`--preferred-chain "ISRG Root X1"`

https://github.com/home-assistant/addons/blob/dc81626fadedd3a89ed1b8e2ede1fb3ffd453d74/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L131

According to this thread, its the alternate chain

https://community.letsencrypt.org/t/production-chain-changes/150739

Is it possible to make an config option, so we can choose the chain type? seems the "X1" chain is not supported anymore on older android types

Thnx in advance

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

4.12.5

### Steps to reproduce the issue

No steps to reproduce

mdegat01:
[Platforms that trust ISRG Root X1](https://letsencrypt.org/docs/certificate-compatibility/#platforms-that-trust-isrg-root-x1):
> ...
> - [Android >= 7.1.1](https://android.googlesource.com/platform/system/ca-certificates/+/android-7.1.1_r15) (but Android >= 2.3.6 will work by default [due to our special cross-sign](https://letsencrypt.org/2020/12/21/extending-android-compatibility.html))

[Platforms that trust DST Root CA X3 but not ISRG Root X1](https://letsencrypt.org/docs/certificate-compatibility/#platforms-that-trust-dst-root-ca-x3-but-not-isrg-root-x1) - Notice no version of android is in this list

[Known Incompatible](https://letsencrypt.org/docs/certificate-compatibility/#known-incompatible):
> ...
> - Android < v2.3.6

If your android version is older then 2.3.6 then according to Let's Encrypt nothing will make it work. Except possibly switching to firefox as your mobile browser. If your android version is >= 2.3.6 then you should be using ISRG Root X1.

What is your android version and what actually is the issue you're facing? Did you confirm that changing the chain fixes it by running Let's Encrypt manually on some other system? If so please share the chain you used.

[...]

PR #2230, opened 2021-10-08
Upgrade certbot; add preferred chain flag

nkm8:
Context: Running the Adguard Home add-on with DNS-over-TLS enabled, I noticed that DoT on my Android phone stopped working earlier today. After some research, the issue is a a side-effect from the root cert expiration earlier today: https://scotthelme.co.uk/lets-encrypt-old-root-expiration/

The required flag `--preferred-chain <issuer CN>` was added in certbot 1.6.0: https://github.com/certbot/certbot/releases/tag/v1.6.0

Related discussions (with solutions provided):
- https://community.letsencrypt.org/t/android-devices-with-dot-configured-interaction-with-new-default-chain/161020/14
- https://github.com/AdguardTeam/AdGuardHome/issues/3689#issuecomment-931643378

PR #2791, opened 2022-12-08
Add option to configure the preferred chain

rrooggiieerr:
This should add an extra option to configure the preferred chain for certbot.

Configuring a different chain can help older Android versions to be able to use the generated certificate.

### DIFF
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -35,6 +35,7 @@ schema:
   challenge: list(dns|http)
   acme_root_ca_cert: str?
   acme_server: url?
+  chain: str?
   dns:
     aws_access_key_id: str?
     aws_secret_access_key: str?
--- a/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
+++ b/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
@@ -15,6 +15,7 @@ CHALLENGE=$(bashio::config 'challenge')
 DNS_PROVIDER=$(bashio::config 'dns.provider')
 ACME_SERVER=$(bashio::config 'acme_server')
 ACME_ROOT_CA=$(bashio::config 'acme_root_ca_cert')
+CHAIN=$(bashio::config 'chain')
 
 if [ "${CHALLENGE}" == "dns" ]; then
     bashio::log.info "Selected DNS Provider: ${DNS_PROVIDER}"
@@ -147,13 +148,13 @@ if [ "$CHALLENGE" == "dns" ]; then
         --email "$EMAIL" --agree-tos \
         --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
         --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}" "${PROVIDER_ARGUMENTS[@]}" \
-        --preferred-chain "ISRG Root X1"
+        --preferred-chain "$CHAIN"
 else
     certbot certonly --non-interactive --keep-until-expiring --expand \
         --email "$EMAIL" --agree-tos \
         --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
         --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}" "${ACME_CUSTOM_SERVER_ARGUMENTS[@]}" --standalone \
-        --preferred-chain "ISRG Root X1"
+        --preferred-chain "$CHAIN"
 fi
 
 # Get the last modified cert directory and copy the cert and private key to store
--- a/letsencrypt/translations/en.yaml
+++ b/letsencrypt/translations/en.yaml
@@ -29,6 +29,12 @@ configuration:
       By default, The addon uses Let's Encrypt's default server at
       https://acme-v02.api.letsencrypt.org/. You can instruct the addon to use a
       different ACME server.
+  chain:
+    name: Preferred Certificate Chain
+    description: >-
+      If the CA offers multiple certificate chains, prefer the chain whose
+      topmost certificate was issued from this Subject Common Name. If no match,
+      the default offered chain will be used.
   dns:
     name: DNS
     description: DNS Provider configuration


tjorim:
Would this not be a breaking change? What if a user does not specify the chain? Maybe we should add the existing one as default in case it is empty?

rrooggiieerr:
@tjorim That's totally fine with me. Is there a specific reason the ISRG Root X1 is chosen over the default Let's Encrypt DST Root CA X3 chain? Also the SRG Root X1 works if the Let's Encrypt ACME server is used (which is the default) but the add-on also gives the option to select a different ACME server. In that case the ISRG Root X1 chain is not working because it will be unknown the the other ACME server.

So if you like the ISRG Root X1 to be the default chain if none is given then that can only be the case if Let's Encrypt ACME server is used.

On a side note, when a given chain is not known by a given ACME server it will fall back to it's default chain, so nothing will break if you would use the ISRG Root X1 chain on an other ACME server. It will just not be recognised and ignored. Just like the case is now since the ISRG Root X1 chain is hard coded for any ACME server.

rrooggiieerr:
@mdegat01, @nkm8, @ludeeus, what are your thoughts on this change?

nkm8:
The context for why this default was chosen can be found here: https://github.com/home-assistant/addons/pull/2230

As far as the PR, I think it looks good. But I concur with @tjorim - I think the current preferred chain value should be explicitly set in the config. This way, users will know what the default is without having to look it up.

rrooggiieerr:
@nkm8 I think #2230 actually pointed me to your name and why I mentioned you here ;-)

So #2230 is the reason users with older Android hardware were no longer able to use https since last April. Took me some time to find out it was the chain which was not the default Let's Encrypt anymore. Honestly I think the alternative chain as now used in the HA Let's Encrypt add-on should be the optional one and the default chain should be... default. Also since the optional chain doesn't make much sense when someone configures a different ACME server (I don't think there are many who do that though).

Anyways it would be great if I could use my old Android tablet again as a HA wall controller one way or the other so for me it doesn't really matter which chain the HA add-on uses as a default as long as I can override it in the configuration.

I don't have the development environment setup to fully develop the proposed change, I think you could see this pull request as a step toward implementing the final solution. What I would propose is to have an option in the add-on interface which becomes visible when "Show unused optional configuration options" is enabled. The option probably does not really matter to most users, only to users with older Android hardware or third-party apps like Adguard Home or to users with custom ACME servers.

nkm8:
Is there a situation where the currently configured chain doesn't work? My understanding is that  the "current" chain works for all clients and the "default" chain breaks some older devices. If the current chain works for all devices, I'm just trying to understand the motivation for making it configurable.

rrooggiieerr:
The current configured chain doesn't work on old Android 5.0 devices, the default chain does

rrooggiieerr:
https://github.com/home-assistant/addons/issues/2565

nkm8:
Based on reading through the newer issues, it sounds like what Let's Encrypt is saying in their documentation doesn't match what users are experiencing. I agree with making this configurable, but I think there is a debate to be had for what the default value should be.

Things may have changed, but when Let's Encrypt originally switched and I opened https://github.com/home-assistant/addons/pull/2230 I was experiencing issues with much newer versions of Android - 11.0 or 12.0, I think. I am not sure if Android 11/12/13 still requires `ISRG Root X1` to be set to work, but I also don't want to risk finding out, either.

Android 5.0 is ancient (released in 2014), so I would argue that the current `ISRG Root X1` should continue to be the default. I am willing to bet  that the number of people trying to get Android 5.0 working is slim to none.

rrooggiieerr:
Yes, Android 5.0 is ancient, but with the HA Android App my 7 inch tablet still makes a great control device for on the wall. And since there are more reports of the Let's Encrypt add on not working with Android 5 devices I'm not the only one who still has one laying around.

As said before, I don't really mind which chain is the default. Making it configurable might also give some more options to whoever used a custom ACME server or for whenever Let's Encrypt decides to add more alternative chains to chose from.

rrooggiieerr:
Is this pull request going to get any follow up?

pergolafabio:
I hope so, i also have android 5 here for my indoor tablet devices

letsencrypt/config.yaml
```diff
@@ -35,6 +35,7 @@ schema:
   challenge: list(dns|http)
   acme_root_ca_cert: str?
   acme_server: url?
+  chain: str?
```
mdegat01, line 38:
```suggestion
  chain: list(ISRG Root X1|DST Root CA X3)?
```
Based on the conversation, these are the two options. In that case we should list them rather then making users look them up and copy and paste in or potentially mistype things.

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -15,6 +15,7 @@ CHALLENGE=$(bashio::config 'challenge')
 DNS_PROVIDER=$(bashio::config 'dns.provider')
 ACME_SERVER=$(bashio::config 'acme_server')
 ACME_ROOT_CA=$(bashio::config 'acme_root_ca_cert')
+CHAIN=$(bashio::config 'chain')
```
mdegat01, line 18: This must default to `ISRG Root X1`. Most modern devices do not trust `DST Root CA X3` anymore which means certificates generated with the old default chain are useless for a majority of users. As is this is a breaking change that will break a vast majority of users on upgrade to satisfy a few.

Either this variable has to default to `ISRG Root X1` when omitted or add `chain: ISRG Root X1` to the `options` in `config.yaml` (which causes `chain` to default to that when not explicitly provided by the user).

I'll note that you are right the option makes no sense for a custom acme server. `preferred-chain` should be omitted when that option is used. If you want to handle that feel free else it can be done in a separate PR.

letsencrypt/translations/en.yaml
```diff
@@ -29,6 +29,12 @@ configuration:
       By default, The addon uses Let's Encrypt's default server at
       https://acme-v02.api.letsencrypt.org/. You can instruct the addon to use a
       different ACME server.
+  chain:
+    name: Preferred Certificate Chain
+    description: >-
+      If the CA offers multiple certificate chains, prefer the chain whose
+      topmost certificate was issued from this Subject Common Name. If no match,
+      the default offered chain will be used.
```
mdegat01, line 37:
```suggestion
      If you are using a much older Android device you may need to change this
      to `DST Root CA X3` to prevent certificate errors. Otherwise you should
      leave this option at `ISRG Root X1`.
```

mdegat01 (changes requested):
In addition to the suggested changes you need to bump the version of the addon in `config.yaml` and add an entry to the changelog.

However before that I have to note I'm not sure this change should be made. The use case seems extremely niche. All reasonably up to date devices (like updated in the past year) do not trust `DST Root CA X3` anymore. The only devices that could use this are old android devices.

But keep in mind that means this is the use case:
1. The *only* client you have connecting to HA is an old android device
2. That device is using SSL to talk to HA

Note that I really mean *only* with 1. Because if you also have an up to date device you want to talk to HA with it won't trust the certificate you make with this chain.

And I would guess that most people with an old android device they want to use to talk to HA have an old android tablet their trying to put on the wall in their house. Which means 2 is likely also pretty rare since devices that live on the same wifi as HA usually talk to it unencrypted.

I'm not really convinced we should add a new option for such a niche use case. Particularly one that will be quite confusing to nearly all users.

mdegat01:
@nkm8 old android devices did use to trust certs signed with ISRG Root X1 until the DST Root CA X3 cross-sign expired in September 2021 (https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/). The behavior being described makes sense to me based on what Let's Encrypt did and said. That being said the default definitely needs to remain at ISRG Root X1 and I'm really not convinced we should add this option at all. See my review notes above, the use case seems extraordinarily niche.

Personally I would just say anyone with an old android tablet should just have it talk to HA unencrypted. If they also want to enable encrypted communication with other devices then use the [nginx proxy addon](https://github.com/home-assistant/addons/tree/master/nginx_proxy). That way devices on LAN can talk to HA unencrypted and devices outside of LAN can use encryption.

pergolafabio:
Unencrypted doesn't allow chrome to create home shortcut apps, that's why I needed https on my android 5 :-)

mdegat01:
@pergolafabio why not just use the android app? I just asked dshokouhi and its supposed to support android 5 and above.

pergolafabio:
No, a specific WebView is needed, so it doesn't work anymore

dshokouhi:
> No, a specific WebView is needed, so it doesn't work anymore

If you have webview the app should work, one of the changes in android 5 was that webview is now updateable in the play store so it can keep up with new web technology. Are you by chance running a custom ROM? If so I recommend changing the default webview to Googles version as it has been found to be incompatible. 

On android 5 the app should work, if it doesn't feel free to check the android repo closed issues as many other users have been in a similar situation and solved the issue.

pergolafabio:
no, problem is, my android is doesnt have playstore on it, so bounded to installing apk, i tried with old a newer webview apk
HA was just hanging at login screen :-)

Therefore i wanted to use https connection and creating an chrome webapp , but i need a valid ssl for that

dshokouhi:
> no, problem is, my android is doesnt have playstore on it, so bounded to installing apk, i tried with old a newer webview apk
> HA was just hanging at login screen :-)

definitely off topic in this issue, but what you need to do is not only install that webview but select it or install it as the default webview on the device. Definitely check closed issues in the android repo, others have solved your exact issue :)

pergolafabio:
But isnt this PR about https certificates and the lets encrypt addon, lets focus on that, i dont want to use apk or companion :-)

Just want to create a SSL certficate i can use on android 5 :-)

mdegat01:
So just to be clear, the use case for this option is:
1) User has an old android device
2) User wants to encrypt communication with HA from that device
3) User has no newer devices which want to talk with HA encrypted (since they won't trust certs made with `DST Root CA X3`)
4) User wants a shortcut icon for HA on their home screen and does not want to install the companion app to get that

That is extremely niche. Adding options that are confusing to most users to support a tiny subset of users is not really the direction we take with addons. The goal is to support what most users want with minimal options. Users that need advanced options can run the software themselves, either by making their own addon or running it on another machine.

If that really is the only use case I'm going to say this PR should not be merged.

pergolafabio:
Hi, well, that's the use case for me , not sure about others

rrooggiieerr:
@mdegat01 can you point me to instructions on how to access HA on both port 80 and port 443? Using your suggested nginx proxy add-on always forwards port 80 to 443

nexeck:
My old tablet stopped accepting the letsencrypt cert yesterday.
Would be greater to have an easy option to fix this.

rrooggiieerr:
I moved to using the NGINX SSL proxy and using plain HTTP on port 8123 for my old Android tablet. Still would like it to be HTTPS or be able to run HA on port 80, however this is a workable solution for me. So I won't be putting more time in this change as it doesn't seem to be picked up by the code owner.

dedors:
I liked the idea of using old touch devices for home assistant, but I'm facing this issue with 2 android devices.
Unfortunately, the companion has no option to connect anyway, too.
I am hesitant to use a reverse proxy solely for this purpose.

PR #2817, opened 2022-12-28
Require minimum HA Core version for matter server

MartinHjelmare:
- Set minimum version of Home Assistant Core needed for the add-on. This version is decided based on the chip SDK version available in Home Assistant Core.
- Add update procedure docs for developers of the add-on.
- Simplify docker file installation of chip SDK. It's included in the matter server extra requirements.

### DIFF
--- a/matter_server/Dockerfile
+++ b/matter_server/Dockerfile
@@ -1,8 +1,6 @@
 ARG BUILD_FROM
 FROM $BUILD_FROM
 
-ARG HOME_ASSISTANT_CHIP_VERSION
-
 # Set shell
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
@@ -41,9 +39,6 @@ ARG MATTER_SERVER_VERSION
 
 # hadolint ignore=DL3013
 RUN \
-    pip3 install \
-       home-assistant-chip-clusters==${HOME_ASSISTANT_CHIP_VERSION} \
-       home-assistant-chip-core==${HOME_ASSISTANT_CHIP_VERSION} \
-    && pip3 install --no-cache-dir python-matter-server=="${MATTER_SERVER_VERSION}"
+    pip3 install --no-cache-dir python-matter-server[server]=="${MATTER_SERVER_VERSION}"
 
 COPY rootfs /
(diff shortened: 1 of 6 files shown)


matter_server/Dockerfile
```diff
@@ -41,9 +39,6 @@ ARG MATTER_SERVER_VERSION
 
 # hadolint ignore=DL3013
 RUN \
-    pip3 install \
-       home-assistant-chip-clusters==${HOME_ASSISTANT_CHIP_VERSION} \
-       home-assistant-chip-core==${HOME_ASSISTANT_CHIP_VERSION} \
-    && pip3 install --no-cache-dir python-matter-server=="${MATTER_SERVER_VERSION}"
+    pip3 install --no-cache-dir python-matter-server[server]=="${MATTER_SERVER_VERSION}"
```
marcelveldt, line 42: is it safe to assume that these versions will stay on par ?

MartinHjelmare: Which versions? The SDK version in the add-on and the SDK version required by the server?

marcelveldt: ah yes, now I see. Instead of specifying which version of the chip sdk to include, you just choose which version of the python lib, much better!

matter_server/README.md
```diff
@@ -13,3 +13,13 @@ integration communicates via WebSocket with this server.
 
 [aarch64-shield]: https://img.shields.io/badge/aarch64-yes-green.svg
 [amd64-shield]: https://img.shields.io/badge/amd64-yes-green.svg
+
+## Matter Server Update Procedure
```
MartinHjelmare, line 17: This will show in the add-on info page. Maybe we should move it to a separate markdown file, less visible for users?

MartinHjelmare:
I've tested building the add-on locally with this change successfully. :+1:

matter_server/README.md
MartinHjelmare: @frenck what would be your recommendation for how to document this? Move to eg `RELEASE.md` or keep it in the README.md, or something else?

frenck: Well it is user-facing, so I like your suggestion for a `RELEASE.md` 👍

frenck (approved):

frenck (approved):

Issue #2792, opened 2022-12-09
File editor - git version too old for SSH commit signing


caraar12345:
### Describe the issue you are experiencing

`File editor` is running `git version 2.32.4` which is too old to use SSH commit signing.

Would it be possible to get it updated, please?

_I'm really sorry if this doesn't count as an "issue related to one of the official add-ons"! But it technically is one I suppose? 😅_

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

File editor

### What is the version of the add-on?

5.4.2

### Steps to reproduce the issue

1. Set up SSH commit signing with `git config gpg.format ssh`
2. Use `git commit` then `git push` in `File editor`
3. Be met with an error: `error: unsupported format for gpg.format: ssh`

## System Information

version | core-2022.12.0b6
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.7
os_name | Linux
os_version | 5.15.72
arch | x86_64
timezone | Europe/London
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.28.4
Stage | running
Available Repositories | 1225
Downloaded Repositories | 54

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | 5 January 2023 at 00:00
relayer_connected | true
remote_enabled | true
remote_connected | true
alexa_enabled | true
google_enabled | true
remote_server | eu-west-2-3.ui.nabu.casa
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 9.2
-- | --
update_channel | beta
supervisor_version | supervisor-2022.11.2
agent_version | 1.4.1
docker_version | 20.10.17
disk_total | 56.6 GB
disk_used | 53.4 GB
healthy | true
supported | failed to load: Unsupported
board | generic-x86-64
supervisor_api | ok
version_api | ok
[...]

PR #2827, opened 2023-01-09
Update File editor to Alpine 3.16

caraar12345:
Continuation of #2792 - an actual PR rather than just a feature request 😊

### DIFF
--- a/configurator/config.yaml
+++ b/configurator/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 5.4.2
+version: 5.4.3
 slug: configurator
 name: File editor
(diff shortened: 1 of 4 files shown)


configurator/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 5.4.2
+version: 5.4.3
```
agners, line 2: We should bump at least minor.

caraar12345: No arguments here - bumped to 5.5.0 👍

agners (approved):

PR #2830, opened 2023-01-11
Clarify explanation for require_certificate option

HelloThisIsFlo:
Hey,

I was setting up this add-on on my home-assistant, and I couldn't figure out why I couldn't connect via SSL/TLS. I enabled the debug logs, nothing was showing, just `Error: The connection was lost` on the client side 🤷‍♂️

Then I looked in the code of this repo, and tried to understand what the `require_certificate` option really did. That led me to the [documentation of mosquito](https://mosquitto.org/man/mosquitto-conf-5.html) and I read this:
> When using certificate based encryption there are three options that affect authentication. The first is require_certificate, which may be set to true or false. If false, the SSL/TLS component of the client will verify the server but there is no requirement for the client to provide anything for the server: authentication is limited to the MQTT built in username/password. If require_certificate is true, the client must provide a valid certificate in order to connect successfully.

The original explanation on this project was
> If set to `true` encryption will be enabled using the cert- and keyfile options.

This led me to believe that it had to be set to `true` for mosquito to provide an encrypted connection. But in reality, leaving it to `false` would still offer encryption, it just wouldn't authenticate the client (which tbh I'm not 100% sure what it means).

I figured I'd submit a PR to clarify that. I copy pasted the definition from the [official mosquito doc](https://mosquitto.org/man/mosquitto-conf-5.html) (with some edit to make it look nice like the other options).

I hope that helps 👍


mosquitto/DOCS.md
```diff
@@ -78,7 +78,8 @@ A file containing the private key. Place this file in the Home Assistant `ssl` f
 
 ### Option: `require_certificate`
 
-If set to `true` encryption will be enabled using the cert- and keyfile options.
+If set to `true`, the client must provide a valid certificate in order to connect successfully.  
+If set to `false`, the SSL/TLS component of the client will verify the server but there is no requirement for the client to provide anything for the server, authentication is limited to the MQTT username/password.
```
mdegat01, line 82:
```suggestion
If set to `true`, require a client certificate. Means encrypted connections are required, the client must verify and trust the server's certificate and must provide a certificate signed by the CA in `cafile`.
If set to `false`, do not require a client certificate. Can still enable encrypted connections using the `certfile` and `keyfile` options.
```
Good idea, took another stab at it though. It's more then just requiring a client certificate, this option also means only encrypted connections are possible and both parties must trust and verify each others' certificate. Also the client certificate must be signed by the CA in `cafile`. The `cafile` option doesn't do anything if `require_certificate` is false.

mdegat01 (changes requested):

FlorianKempenich:
On top of the change suggested by the conversation above, I also updated the short description of the option in `translations/en.yaml`

mosquitto/translations/en.yaml
```diff
@@ -9,7 +9,7 @@ configuration:
   require_certificate:
     name: Require Certificate
```
agners, line 10:
```suggestion
    name: Require Client Certificate
```

agners (approved):

mosquitto/DOCS.md
```diff
@@ -78,7 +78,19 @@ A file containing the private key. Place this file in the Home Assistant `ssl` f
 
 ### Option: `require_certificate`
 
-If set to `true` encryption will be enabled using the cert- and keyfile options.
+Whether this option is `true` or `false`:
+- Unencrypted connections are possible on the unencrypted ports (default: `1883`, `1884` for websockets)
+- **If `certfile` and `keyfile` are provided:** encrypted connections are possible on the encrypted ports (default: `8883`, `8884` for websockets) 
+   - In that case, the client must trust the server's certificate
+
```
agners, line 85: Explaining what `certfile` and `keyfile` is doing here is confusing. We should add this to the relevant options above.

```suggestion
```

If anything, we could just say at the end of this option that `certfile` and `keyfile` are not relevant for the `require_certificate` option, or something along these lines.

HelloThisIsFlo: I updated it, let me know if that works

agners (approved):

mosquitto/translations/en.yaml
```diff
@@ -7,9 +7,9 @@ configuration:
       don't need to do this because you can use Home Assistant users too,
       without any configuration.
   require_certificate:
-    name: Require Certificate
+    name: Require Client Certificate
     description: >-
-      If enabled encryption will be enabled using the cert- and keyfile options.
+      If enabled client will need to provide it's own certificate on top of username/password. 'cafile' must be set.
```
agners, line 12:
```suggestion
      If enabled client will need to provide it's own certificate on top of
      username/password. 'cafile' must be set.
```

mosquitto/translations/en.yaml
```diff
@@ -18,7 +18,7 @@ configuration:
   cafile:
     name: CA File
     description: >-
-      A file containing a root certificate. Place this file in the Home
+      A file containing a root certificate that signed the client certificate (Only used if 'require_certificate' enabled). Place this file in the Home
```
agners, line 21:
```suggestion
      A file containing a root certificate that signed the client certificate
      (only used if 'require_certificate' enabled). Place this file in the Home
```

agners (approved):

mosquitto/translations/en.yaml
HelloThisIsFlo: Done both 👍

HelloThisIsFlo:
Thanks for the merge 👍

PR #2834, opened 2023-01-12
silabs-multiprotocol: Make HTTP REST server HTTP compliant part 2

agners:
Explicitly set `Connection` header to `close` to tell the client that the server closes the TCP/IP connection after each request.


emontnemery (approved):
LGTM

PR #2837, opened 2023-01-15
rpc_shutdown: Allow '%' sign in password

RyanFlm:
Changed the credentials-schema to allow percent(%)-symbols in passwords.


mdegat01:
Did you test this? The reason for the restrictions on `%` is because here's how this information is actually used in this addon:
https://github.com/home-assistant/addons/blob/b734b3ca9fb25b9e051891bfbb750865f1e59ba1/rpc_shutdown/data/run.sh#L29

Here's the man page for the [net command](https://linux.die.net/man/8/net). Notice that in this doc the `-U` option is documented like this `[-U username[%password]]`. It requires a `%` symbol separating the username and password.

Because of that `%` is definitely not allowed in the username. I don't know if it has a problem with `%` in the password or not, depends on how its implemented. I can't seem to find any info on this googling and don't have an easy way to test this.

Can you show me this working fine somewhere else? Just run the following command after plugging in your info to confirm it work.
```
ADDRESS="<host address>"
USERNAME="<username>"
PASSWORD="<password with %>"
net rpc shutdown -I "$ADDRESS" -U "${USERNAME}%${PASSWORD}" -t 0
```
Please blank out any sensitive info in things you share (credentials mostly).

If doing a shutdown is too disruptive I think you can also just request info with `net rpc info`.

mdegat01 (changes requested):
Please run the requested test in the comments. Assuming that test and this change is safe to make you also need to bump the version of the addon in `config.yaml` and add an entry to the changelog.

RyanFlm:
I've run the required command in my local network and it worked fine. After the "-U" you can see the Username in front of the first '%'. After that, the password with a second '%'. The command succeeded.

prove

I found a comment on the internet where someone stated that this should be possible, but I can't find the source anymore. But I've tried it myself and it worked.

I'll make the version bump and changelog tomorrow.

Joshndroid:
can we get this looked at further and merged... I have a % in a password sadly and i can't use this pc in RPC

RyanFlm:
I've messed up the branch where I made the version changes in my fork so the commits look a bit wonky but the changes are what is requested.
- Actual change
- Version bump
- Changelog entry

rpc_shutdown/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.3
+
+- Modified the regex for password verification to allow '%' in the password
```
agners, line 5:
```suggestion
- Allow '%' sign in the password
```

agners (approved):

Issue #2783, opened 2022-12-05
Error in 2.19.1, lost all connection to Zigbee devices. Need update to 2.19.3


Brian-lub:
### Describe the issue you are experiencing

I have no connection to any of my Zigbee devices, they all droppede at after a reboot. i have tried everything i can but nothing worked. then took the mater to Discord. I have been talking to "manup" and "Mimiix" on Deconz Discord
https://discord.com/channels/494922323518947329/543511893382529064/1049323615100874843 and thay found that there is a bug in 2.19.1 and that should be fixed in 2.19.3, 
![image](https://user-images.githubusercontent.com/47833231/205736717-72a9b504-2639-4c4e-a775-7c94324794ff.png)
  

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Other (e.g., Raspbian/Raspberry Pi OS/Fedora)

### Which add-on are you reporting an issue with?

deCONZ

### What is the version of the add-on?

6.17.0

### Steps to reproduce the issue

1.It not working at all time
2.
3.
...

## System Information

version | core-2022.11.5
-- | --
installation_type | Home Assistant Supervised
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.7
os_name | Linux
os_version | 5.4.0-135-generic
arch | x86_64
timezone | Europe/Copenhagen
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.28.4
Stage | running
Available Repositories | 1156
Downloaded Repositories | 15

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | December 12, 2022 at 01:00
relayer_connected | true
remote_enabled | true
remote_connected | true
alexa_enabled | false
google_enabled | true
remote_server | eu-west-2-4.ui.nabu.casa
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Ubuntu 20.04.5 LTS
-- | --
update_channel | stable
supervisor_version | supervisor-2022.11.2
agent_version | null
docker_version | 20.10.21
disk_total | 109.5 GB
disk_used | 34.0 GB
healthy | true
supported | failed to load: Unsupported
[...]

PR #2842, opened 2023-01-18
Bump deconz to 2.20.1

kiwikern:
Update to the latest stable deconz release [v2.20.1](https://github.com/dresden-elektronik/deconz-rest-plugin/releases/tag/v2.20.1)

PR #2845, opened 2023-01-23
Bump deCONZ to 2.20.1 (stable)

Zehir:
The current stable version of deCONZ have some stability issues. See #2783.

The stable version 2.20.1 don't contain all fix but it's way better than current 2.19.3.


Zehir:
Duplicate of #2842 oops

PR #2860, opened 2023-01-29
nginx_proxy documentation update: clarified configuration.

greg-szabo:
This is a simple documentation update that clarifies that you need to restart the core service after you updated the `configuration.yaml`. This is necessary for the setting to "enable reverse proxies to access the frontend" to apply.

This is a trivial step for veteran Home Assistant users. But as a newbie, I spent days understanding Home Assistant's underlying structure (and the clever use of shared docker volumes) until I realized my trivial mistake. I hope this saves some time for others in the future.

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -17,7 +17,7 @@ The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https
    - `ssl_certificate`
    - `ssl_key`
    - `server_port`
-3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set).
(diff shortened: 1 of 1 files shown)


nginx_proxy/DOCS.md
```diff
@@ -17,7 +17,7 @@ The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https
    - `ssl_certificate`
    - `ssl_key`
    - `server_port`
-3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set).
+3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set). Restart the Home Assistant Core service after this change from `Developer Tools -> Restart`.
```
agners, line 20: The location is somewhat outdated. I'd just add the hint that a restart 

```suggestion
3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set). You'll need to restart Home Assistant Core after `configuration.yaml` changes.
```

PR #2878, opened 2023-02-15
Bump python-matter-server to 2.1.0

agners:
This allwos to drop some unnecessary Python dependencies.

### DIFF
--- a/matter_server/config.yaml
+++ b/matter_server/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 3.0.2
+version: 3.0.3
 slug: matter_server
 name: Matter Server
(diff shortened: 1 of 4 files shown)


balloob (approved):

matter_server/config.yaml
```diff
@@ -11,7 +11,7 @@ arch:
 discovery:
   - matter
 hassio_api: true
-homeassistant: 2023.2.0b0
+homeassistant: 2023.2.5
```
MartinHjelmare, line 14: Why did we change the required Home Assistant Core version? This change is a breaking change now.

agners: Because only in that version the client changes got in? This was advised by @marcelveldt

PR #2883, opened 2023-02-17
Use official Debian bullseye Python 3.10 image for matter-server

agners:


marcelveldt:
LGTM
Shall we wait with merging until beta week just in case ?

agners:
It is the same server version, so it should be safe.

I have tested it against current dev, I was able to commission a device, so should be fine.

MartinHjelmare (approved):

PR #2884, opened 2023-02-18
Add support for certbot-dns-multi

daenney:
This is a DNS plugin for certbot supporting a massive amount of different DNS providers, many of whom do not have a dedicated certbot plugin.

I have one issue with this PR that I don't know how to resolve. Due to the huge amount of providers dns-multi supports, the configuration is variable per provider. Depending on which provider you set for `multi_provider` you need to set a different set of environment variables, as documented for the provider on https://go-acme.github.io/lego/dns/.

The problem is, I don't know how to express that in the schema or make that happen. Any pointers would be hugely appreciated.


daenney:
There's an additional challenge in that Lego being written in Go, the wheels are platform-specific and currently on x86_64 appears to be uploaded by the project: https://pypi.org/project/certbot-dns-multi/4.10.0/#files

daenney:
Well that's just rude.

agners:
@daenney don't take it personally, that is just GitHub bot which closes things which don't have any traffic (and usually it means that people lost interest).

Since you seem to still care about this change, I'll reopen it.

agners:
> The problem is, I don't know how to express that in the schema or make that happen. Any pointers would be hugely appreciated.

Hm, Schema in general is optional, however, we probably don't want to loose full validation.

I am not sure, but maybe you can just invent something which isn't supported, and it won't get validated? E.g. `object`. If that doesn't work, we might consider adding support for it.

See also:
https://developers.home-assistant.io/docs/add-ons/configuration#options--schema

PR #2885, opened 2023-02-19
Mini security fix

kjasiak:
the private key was stored in a place accessible to everyone

### DIFF
--- a/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
+++ b/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
@@ -166,5 +166,7 @@ fi
 # Get the last modified cert directory and copy the cert and private key to store
 # shellcheck disable=SC2012
 CERT_DIR_LATEST="$(ls -td $CERT_DIR/live/*/ | head -1)"
+mkdir "/ssl"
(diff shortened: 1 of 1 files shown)


letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -166,5 +166,7 @@ fi
 # Get the last modified cert directory and copy the cert and private key to store
 # shellcheck disable=SC2012
 CERT_DIR_LATEST="$(ls -td $CERT_DIR/live/*/ | head -1)"
+mkdir "/ssl"
```
agners, line 169: That directory is bind mounted into the container and essentially guaranteed to exist.

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -166,5 +166,7 @@ fi
...
+chmod 600 "/ssl"
```
agners, line 170: Hm, that also means that the public key won't  be accessible. Is that really needed? The private key seems to have the proper permissions:

```
-rw-------    1 root     root        1.7K Feb 20 02:00 privkey.pem
```

Issue #2872, opened 2023-02-06
Silicon Labs Multiprotocol OTBR bind: Address already in use


mlacayoemery:
### Describe the issue you are experiencing

The OTBR appears to be in a start up loop that continuously exits with code 256 (signal 6) and then retries. Disabling the option "Enable OpenThread Border Router web interface" makes this error go away.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Silicon Labs Multiprotocol

### What is the version of the add-on?

0.13.0

### Steps to reproduce the issue

After doing the suggested fix for issue #2869 by flashing [firmware 4.2.0](https://github.com/NabuCasa/silabs-firmware/blob/main/RCPMultiPAN/NabuCasa_SkyConnect_RCP_v4.2.0_rcp-uart-hw-802154_115200.gbl), the Silicon Labs Multiprotocol seems to be getting farther along in the startup process, but is still encountering problems.

1. Start the Silicon Labs Multiprotocol addon
2. Open the addon log
3. Watch 

## System Information

version | core-2023.2.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.7
os_name | Linux
os_version | 5.15.84-v8
arch | aarch64
timezone | Europe/Zurich
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.30.1
Stage | running
Available Repositories | 1194
Downloaded Repositories | 4

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 9.5
-- | --
update_channel | stable
supervisor_version | supervisor-2023.01.1
agent_version | 1.4.1
docker_version | 20.10.22
disk_total | 57.8 GB
disk_used | 21.8 GB
healthy | true
supported | true
board | rpi3-64
supervisor_api | ok
version_api | ok
[...]

PR #2896, opened 2023-02-28
silabs-multiprotocol: Use dynamic ingress port

agners:
Fixes: #2872

### DIFF
--- a/silabs-multiprotocol/config.yaml
+++ b/silabs-multiprotocol/config.yaml
@@ -36,7 +36,6 @@ ports:
   8081/tcp: null
 ports_description:
   9999/tcp: EmberZNet EZSP/ASH port
-  8080/tcp: OpenThread Web port
   8081/tcp: OpenThread REST API port
 schema:
   device: device(subsystem=tty)?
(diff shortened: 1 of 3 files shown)


silabs-multiprotocol/translations/en.yaml
```diff
@@ -39,4 +39,5 @@ configuration:
       Use OpenThread Border Router firewall to block unnecessary traffic.
 network:
   9999/tcp: EmberZNet EZSP/ASH port
+  8080/tcp: OpenThread Web port
```
frenck, line 42: Seems to be unused?

silabs-multiprotocol/config.yaml
```diff
@@ -36,6 +36,7 @@ ports:
   8081/tcp: null
 ports_description:
   9999/tcp: EmberZNet EZSP/ASH port
+  8080/tcp: OpenThread Web port
```
frenck, line 39: Seems to be unused?

frenck (changes requested):

silabs-multiprotocol/translations/en.yaml
agners: Right, I don't need that when allocating automatically. Will remove those.

frenck (approved):

agners:
Since the Web interface requires the REST API as well (which doesn't go through ingress) we decided to remove ingress completely.

Instead we plan to start the web interface when the relevant ports (Web UI/REST API) are exposed externally.

PR #2898, opened 2023-03-01
silabs-multiprotocol: Update vendor name/product name 

agners:
Use "Home Assistant" with space since other vendors use vendor names
with spaces as well. Choose a product name where we can distinguish
which OTBR add-on is used exactly.

### DIFF
--- a/silabs-multiprotocol/CHANGELOG.md
+++ b/silabs-multiprotocol/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.0.0
+
+- Remove Web UI via ingress (expose ports to use the Web UI, see documentation)
+- Change vendor name to "Home Assistant" and product name to Silicon Labs
+  Multiprotocol" (used in OTBR mDNS/DNS-SD announcments)
+
 ## 0.13.1
 
 - Set default baudrate 115200 correctly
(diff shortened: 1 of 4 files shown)


agners:
Update in Core to detect `Home Assistant` with space as HA brand: https://github.com/home-assistant/core/pull/88976

balloob (approved):

agners:
Instead of calling `./script/setup` this calls directly `./script/cmake-build`. This bypasses a lot of setup/initialization scripts. However, quite some did not do anything because features have been disabled, or did not make any sense in our environment (container).

Here is a summary of the install functions which are called (see [scripts/ dir of the ot-br-posix repo](https://github.com/openthread/ot-br-posix/tree/d83fee189a5d55e10a8d1494f1600c23f4f2df80/script))

* firewall_install: we use our own s6 service
* ipforward_install: is configured on OS level (see https://github.com/home-assistant/operating-system/pull/1832)
* rt_tables_install: address in `Dockerfile` and on OS level (see https://github.com/home-assistant/operating-system/pull/2375)
* nat64_install: defaults to off for debian platform and not enabled by us
* dns64_install: defaults to off for debian platform and not enabled by us
* network_manager_install: defaults to off for debian platform and not enabled by us
* dhcpv6_pd_install: defaults to off for debian platform and not enabled by us
*  border_routing_install: dhcpcd not used, Route Information Option handled by NetworkManager (under investigation, see https://github.com/home-assistant/operating-system/discussions/2333)
* otbr_install: using final configuration directly, build and install is done in `Dockerfile` directly

PR #2908, opened 2023-03-08
silabs-multiprotocol: Use host namespace for hostname

agners:
This makes sure that the BR is announced with the systems real hostname.


agners:
With https://github.com/home-assistant/supervisor/pull/3596 deployed on the stable channel, we can safely use this new option now.

MattWestb:
I think this is very needed that having more HA OTBR up and running then they is fighting for have the name and trying other for finding one free one and its not good for the system having this kind of behavior.

mdegat01 (approved):

Issue #2912, opened 2023-03-13
Nginx proxy addon: No config option for `X-Forwarded-Host`


TheHolyRoger:
### Describe the issue you are experiencing

As the title says, there's no config option for `X-Forwarded-Host` which should go just here: https://github.com/home-assistant/addons/blob/master/nginx_proxy/data/nginx.conf#L66

As noted in the vscode issue here: https://github.com/hassio-addons/addon-vscode/issues/584#issuecomment-1462976192

The latest version of vscode no longer functions without adding:
`proxy_set_header X-Forwarded-Host $http_host;`
To the nginx config

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

NGINX Home Assistant SSL proxy

### What is the version of the add-on?

3.2.0

### Steps to reproduce the issue

1. Use nginx proxy
2.  Use latest version of vscode addon

### System Health information

N/A

### Anything in the Supervisor logs that might be useful for us?

```txt
N/A
```

### Anything in the add-on logs that might be useful for us?

```txt
N/A
```

agners:
Fixed by: #2915.

PR #2913, opened 2023-03-14
NGINX: Reload NGINX on certificate renewal

lmagyar:
fixes #2870 #2725 #2600

This modification checks daily (crond default daily 02:00:00) whether the configured certificate file has been changed and reloads (not restarts) nginx to pick up the new certificate.
- If there is no backup connection to the device (eg. ssh, cloud, vpn), we can lose access to it. In case of a remote, not regularly updated device, this can be a real problem.
- DuckDNS add-on checks the certificate in each 12h.
- Letsencrypt renews the certificate if it expires within the next 30 days and did not revoke the previous certificate.
- So checking certificate renewal daily should be fine.

It also uses the new s6-overlay layout.

```
s6-rc: info: service nginx: starting
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service nginx successfully started
s6-rc: info: service crond: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service crond successfully started
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
s6-rc: info: service legacy-cont-init successfully started
s6-rc: info: service legacy-services: starting
s6-rc: info: service legacy-services successfully started
[21:09:33] INFO: Running nginx...
[02:00:00] INFO: Reloading nginx to reload new certificate file
```

PR #2915, opened 2023-03-15
Nginx: Add X-Forwarded-Host to fix vscode origin issues

TheHolyRoger:
Fixes:
https://github.com/home-assistant/addons/issues/2912
https://github.com/hassio-addons/addon-vscode/issues/584

### DIFF
--- a/nginx_proxy/data/nginx.conf
+++ b/nginx_proxy/data/nginx.conf
@@ -66,6 +66,7 @@ http {
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection $connection_upgrade;
+            proxy_set_header X-Forwarded-Host $http_host;
         }
     }
 


rokam (approved):

vdemidov:
What are we waiting for?

thor0215 (approved):
This looks good to me.

thor0215:
@frenck What else must be done before this change can be merged into the master branch?

agners:
@TheHolyRoger I've just tested and merged #2913 which was an older PR. Can you rebase yours to current master again?

Also, please bump the version number and add an entry to the `CHANGELOG.md`.

TheHolyRoger:
@agners done sir

agners:
@TheHolyRoger I've just installed the Visual Studio add-on and access it through Nginx (using TLS), things seem to work normally. Did maybe something else change which fixes the problem #2912 is describing? Or is there a particular feature of VS Code Server which doesn't work without this header?

TheHolyRoger:
@agners there's heaps of people experiencing the same issue on here and on the hass forums. It also affects some other addons although I forget which.

`X-Forwarded-Host` is also one of the standard headers which should be passed when proxying: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

thor0215:
@agners v4.10.1 of code-server now requires the X-Forwarded-Host entry

> https://github.com/coder/code-server/releases/tag/v4.10.1
> 
> Code v1.75.1
> Security
> 
> Added an origin check to web sockets to prevent cross-site hijacking attacks on
> users using older or niche browser that do not support SameSite cookies and
> attacks across sub-domains that share the same root domain.
> 
> The check requires the host header to be set so if you use a reverse proxy
> ensure it forwards that information otherwise web sockets will be blocked.

TheHolyRoger:
@agners thank you very much sir. I disabled my workaround, rebooted hass and confirmed vscode still did NOT load on the previous version of the nginx addon (3.2.0)

Then updated to this release (3.4.0) and all is fixed. 👍

PR #2925, opened 2023-03-21
Replace Almond with Genie

wrt54g:
Replaces Almond with Genie in the readme file.


frenck (approved):
Thanks, @wrt54g 👍 

../Frenck

PR #2931, opened 2023-03-22
Add URLs to Z-Wave JS changelog

wrt54g:
Adds URLs for all releases to the Z-Wave JS changelog.
The Z-Wave JS server release 2.0.0 was not available in the z-wave JS server repository, so that one is not added.


MartinHjelmare (approved):
Thanks!

PR #2933, opened 2023-03-24
Bump zwave-js-server to 1.27.0 and zwave-js to 0.1.77

raman325:
One thing I changed in the changelog this time was including changelogs for versions in between the current version of zwave-js and the one from the last release. For completeness!

My deviceless network came up clean


frenck (approved):
Thanks, @raman325 👍 

../Frenck

PR #2941, opened 2023-03-27
configurator: Use correct wheels path

agners:


frenck (changes requested):
Let's keep the changelog up to date

frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #2950, opened 2023-03-30
silabs-multiprotocol: Use firmware with baudrate of 460800

agners:


agners:
Currently, when updating users have to manually change the add-ons baud rate. This likely will lead to many broken systems.

puddly (approved):

MattWestb:
Then the flasher is having the information and after have flashing one new firmware is it not good implanting function for asking the user if it shall updating the config for the addon if the baudrate is changed from the saved one  `yes / no` ?

agners:
We don't really have a way to interact with the user or an upgrade path for settings.

Why would a user even say "no"? We know we flashed a firmware with 460800. A "no" will just not work...

MattWestb:
I was thinking of updating on one other machine if having more and moving the device or having mode dongles with different speed.
But you have right its not for normal users its for advance ones.

And i think its better you look on the new issue with this PR https://github.com/home-assistant/addons/issues/2967

Thanks for feedback !

agners:
> I was thinking of updating on one other machine if having more and moving the device or having mode dongles with different speed.

It still will respect your setting if the firmware update doesn't apply to your hardware. Also, you can always disable the firmware update, then the configured baudrate will be used *always*.

PR #2919, opened 2023-03-17
Replace deprecated terms and update GitHub URLs

wrt54g:
Replaces deprecated terms and updates the GitHub URLs from "hassio-addons" to "addons".

PR #2954, opened 2023-03-30
Fix small link issue

wrt54g:
Fix issue related to #2919

### DIFF
--- a/configurator/DOCS.md
+++ b/configurator/DOCS.md
@@ -73,4 +73,3 @@ In case you've found a bug, please [open an issue on our GitHub][issue].
 [forum]: https://community.home-assistant.io
 [issue]: https://github.com/home-assistant/addons/issues
 [reddit]: https://reddit.com/r/homeassistant
-[repository]: https://github.com/hassio-addons/repository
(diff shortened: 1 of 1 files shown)


configurator/DOCS.md
```diff
@@ -71,6 +71,6 @@ In case you've found a bug, please [open an issue on our GitHub][issue].
 
 [discord]: https://discord.gg/c5DvZ4e
 [forum]: https://community.home-assistant.io
-[issue]: https://github.com/home-assistant/hassio-addons/issues
+[issue]: https://github.com/home-assistant/addons/issues
 [reddit]: https://reddit.com/r/homeassistant
-[repository]: https://github.com/addons/repository
+[repository]: https://github.com/hassio-addons/repository
```
agners, line 76: That seems wrong

wrt54g: @agners Why?

wrt54g: This fixes an issue with the larger PR #2919.

agners: Hm I see. But that URL is not really used here, we don't point to the community add-ons.

Just drop it then.

```suggestion
```

frenck (approved):
Thanks! 👍 

../Frenck

PR #2960, opened 2023-03-31
Dnsmasq: Added switch that disables dnsmasq "log-queries" parameter

odya:
### Problem:
Dnsmasq floods log with dns requests.

### Solution:
Added switch that disables `log-queries` parameter so log file became much cleaner.

### DIFF
--- a/dnsmasq/translations/en.yaml
+++ b/dnsmasq/translations/en.yaml
@@ -20,7 +20,7 @@ configuration:
     description: Provide cname records for DNSMasq to resolve.
   log_queries:
     name: Log Queries
-    description: Log all DNS requests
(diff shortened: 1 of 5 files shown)


agners (approved):

dnsmasq/translations/en.yaml
```diff
@@ -18,6 +18,9 @@ configuration:
   cnames:
     name: Cnames
     description: Provide cname records for DNSMasq to resolve.
+  log_queries:
+    name: Log Queries
+    description: Log all DNS requests
```
agners, line 23:
```suggestion
    description: Log all DNS requests.
```

PR #2964, opened 2023-04-02
DuckDNS: Remove support for DNS aliases

lildude:
Automatically renewing Let's Encrypt certificates with the DuckDNS addon when aliases are configured has been broken for quite some time with https://github.com/home-assistant/addons/issues/2505 being the most recently opened issue.

As I've detailed in the issue https://github.com/home-assistant/addons/issues/2505#issuecomment-1242714582:

> The issue here is dehydrated that is used for getting/renewing the certificates deploys the challenge tokens for all the domains and then performs the validation for each domain.
>
> This causes a problem with DuckDNS as it only has a single TXT record which will always be overwritten by the challenge for the last domain in the list.

I attempted to fix this in https://github.com/home-assistant/addons/pull/2662 but this wasn't optimal and it was decided the big mistake was in adding support for aliases in the first place and as such the best fix would be to remove support for aliases. See @mdegat01's comment at https://github.com/home-assistant/addons/pull/2662#discussion_r1084602655

This PR does just that. I've also added a section to the docs detailing how to use the Let's Encrypt addon to obtain certificates for DNS aliases.

I've also updated the docs and reordered the options to match the configuration.

Fixes https://github.com/home-assistant/addons/issues/2505

/cc @mdegat01


richardwonka:
Please leave aliases in. 

My setup relies heavily on the device being in my own domain.

lildude:
@richardwonka Perfect!!! You should be able to use the Let's Encrypt addon without any problems.

You've already got a CNAME record for your own domain pointing to your `duckdns.org` subdomain which in turn points to your regularily updated IP address (thanks to this addon) so you only need to configure the Let's Encrypt addon to generate the certificates for your own domain.

If your DNS provider is on the supported list, it's really easy, and you can use the `dns` challenge. If it's not, you'll need to use the `http` challenge and temporarily allow access on port 80 whenever a renew is performed, which is manually by default, but can be [automated](https://community.home-assistant.io/t/blueprint-for-automatic-renewal-of-a-lets-encrypt-certificate/300533).

jonathanebetts:
If domain alias support is going to be removed from DuckDNS, is there a guide/documentation on how to use DuckDNS alongside Let's Encrypt to set up multiple subdomains which all point to the same duckdns domain @lildude ?

lildude:
> is there a guide/documentation on how to use DuckDNS alongside Let's Encrypt to set up multiple subdomains which all point to the same duckdns domain @lildude ?

There is now 😁 I've just added a section to this addon's docs in this PR.

richardwonka:
> @richardwonka Perfect!!! You should be able to use the Let's Encrypt addon without any problems.
> 
> You've already got a CNAME record for your own domain pointing to your `duckdns.org` subdomain which in turn points to your regularily updated IP address (thanks to this addon) so you only need to configure the Let's Encrypt addon to generate the certificates for your own domain.

_BUT_ I cannot configure the domain to point to my own domain name. the addon complains that domain names need to match `*.duckdns.org`. 

Once that discrepancy is fixed, this would work for me, but it seems that aliases are not such an odd use case to just drop them.

lildude:
> _BUT_ I cannot configure the addon to point to my own domain name. the addon complains that domain names need to match `*.duckdns.org`.
>
> Once that discrepancy is fixed, this would work for me, but it seems that aliases are not such an odd use case to just drop them.

That's expected and it won't and shouldn't change. All the alias function does is add the custom domain to the certificate. This is now handled by the Let's Encrypt addon.

In short:

- The DuckDNS addon updates the IP address for your instance on duckdns.org.
- You will already have a CNAME DNS record pointing your own domain to your DuckDNS domain, which will effectively be resolving your domain to your HA IP address.
- You then configure the Let's Encrypt addon to obtain the certificate for your domain.

Please try following the details I'm adding to this PR (see rendered [here](https://github.com/home-assistant/addons/blob/134fe8a9ff3ccb7ab2e25257d517282ba9f930ee/duckdns/DOCS.md#using-duckdns-and-lets-encrypt-addons-to-get-certificates-for-your-own-domains)) and let me know how it goes.

arbo-milestone:
_TL;DR: You might be able to automate listening to port 80 from home assistant (and the computer it's running on) itself, but as soon as you have other firewalls and security gateways in front of your home assistant instance this becomes a mess_

Just wanted to let you know this sounds like a nightmare :)

I'm one of those people who use duckdns with my own domain which don't use one of the around 20 supported dns providers in the let's encrypt add-on.

Currently I need to go into my duckdns configuration every 3 months, change the alias configuration, restart it, change it back and everything works. Elegant? no, but it's what I need to do.

However, if you remove this some people are going to be really sad since the let's encrypt add-on it not really an option.
Not using one of the supported DNS providers requires the use of http, but this comes with a range of other problems.

The entire reason for using DNS is not to expose a webserver on 80 to the internet in the first place.
You might be able to automate listening to port 80 from home assistant (and the computer it's running on) itself, but as soon as you have other firewalls and security gateways in front of your home assistant instance this becomes a mess

As I can see it this leaves me with the following choices:

- Switch DNS provider
- Open up for unencrypted traffic on port 80 into my network
- Turn off autoupdate on the duckDNS add-on and live with the current solution

I think I'm going with the last option, but I don't like it :)

I have not tried to understand the fundamental problem you are fighting with in the duckDNS add-on, but it feels like the problem should be fixed there. Good luck :)

richardwonka:
> @richardwonka Perfect!!! You should be able to use the Let's Encrypt addon without any problems.
Other than having to automate something for which an automated process is already exists (in the months-old pull request the duckdns add-on) and which has no practical downsides.
 
> [...]
> 
> If your DNS provider is on the supported list, it's really easy, and you can use the `dns` challenge. If it's not, you'll need to use the `http` challenge and temporarily allow access on port 80 whenever a renew is performed, which is manually by default, but can be [automated](https://community.home-assistant.io/t/blueprint-for-automatic-renewal-of-a-lets-encrypt-certificate/300533).

Opening port 80 *and* requiring manual interaction instead of merging an existing automated solution is _far from perfect_.

Neither of the available solutions may be _perfect_, but each of them works for _some_ users and both make sense to be available.

MikeGuest:
OK, I see this, however,

How do I configure the LetsEncrypt addon if i want a certificate that's valid for BOTH mydomain.duckdns.org and Mydomain.com?

If i'm specifying a DNS supplier, that will provide details for my domain, but won't have any details for the duckdns domain

I'm probably confused, but all I want is a cert with the 2 domains as a SAN, i don't really care whether it comes from the duckdns add on or the LetsEncrypt add-on.

I've tried reading the docs, but either i'm not understanding them properly or they don't cover this

iskiselev:
I've shared an idea in #2662 how it potentially can be fixed. Unfortunately I don't know enough about project structure to fully test it - but probably someone here would like to take a look and give alias one more chance before deleting them.

PR #2968, opened 2023-04-05
silabs-multiprotocol: Fix firmware flashing for amd64

puddly:
The newly-introduced `UNIVERSAL_SILABS_FLASHER` argument wasn't used in `Dockerfile.amd64` so an old version of the flasher was still being installed, breaking startup. No other platform was affected.


agners (approved):
Good catch, thanks for the find!

PR #2977, opened 2023-04-15
Update letsencrypt Addon

snow2k9:
Hey,
got problems with DNS Challenge and INWX, so tried it locally and compared versions. Seems like updating certbot and acme solved my problems.
- Updated base image to newest version
- Updated all dns-plugins to newest versions
- Escaped passwords with """password""", found it in certbot docs
- added german translation

Tested with INWX
I don't use any other provider but checked the docs for changes


XantXant:
Hi,
i have made another pull request for the password problem (only for INWX).
https://github.com/home-assistant/addons/pull/2853
And corrected my mistake in the documentation for INWX.
I think my pull request would be obsolete (only documentation fix would be nice).

Issue #2742, opened 2022-11-10
There is already a MySQL service in use from core_mariadb


magiva:
### Describe the issue you are experiencing

since updating to 2022.11 mariadb wont start.
it states an error that there is already a mysql db running for core-mariadb.
[22:56:27] INFO: Ensuring internal database upgrades are performed
[22:56:27] INFO: Ensure databases exists
[22:56:28] INFO: Create database homeassistant
[22:56:28] INFO: Ensure users exists and are updated
[22:56:30] INFO: Update user homeassistant
[22:56:31] INFO: Init/Update rights
[22:56:32] INFO: Granting all privileges to homeassistant on homeassistant
[22:56:36] ERROR: Got unexpected response from the API: There is already a MySQL service in use from core_mariadb
[22:56:36] INFO: Successfully send service information to Home Assistant.

this does not recover from a restore either, some blogs note that the mariadb should be stopped before snapshot.
these suggestions to stop, snapshot, restore, start.... were followed but the problem remains.
uninstalling and re-installing did not fix the issue.

ive had to move the record back to stock method to get that working but i also run other tables in the mariadb that nodered was talking to.

this was also originally noted here
https://github.com/home-assistant/addons/issues/2131

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Debian

### Which add-on are you reporting an issue with?

MariaDB

### What is the version of the add-on?

2.5.1

### Steps to reproduce the issue

start addon.... doesnt start

## System Information

version | core-2022.11.2
-- | --
installation_type | Home Assistant Supervised
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.7
os_name | Linux
os_version | 5.10.103-v8+
arch | aarch64
timezone | Europe/Oslo
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4541
Installed Version | 1.28.3
Stage | waiting
Available Repositories | 1148
Downloaded Repositories | 75

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | 22 November 2022 at 01:00
relayer_connected | true
remote_enabled | true
remote_connected | true
alexa_enabled | false
google_enabled | true
remote_server | eu-west-2-4.ui.nabu.casa
can_reach_cert_server | ok
[...]

Issue #2906, opened 2023-03-08
MariaDB: "There is already a MySQL service in use from core_mariadb"


kongjudas:
### Describe the issue you are experiencing

All entity history and Energy data is lost and the database is reset after upgrade. There are no errors, except:
`There is already a MySQL service in use from core_mariadb`. So, there seems to be a duplicate process running somewhere, with a new, clean database.

I have a long history, back to 2021 and can't upgrade until this is fixed. I've tried several routes to get around this, but no cigar.

I know this is a duplicate, but the other entries are ignored by devs.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

MariaDB

### What is the version of the add-on?

2.5.2

### Steps to reproduce the issue

1. Upgrade from 2023.5 to 2023.3.*
2. Check Energy and any other history/ statistic

### System Health information

Nothing to report

### Anything in the Supervisor logs that might be useful for us?

```txt
There is already a MySQL service in use from core_mariadb
```

### Anything in the add-on logs that might be useful for us?

2023-03-09 16:04:16 0 [Note] InnoDB: File './ibtmp1' size is now 12 MB.
2023-03-09 16:04:16 0 [Note] InnoDB: 10.6.10 started; log sequence number 178137603563; transaction id 6836767
2023-03-09 16:04:16 0 [Note] InnoDB: Loading buffer pool(s) from /data/databases/ib_buffer_pool
2023-03-09 16:04:16 0 [Note] Plugin 'FEEDBACK' is disabled.
2023-03-09 16:04:17 0 [Note] InnoDB: Buffer pool(s) load completed at 230309 16:04:17
2023-03-09 16:04:17 0 [Note] Server socket created on IP: '0.0.0.0'.
2023-03-09 16:04:17 0 [Note] Server socket created on IP: '::'.
2023-03-09 16:04:17 0 [Note] /usr/bin/mariadbd: ready for connections.
Version: '10.6.10-MariaDB'  socket: '/run/mysqld/mysqld.sock'  port: 3306  MariaDB Server
[16:04:17] INFO: Check data integrity and fix corruptions
mysql.column_stats                                 OK
mysql.columns_priv                                 OK
mysql.db                                           OK
mysql.event                                        OK
mysql.func                                         OK
mysql.global_priv                                  OK
mysql.gtid_slave_pos                               OK
mysql.help_category                                OK
[...]

Issue #2131, opened 2021-07-28
There is already a MySQL service in use from core_mariadb


kongo09:
## The problem

The MariaDB log started to report a problem:
```txt
[18:21:12] ERROR: Got unexpected response from the API: Unknown error, see supervisor
```
The supervisor log says:
```txt
21-07-28 18:21:12 ERROR (MainThread) [supervisor.services.modules.mysql] There is already a MySQL service in use from core_mariadb
```
The system, however, seems to work normally.

## Environment

- Add-on with the issue: MariaDB
- Add-on release with the issue: 2.4
- Last working add-on release (if known): 
- Operating environment (OS/Supervised): supervised

## Problem-relevant configuration

```yaml
databases:
  - homeassistant
logins:
  - username: homeassistant
    password: xxx
rights:
  - username: homeassistant
    database: homeassistant
```

## Traceback/Error logs

Full MariaDB log:
```txt
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
[18:21:09] INFO: Using existing mariadb initial system
[18:21:09] INFO: Starting MariaDB
210728 18:21:09 mysqld_safe Logging to '/data/databases/mariadb.err'.
210728 18:21:10 mysqld_safe Starting mysqld daemon with databases from /data/databases
210728 18:21:10 mysqld_safe Starting mysqld daemon with databases from /data/databases
2021-07-28 18:21:10 0 [Note] /usr/bin/mysqld (mysqld 10.4.19-MariaDB) starting as process 397 ...
2021-07-28 18:21:10 0 [Note] mysqld: Aria engine: starting recovery
recovered pages: 0% 32% 51% 80% 100% (0.0 seconds); tables to flush: 2 1 0
 (0.0 seconds); 
2021-07-28 18:21:10 0 [Note] mysqld: Aria engine: recovery done
2021-07-28 18:21:10 0 [Note] InnoDB: Using Linux native AIO
2021-07-28 18:21:10 0 [Note] InnoDB: Mutexes and rw_locks use GCC atomic builtins
2021-07-28 18:21:10 0 [Note] InnoDB: Uses event mutexes
[...]

PR #2983, opened 2023-04-18
MariaDB: Migrate add-on layout to S6 Overlay

lmagyar:
Fixes: #2742 #2906 #2131

The current MariaDB add-on source is incompatible with the new S6-overlay version.

The `/mariadb/rootfs/etc/services.d/mariadb/run` has a bug:
```
# Register stop
function stop_mariadb() {
    bashio::services.delete "mysql"
    mysqladmin shutdown
    # Successful exit, avoid wait exit status to propagate
    exit 0
}
trap "stop_mariadb" SIGTERM SIGHUP

wait "${MARIADB_PID}"
```

This trap is never executed. I've run into this problem in my own database add-on, contacted S6 developers, and the current "run script is way outside the specs for a supervised service", see: https://github.com/just-containers/s6-overlay/issues/457

The solution is to rewrite this add-on:
- use new S6 folder structure
- cut the mariadb/run.sh script into 3 services (pre, core and post activities)
- cut the mariadb-lock/run.sh script into 2 services (core and post activities)
- forget `mysqld_safe` also (this has another `trap` inside)
- modify a bit the startup logic


lmagyar:
Rebased and removed draft flag.

Though nobody who originally reported the issue tested this fork in live environment. So this is still in "works on my test config" state.

mariadb/config.yaml
```diff
@@ -10,8 +10,8 @@ arch:
   - aarch64
   - amd64
   - i386
-backup_post: s6-svc -wd -d /run/s6/legacy-services/mariadb-lock/
-backup_pre: s6-svc -wU -u /run/s6/legacy-services/mariadb-lock/
+backup_post: /command/s6-rc -d change mariadb-lock
+backup_pre: /command/s6-rc -u change mariadb-lock
```
agners, line 14: Shouldn't this be

```suggestion
backup_post: /command/s6-rc -d change mariadb-lock-post
backup_pre: /command/s6-rc -u change mariadb-lock-post
```

Or is that some s6 naming magic?

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-lock-post/finish
```diff
@@ -0,0 +1,21 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+# ==============================================================================
+# Unlock tables locked for backups
+# ==============================================================================
+
+# Due to it is started by the supervisor calling the backup_post: `/command/s6-rc -d change mariadb-lock` command,
+# it's stdout won't be redirected automatically to the add-on's log
```
agners, line 8: So this enables logging of the locking, nice! :+1:

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-lock-post/run
```diff
@@ -0,0 +1,21 @@
...
+MARIADB_LOCK_CORE_PID=$(s6-svstat -o pid "/run/service/mariadb-lock-core")
+if bashio::var.equals ${MARIADB_LOCK_CORE_PID} -1; then
+    bashio::log.error "Invalid mariadb-lock-core process id ${MARIADB_LOCK_CORE_PID}"
+    bashio::exit.nok
+fi
+
+# File descriptor &4 is used as stdin for mysql, because &0 is closed after the mariadb-lock-core service is started
+echo "FLUSH TABLES WITH READ LOCK;" > /proc/${MARIADB_LOCK_CORE_PID}/fd/4
+
+bashio::log.info "MariaDB tables locked"
```
agners, line 21: The whole locking/unlocking got a quite a bit more complex, but it seems solid :+1: .

agners (commented):
I'll give this a try here, and if successful, I'll am planning on merging it.

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-pre/run
```diff
@@ -0,0 +1,28 @@
...
+printf "${NEW_INSTALL}" > /var/run/s6/container_environment/NEW_INSTALL
+
+# Redirect log output
+mkdir -p "${MARIADB_DATA}"
+rm -f "${MARIADB_DATA}/mariadb.err"
+ln -s /proc/1/fd/1 "${MARIADB_DATA}/mariadb.err"
+
+# Init mariadb
+if bashio::var.true "${NEW_INSTALL}"; then
+    bashio::log.info "Create a new mariadb initial system"
```
agners, line 24:
```suggestion
    bashio::log.info "Create a new mariadb"
```

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-pre/run
```diff
@@ -0,0 +1,28 @@
...
+    mysql_install_db --user=root --datadir="${MARIADB_DATA}" --skip-name-resolve --skip-test-db > /dev/null
+else
+    bashio::log.info "Using existing mariadb initial system"
```
agners, line 27:
```suggestion
    bashio::log.info "Using existing mariadb"
```

mariadb/config.yaml
lmagyar: Yeah, magic. I've spent a few hours figuring out how to do this. This is a working solution, seems to be right, but I'm not a professional developer, so I don't have enough experience to say, that "this is the correct solution". :/

So, `mariadb-lock` is a bundle, that consist of longrun `mariadb-lock-core` and oneshot `mariadb-lock-post`, we `up` and `down` the bundle and S6 manages the details. The `mariadb-lock` bundle is not part of the `user` bundle, so it is not started when the add-on starts.

I've experimented with a single service, but there were always some problems, this longrun+oneshot combination seems to be right.

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-post/run
```diff
@@ -1,32 +1,26 @@
...
+}
 
 # Wait until DB is running
 while ! mysql -e "" 2> /dev/null; do
     sleep 1
 done
+bashio::log.info "MariaDB started"
+
+bashio::log.info "Deleting service information from Home Assistant"
+execute_without_error_messages bashio::services.delete "mysql" || true
```
agners, line 23: Is this required with this changes in? From what I understand the finish script is always called, so this shouldn't be necessary, unless for some reason the API wasn't available on shutdown but is again now?

I guess it also doesn't hurt so :man_shrugging:

mariadb/config.yaml
agners: I see and it seems to work well here:

```
[11:24:05] INFO: Start MariaDB client (to lock tables for backups)
[11:24:05] INFO: MariaDB tables locked
[11:24:07] INFO: MariaDB tables unlocked
[11:24:07] INFO: MariaDB client (to lock tables for backups) stopped
```

Do you happen to have a pointer where this bundle feature is documented?

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-lock-post/finish
lmagyar: The strange thing is, that the longrun `mariadb-lock-core` service emits the log correctly, only this oneshot `mariadb-lock-post` service emits the log to the "wrong" place. Both of them are part of the same `mariadb-lock` bundle, and both of them started by the same command line (`/command/s6-rc -u change mariadb-lock`), but only the oneshot behaves this strange way.

During add-on start, any oneshot service that is part of the user bundle emits the log to the correct place, only oneshot services started manually behave this strange way.

So this smells to me, and I'm not sure this is the correct way, but maybe nobody at S6 thought these strange use-cases.

lmagyar: Or maybe the supervisor, who executes these `backup_pre` and `backup_post` commands should redirect any output from these commands to the add-on's log. I can't remember whether the output (without the redirection) was part of the supervisor logs. :(

agners: > During add-on start, any oneshot service that is part of the user bundle emits the log to the correct place, only oneshot services started manually behave this strange way.

Hm, maybe oneshot inherit the stdout somehow from the process which initializes it. On add-on start, that is init  I guess, whereas when starting from command line, it is something else :thinking: 

Anyhow, it seems to work with this fiix :+1: 

> Or maybe the supervisor, who executes these backup_pre and backup_post commands should redirect any output from these commands to the add-on's log.

Yeah maybe we should consider that anyhow.

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-post/run
lmagyar: This is just a failsafe trick, this is for the possibility, that the finish script wasn't called, and that can be caused eg. by a manual `docker kill` command.

lmagyar: Maybe remove the log message, just do it silently, no harm, and no questions about the log message.

agners: Yes, let's do that.

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-post/run
```diff
@@ -1,32 +1,26 @@
...
+    return ${exit_code}
+}
 
 # Wait until DB is running
 while ! mysql -e "" 2> /dev/null; do
     sleep 1
 done
+bashio::log.info "MariaDB started"
+
+bashio::log.info "Deleting service information from Home Assistant"
```
agners, line 22:
```suggestion
# Delete service information again, just in case the previous instance somehow did not cleanup
```

Issue #2426, opened 2022-04-05
ssh: ttyd crashing with SIGSEGV


agners:
### Describe the issue you are experiencing

It seems that ttyd (part of the SSH add-on) is crashing every now and then:

```
coredumpctl
TIME                           PID UID GID SIG     COREFILE EXE           SIZE
Fri 2022-04-01 13:57:28 UTC  12895   0   0 SIGSEGV none     /usr/bin/ttyd  n/a
Fri 2022-04-01 17:20:48 UTC  13274   0   0 SIGSEGV none     /usr/bin/ttyd  n/a
Mon 2022-04-04 11:29:38 UTC  15490   0   0 SIGSEGV none     /usr/bin/ttyd  n/a
```

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Terminal & SSH

### What is the version of the add-on?

9.3.0

### Steps to reproduce the issue

1. Install the add-on
2. Let the system run for multiple days
3. Observe SIGSEGV in `coredumpctl`

It seems that aarch64 systems are affected.

### Additional information

From journalctl

```
Apr 04 11:29:17 ha-prod homeassistant/aarch64-addon-ssh:9.3.0/addon_core_ssh[429]: [2022/04/04 13:29:17:1154] N:  ++ [wsisrv|0|adopted] (1)
Apr 04 11:29:17 ha-prod homeassistant/aarch64-addon-ssh:9.3.0/addon_core_ssh[429]: [2022/04/04 13:29:17:1178] N: HTTP / - 172.30.32.2
Apr 04 11:29:17 ha-prod homeassistant/aarch64-addon-ssh:9.3.0/addon_core_ssh[429]: [2022/04/04 13:29:17:2805] N: HTTP /token - 172.30.32.2
Apr 04 11:29:17 ha-prod homeassistant/aarch64-addon-ssh:9.3.0/addon_core_ssh[429]: [2022/04/04 13:29:17:3412] N: WS   /ws - 172.30.32.2, clients: 1
Apr 04 11:29:17 ha-prod homeassistant/aarch64-addon-ssh:9.3.0/addon_core_ssh[429]: [2022/04/04 13:29:17:3458] N: started process, pid: 598
[...]

PR #2993, opened 2023-04-20
ssh: Update HA CLI and other add-on dependencies

agners:
Fixes: #2426


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3011, opened 2023-04-27
Add hability to override NGINX default_server

wolfwander:
This pull request adds the necessary code to provide the hability to override the NGINX default_server .
The default_server is declared in the separated file [default.conf](https://github.com/wolfwander/addons/blob/master/nginx_proxy/rootfs/etc/default.conf).
The [nginx.conf](https://github.com/wolfwander/addons/blob/master/nginx_proxy/rootfs/etc/nginx.conf) has an include directive for including the [default.conf](https://github.com/wolfwander/addons/blob/master/nginx_proxy/rootfs/etc/default.conf) file.

The s6 [run](https://github.com/wolfwander/addons/blob/master/nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/nginx/run) tests if there is a file with default_server directive inside /share/$CUSTOMIZE_SERVERS and delete the include directive from [nginx.conf](https://github.com/wolfwander/addons/blob/master/nginx_proxy/rootfs/etc/nginx.conf) in positive case.

With the hability to override the NGINX default_server, the NGINX Proxy could act as recerse proxy for unamed HTTP hosts, like Home Assistant Emulated Hue.


wolfwander:
Any news about this pull request?

mdegat01:
That's not the point of this addon. This addon is supposed to be an extremely simple and easy configure way to get Home Assistant running with SSL for external access while still running with SSL for internal access. It's not supposed to be insert-your-own config NGinx. The insert-your-config options already present are probably a bit of a mistake to be honest but we definitely don't want to expand the scope of them.

If you want a generic reverse proxy usable for other self-hosted services I would advise you to look into the Nginx Proxy Manager add-on in the community add-on repo. Or configure and host your own Nginx, Caddy or reverse proxy tool of your choice.

Issue #3013, opened 2023-04-27
Crond logs in Home Assistant CLI after nginx update


janrg:
### Describe the issue you are experiencing

About 1-2 minutes after the “Welcome to the Home Assistant command line” message appears, the following message shows up:

`crond[58] crond (busybox 1.35.0) started, log level 8`

This is then followed on every cron execution with a message such as this one:

`crond[58] USER root pid 148 cmd run-parts /etc/periodic/15min`

A downgrade of the nginx addon to 3.2 appears to get rid of these messages. From the commit history, it also seems likely that it was introduced in 3.3.

More information here:
https://community.home-assistant.io/t/cron-logs-in-cli-after-move-to-another-computer/563355

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

NGINX Home Assistant SSL proxy

### What is the version of the add-on?

3.4

### Steps to reproduce the issue

Start homeassistant (in HAOS), wait for logs to appear.

## System Information

version | core-2023.4.6
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.10
os_name | Linux
os_version | 6.1.25
arch | x86_64
timezone | Europe/Berlin
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4812
Installed Version | 1.32.1
Stage | running
Available Repositories | 1272
Downloaded Repositories | 29

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 10.1
-- | --
update_channel | stable
supervisor_version | supervisor-2023.04.1
agent_version | 1.5.1
docker_version | 23.0.3
disk_total | 62.3 GB
disk_used | 14.7 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
[...]

PR #3018, opened 2023-05-01
nginx_proxy: Avoid logging to system console

agners:
Use the -d option to log to stderr instead.

Fixes: #3013


frenck (approved):
👍 Nice catch, thanks @agners 👍 

../Frenck

n00bster-za:
I am not sure what to do here where do I put the -d option?

frenck:
@n00bster-za You don't put it anywhere, this is an add-on change. It will become available as an update on your system.

PR #3020, opened 2023-05-02
nginx_proxy: Decrease crond log level

lmagyar:
This is a follow up on #3018

After #3018 the add-on's log is filled up with debug level messages from crond like:
```
crond: USER root pid 393 cmd run-parts /etc/periodic/15min
crond: USER root pid 411 cmd run-parts /etc/periodic/15min
crond: USER root pid 426 cmd run-parts /etc/periodic/15min
crond: USER root pid 427 cmd run-parts /etc/periodic/hourly
crond: USER root pid 484 cmd run-parts /etc/periodic/15min
```

Crond log level > 8 means none of these debug messages are logged, though errors and warnings are still emitted into the log. Crond uses only log level 5, 7, 8 internally, and level 8 is still too chatty.


nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/crond/run
```diff
@@ -4,4 +4,4 @@
 # Start crond service
 # ==============================================================================
 
-exec crond -f -d 8
+exec crond -f -d 9
```
agners, line 7: Shouldn't we set this to 7 then?

https://github.com/nawawi/busybox/blob/master/miscutils/crond.c#L178

lmagyar: No! :) The crond log_level works "backward"!
If we want to make it silent, we need this:
```
	if (level >= G.log_level) {
		bb_vinfo_msg(msg, va);
	}
```
To be like this:
```
	if (8 >= 9) {
```
So "-d 9" will cause to ***not*** execute the `bb_vinfo_msg` inside the `if`.
I've tested the 8 and 9, and 9 really made it silent.
The 7 will still emit something like this:
```
~ crond -f -d 7
crond: crond (busybox 1.35.0) started, log level 7
crond: USER root pid 546 cmd run-parts /etc/periodic/15min
crond: USER root pid 547 cmd run-parts /etc/periodic/hourly
```

agners (approved):

nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/crond/run
lmagyar: I agree. Note: errors and warnings from crond are not suppressed.

Issue #2910, opened 2023-03-11
multiprotocol high CPU use


jhbruhn:
### Describe the issue you are experiencing

The multiprotocol add-on is using a lot of CPU resources. According to glances, it is the mdnsd process utilizing 99% after a while.

This happens with OTBR enabled and disabled. ZigBee connected to Z2M

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Silicon Labs Multiprotocol

### What is the version of the add-on?

1.0.0

### Steps to reproduce the issue

1. Leave addon running for a couple of hours with any configuration 
2.
3.
...

## System Information

version | core-2023.3.3
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.10.10
os_name | Linux
os_version | 5.15.84-v8
arch | aarch64
timezone | Europe/Berlin
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.31.0
Stage | running
Available Repositories | 1320
Downloaded Repositories | 53

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 9.5
-- | --
update_channel | stable
supervisor_version | supervisor-2023.03.1
agent_version | 1.4.1
docker_version | 20.10.22
disk_total | 458.4 GB
disk_used | 41.7 GB
healthy | true
supported | true
board | yellow
supervisor_api | ok
version_api | ok
[...]

PR #3025, opened 2023-05-04
silabs-multiprotocol: Disable mdns service if otbr agent is disabled

jhbruhn:
As suggested in https://github.com/home-assistant/addons/issues/2910#issuecomment-1534230641, this PR disables the misbehaving mdns service if it is not needed due to disabling of the OTBR.


agners (approved):
Thanks. I'll add this change to the CHANGELOG.md and make it part of the Gecko SDK 4.2.3 update.

PR #3029, opened 2023-05-04
MariaDB: Use proper log redirection during backup

lmagyar:
This is a fix for #2983

cc @agners

In #2983 I used a workaround to write s6 service log messages to the add-on's log (`exec &> /proc/1/fd/1` in each oneshot service script). This is ugly.

After another few hours of googling and experimenting the proper solution seems to be:
- use s6 commands to redirect stdout and stderr before executing the command: `redirfd -wn 1 /proc/1/fd/1 fdmove -c 2 1 <COMMAND>` is roughly equivalent to `/bin/sh -c 'exec &> /proc/1/fd/1; exec <COMMAND>'`
- add `-v 2` to s6-rc to emit messages like `s6-rc: info: service mariadb-lock-core: starting`

Note: I think it is better to put these redirects into the config.yaml than to add it to supervisor, because it is not mandatory for an add-on to use s6 or /proc/1/fd/1 for logging.

With these modifications the add-on's log during backup is:
```
s6-rc: info: service mariadb-lock-core: starting
[16:49:04] INFO: Start MariaDB client (to lock tables for backups)
s6-rc: info: service mariadb-lock-core successfully started
s6-rc: info: service mariadb-lock-post: starting
[16:49:05] INFO: MariaDB tables locked
s6-rc: info: service mariadb-lock-post successfully started
s6-rc: info: service mariadb-lock-post: stopping
[16:49:16] INFO: MariaDB tables unlocked
s6-rc: info: service mariadb-lock-post successfully stopped
s6-rc: info: service mariadb-lock-core: stopping
[16:49:16] INFO: MariaDB client (to lock tables for backups) stopped
s6-rc: info: service mariadb-lock-core successfully stopped
```

### DIFF
--- a/mariadb/rootfs/usr/bin/lock-tables-for-backup
+++ b/mariadb/rootfs/usr/bin/lock-tables-for-backup
@@ -4,5 +4,6 @@
 # Lock tables for backup
 # ==============================================================================
 
+# Redirect stdout to the add-on's log
 exec &> /proc/1/fd/1
 exec s6-rc -v 2 -u change mariadb-lock
--- a/mariadb/rootfs/usr/bin/lock-tables-for-backup
+++ b/mariadb/rootfs/usr/bin/lock-tables-for-backup
@@ -1,0 +1,9 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+# ==============================================================================
+# Lock tables for backup
+# ==============================================================================
+
+redirfd -wn 1 /proc/1/fd/1 \
+    fdmove -c 2 1 \
+    s6-rc -v 2 -u change mariadb-lock
--- a/mariadb/rootfs/usr/bin/lock-tables-for-backup
+++ b/mariadb/rootfs/usr/bin/lock-tables-for-backup
@@ -4,6 +4,5 @@
 # Lock tables for backup
 # ==============================================================================
 
-redirfd -wn 1 /proc/1/fd/1 \
-    fdmove -c 2 1 \
-    s6-rc -v 2 -u change mariadb-lock
+exec &> /proc/1/fd/1
+exec s6-rc -v 2 -u change mariadb-lock
--- a/mariadb/rootfs/usr/bin/unlock-tables-for-backup
+++ b/mariadb/rootfs/usr/bin/unlock-tables-for-backup
@@ -4,5 +4,6 @@
 # Unlock tables for backup
 # ==============================================================================
 
+# Redirect stdout to the add-on's log
 exec &> /proc/1/fd/1
 exec s6-rc -v 2 -d change mariadb-lock
--- a/mariadb/rootfs/usr/bin/unlock-tables-for-backup
+++ b/mariadb/rootfs/usr/bin/unlock-tables-for-backup
@@ -4,6 +4,5 @@
 # Unlock tables for backup
 # ==============================================================================
 
-redirfd -wn 1 /proc/1/fd/1 \
-    fdmove -c 2 1 \
-    s6-rc -v 2 -d change mariadb-lock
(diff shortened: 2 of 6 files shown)


lmagyar:
Or maybe this format looks better:
```
backup_post:
  "redirfd -wn 1 /proc/1/fd/1 \
    fdmove -c 2 1 \
    s6-rc -v 2 -d change mariadb-lock"
backup_pre:
  "redirfd -wn 1 /proc/1/fd/1 \
    fdmove -c 2 1 \
    s6-rc -v 2 -u change mariadb-lock"
```

agners:
Hm, seems quite complicated :sweat_smile: 

Should we maybe just make sure that the output is part of the Supervisor log instead?

Since this is executed as part of the Backup, which is generally logged in the Supervisor log, it is probably more likely that people would look for issues in the Supervisor logs....

@mdegat01 thoughts?

lmagyar:
That has a quite nasty side-effect: all the log entries made by a longrun S6 service, who's state change is caused by these backup pre and post commands, will still log to the add-on's log, only the oneshot service and the s6-rc command would log to the supervisor.

Proof: manually executing `docker exec addon_core_mariadb s6-rc -v 2 -u change mariadb-lock` and `docker exec addon_core_mariadb s6-rc -v 2 -d change mariadb-lock`...

...will emit this to stdout (that can be redirected to supervisor log):
```
s6-rc: info: service mariadb-lock-core: starting
s6-rc: info: service mariadb-lock-core successfully started
s6-rc: info: service mariadb-lock-post: starting
[11:28:58] INFO: MariaDB tables locked
s6-rc: info: service mariadb-lock-post successfully started
s6-rc: info: service mariadb-lock-post: stopping
[11:29:05] INFO: MariaDB tables unlocked
s6-rc: info: service mariadb-lock-post successfully stopped
s6-rc: info: service mariadb-lock-core: stopping
s6-rc: info: service mariadb-lock-core successfully stopped
```

...but these entries will go to the add-on's log:
```
[11:28:58] INFO: Start MariaDB client (to lock tables for backups)
[11:29:05] INFO: MariaDB client (to lock tables for backups) stopped
```

OK, currently only mariadb uses these pre/post commands, we can remove logging from the longrun service's script, but this seems to be also hacky. And these scripts are executing inside the add-on's container, so I think the logs belong more to the add-on's log.

lmagyar:
Oh, and what about adding these commands to a script in eg. /usr/bin, and only calling these scripts from the backup_pre/post commands? Like:
```
backup_post: unlock_tables_for_backup
backup_pre: lock_tables_for_backup
```

That would be short and nice and everything. :)

mariadb/rootfs/usr/bin/lock-tables-for-backup
```diff
@@ -0,0 +1,8 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+# ==============================================================================
+# Lock tables for backup
+# ==============================================================================
+
+exec &> /proc/1/fd/1
```
agners, line 7:
```suggestion
# Redirect stdout to the add-on's log
exec &> /proc/1/fd/1
```

mariadb/rootfs/usr/bin/unlock-tables-for-backup
```diff
@@ -0,0 +1,8 @@
+#!/command/with-contenv bashio
+# shellcheck shell=bash
+# ==============================================================================
+# Unlock tables for backup
+# ==============================================================================
+
+exec &> /proc/1/fd/1
```
agners, line 7:
```suggestion
# Redirect stdout to the add-on's log
exec &> /proc/1/fd/1
```

agners (approved):

PR #3037, opened 2023-05-09
Remove "auto" language option from Whisper

synesthesiam:
The "auto" language option is confusing, slow, and will not work with the current wyoming-whisper library.
Removing from the add-on config.


frenck (commented):
This might lead to an issue with users that have currently configured as `auto` (those add-ons will fail, as the configuration is invalid.

Are you aware of that? And thus this change is correct that case?

../Frenck

PR #3038, opened 2023-05-09
Upgrade wyoming-faster-whisper to 0.0.3

synesthesiam:
Hashes Whisper model files during startup so incomplete files can be re-downloaded.
Fixes "auto" language when passed into Whisper model.


agners (approved):

synesthesiam:
OK, "auto" fix is actually in :)

PR #3040, opened 2023-05-10
silabs-multiprotocol: Use native zigbeed on amd64

agners:


silabs-multiprotocol/rootfs/etc/s6-overlay/scripts/otbr-enable-check.sh
```diff
@@ -9,6 +9,7 @@ if bashio::config.false 'otbr_enable'; then
     rm /etc/s6-overlay/s6-rc.d/user/contents.d/otbr-agent-rest-discovery
     rm /etc/s6-overlay/s6-rc.d/user/contents.d/mdns
     bashio::log.info "The otbr-agent is disabled."
+    exit 0
```
pvizeli, line 12:
```suggestion
    bashio::exit.ok
```

pvizeli (approved):

PR #3044, opened 2023-05-12
letsencrypt: Backup existing cert/key pair before updating

johnavich:
Added backup of existing cert/key pair before updating new cert/key pair


agners:
Why do you need this? Maybe the backup functionality of Home Assistant can help here?

johnavich:
I use the cert/key pair with SmartMeterTexas using node-red. In the period of time that the cert has been renewed, and me providing them the new cert, having an accessible backup of the cert allow me to automate the process and prevent me from losing a y relevant data. Until HA allows for importing fixed historical data, this is the easiest solution. It still protects the cert and key privileges, so security isn't a major concern.

PR #3047, opened 2023-05-13
Add armv7 support for whisper and piper

danijelst:
Add armv7 architecture support so the speach/text addons are available on respberry pi's


piper/build.yaml
```diff
@@ -2,6 +2,7 @@
 build_from:
   amd64: ghcr.io/home-assistant/amd64-base-debian:bullseye
   aarch64: ghcr.io/home-assistant/aarch64-base-debian:bullseye
+  armv7: ghcr.uo/home-assistant/armv7-base-debian:bullseye
```
frenck, line 5: So, upstream support is lacking... how is that fixed by adding these lines?

frenck: @danijelst You updated the PR by merging master... but you haven't answered this review comment. Maybe you missed it?

frenck:
Since there was no response at all, I'm declaring this PR stale and closing it.
Feel free to re-open this PR when you feel like that is the appropriate thing to do.

PR #3061, opened 2023-05-23
matter_server: Actually bump add-on version to 4.5.0

agners:
This has been missed in #3060.


frenck (approved):
🙈 Thanks @agners 👍 

../Frenck

PR #3064, opened 2023-05-24
openthread_border_router: Bump to latest OTBR POSIX version

agners:
Bump the OpenThread border router to the latest version. This includes the new REST API, which is incompatible with the previously patched REST API. Hence this will require a newer version of Home Assistant Core.


agners:
Home Assistant Core is now able to talk to the upstream API, so this is ready for merging (see https://github.com/home-assistant/core/pull/93544).

emontnemery (approved):

PR #3083, opened 2023-06-05
openthread_border_router: enhance REST API to allow dataset reset

agners:
Currently the dataset reset is broken for two reasons:
- The PUT method when not all fields are set uses unitialized data, which causes an invalid dataset to be generated
- Only using PUT method is not enough to generate a new dataset: The current dataset need to be completely removed. This adds the DELETE method as well.


emontnemery (approved):
LGTM

PR #3093, opened 2023-06-09
silabs-multiprotocol: Update to Gecko SDK v4.3.0

agners:
The Gecko SDK v4.3.0 update require some changes on the build side.

The current version of the Simplicity Studio Commander (Linux CLI) requires a newer JRE version.

The OpenThread Border Router got updated to a recent version which makes many of our patches obsolete (which are upstream meanwhile). However, the build flow changed a bit: Silicon Labs introduced a new vendor library support for RCP radios. The Multi-PAN Multiprotocol support using CPC requires this vendor library support (see also https://github.com/openthread/openthread/pull/8950).

With this change no patches are required ontop of the vanilla OTBR, instead some CMake build-time configuration is necessary. The CPC RCP support is still a compile time option, so we can't reuse the same OTBR we are using for regular OpenThread RCP radios, unfortunately.


darkxst:
@agners could we maybe include a note in the changelog that people using community RCP firmware builds will need to update dongles manually to 4.3.0 for this version? given the bump in CPC API.

silabs-multiprotocol/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 2.2.0
+
+- Use Silicon Labs Gecko SDK 4.3.0
+  - With this we get the OTBR version from May 17th, along with Border Agent ID
+    support (required for Android and iOS APIs)
```
agners, line 7:
```suggestion
    support (required for Android and iOS APIs)
- Note: This update needs a new Multi-PAN firmware. Home Assistant SkyConnect and Yellow get automatically updated by this add-on.
```

agners: @darkxst I think that should address your concern, does that look good for you?

puddly (approved):
Works for me.

silabs-multiprotocol/CHANGELOG.md
darkxst: yes that looks good to me.

PR #3112, opened 2023-06-23
matter_server: Bump Python Matter server to 3.5.2

agners:
With this upgrade also use the Python Matter container image as base for this add-on.

### DIFF
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 4.7.0
+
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):

frenck (changes requested):
The version didn't bump in the add-on config, was that intentional?

agners:
> The version didn't bump in the add-on config, was that intentional?

No, that wasn't intentional. It is fixed now!

frenck (approved):

PR #3116, opened 2023-06-27
Exclude Whisper model files from backups

synesthesiam:
Excludes large `.bin` Whisper model files from backups.


frenck (approved):
Thanks, @synesthesiam 👍 

../Frenck

PR #3117, opened 2023-06-27
Bump Matter server version to 3.6.1

marcelveldt:
- Bump Matter server to 3.6.0/3.6.1
- Bump version of Matter add-on to 4.8.0
- Bandaid solution for some issue reports we received where the loglevel retrieved from bashio was "NULL", resulting in a crashing add-on. This will set a default of info when the call returns empty.


frenck (changes requested):
Don't forget to bump the version and update the changelog :)

marcelveldt:
> Don't forget to bump the version and update the changelog :)

Well, I'm in doubt now as I also want to release a beta add-on ....

MartinHjelmare:
Bumping the version of the add-on isn't interesting in the PR title. Almost all PRs targeting add-ons will bump the version of the add-on. We should talk about what the change in the add-on is.

MartinHjelmare:
I suggest doing separate PRs for the server bump and the logging option change.

marcelveldt:
> I suggest doing separate PRs for the server bump and the logging option change.

haha, that's what I did and then was asked to do it all at once :-)

matter_server/build.yaml
```diff
@@ -1,8 +1,8 @@
----
 build_from:
   aarch64: ghcr.io/home-assistant-libs/python-matter-server:3.5.2
   amd64: ghcr.io/home-assistant-libs/python-matter-server:3.5.2
 args:
   BASHIO_VERSION: 0.14.3
   TEMPIO_VERSION: 2021.09.0
   S6_OVERLAY_VERSION: 3.1.5.0
+  MATTER_SERVER_VERSION: 3.6.1
```
agners, line 8: This is no longer used/needed since we base on the Matter Server container image now. You need to bump the base image used in `build.yaml` now.

PR #3118, opened 2023-06-28
Workaround for missing log level in matter add-on

marcelveldt:
We received some reports that sometimes the log_level is read as 'NULL', probably because the api call failed or something.
Although this must be addressed the proper way, this is a temporary bandaid to prevent a crashing add-on.

As a small reminder this behavior is logged.


MartinHjelmare:
This will be released in together will the server bump in the other PR.

MartinHjelmare:
Looks good, but I'm not a bashio expert.

frenck (approved):

PR #3127, opened 2023-07-07
Add Trusted CA User Keys as SSH configuration option

mdawsonuk:
This PR adds an additional option at `server.trusted_user_ca_keys` which sets the TrustedUserCAKeys value within sshd_config, as well as writes the value to `/etc/ssh/ssh_ca.pub`. This will enable users with public keys signed by that Certificate Authority to connect to OpenSSH without needing to list authorized_keys.

### DIFF
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.8.0
+
+- Add Trusted CA User Keys to configuration
+
 ## 9.7.1
 
 - Upgrade Home Assistant CLI to 4.26.0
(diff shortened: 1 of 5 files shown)


mdawsonuk:
I have extended the scope of this PR slightly by adding the authorized principals option too - this enables certificates to be generated with principals without the name `root` (allowing a user to authenticate as root against any device) and still be used to authenticate as the root user in this container.

mdawsonuk:
@frenck Could I get some feedback on this PR please?

mdegat01:
What is the use case for this option? HA is self-hosted software, generally with a single admin. This addon presents a way for that single admin to get to some of the advanced features of the HA CLI that haven't yet been migrated to the UI for one reason or another. Given this target audience, why would a feature like this be needed?

In an enterprise or organization setup where one or more super admins are generating keys to grant some level of access to other admins this would make sense. But that's not a use case that exists in HA right now or one it supports.

PR #3130, opened 2023-07-11
letsencrypt: Add ECDSA key option

WillCodeForCats:
This PR adds support for configuring NIST/SECG elliptic curve keys (ecdsa) in the Let's Encrypt addon.


letsencrypt/config.yaml
```diff
@@ -91,4 +91,6 @@ schema:
     inwx_username: str?
     inwx_password: str?
     inwx_shared_secret: str?
+  elliptic_curve: "list(secp256r1|secp384r1|secp521r1)?"
```
mdegat01, line 94: This should not be an option. Addons are not supposed to present every possible option to user, especially not hyper technical ones such as these. Whenever possible, addon devs should preselect the most sensible option for users so they don't have to think about it. This is one of those cases, please default to the best option and remove this from config.

letsencrypt/config.yaml
```diff
@@ -91,4 +91,6 @@ schema:
...
+  key_type: "list(rsa|ecdsa)?"
```
mdegat01, line 95: Does this need to be an option? I assume one of these is more secure with the other. Assuming there is no backwards compatibility issue we should simply pick the more secure option and remove the option so users don't have to think about it. Unless changing to ecdsa can cause a backwards compatibility issue or is incompatible with some of the providers, in which case please provide this information.

mdegat01 (changes requested):
I am fine with the change but don't think we should add any addon options unless we absolutely have to. I definitely don't think we need a second option to allow users to choose which form of `ecdsa`. That's just forcing them to google and copy from the first comment in a stackoverflow post they see anyway.

Also you need to bump the version in the config file and add to the changelog.

letsencrypt/config.yaml
WillCodeForCats: ECDSA is the future while RSA is the "safe" option.

letsencrypt/config.yaml
WillCodeForCats: I'll say secp384r1 then since it's the middle option.

letsencrypt/config.yaml
WillCodeForCats: I personally don't have a problem with default to ECDSA since I want this option.

The question is not about "more secure", rather it's how much do you care about older clients when deciding what the default should be, and should the user have an option to select the highly compatible key type or the modern one? For example if someone out there is still using Windows XP then ECDSA won't work on XP. I can't answer how many HA users may want support for XP if the default is changed.
 
https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29

WillCodeForCats: Which poses a further question: to what degree should old platforms like XP still be considered when making decisions? I'm sure that if the default gets changed to ECDSA someone will come out of the woodwork saying their ancient whatever can't use https anymore and it worked fine before why should they have to update etc. Old tablets being used as dashboards running who knows what vintage of software on them is another angle.

Ultimately not allowing an option to use ECDSA is undesirable in a modern sense. The option to start using ECDSA should be added first. At some point in time, yes, the default should be changed to ECDSA. Is that time now? I think that's a discussion larger than the scope of this PR.

mdegat01: Windows XP isn't a platform we support so I don't see that as an issue. And if the other platforms that are likely to run into issues are equally as old then I'd say it would be better to simply switch to `ecdsa` and not look back.

The thing to remember is that the Let's Encrypt addon is not the same as the Let's Encrypt software. We are not trying to present every option Let's Encrypt has. The idea of addons is that the developer is the Subject Matter Expert and should present as few options as possible, making the correct choices for users whenever they can. As this is all self-hosted software, users which are themselves subject matter experts in the software and disagree with the choices always have the option of skipping the addon and running the service themselves.

The other thing to remember is its easy to add options, its difficult to remove them. I'll keep an eye on issues and if this causes a bunch we can add the option later quickly enough. Updating addons also takes a backup so users can rollback their own addons quickly and easily in the meantime.

WillCodeForCats: Okay then whatever you want, your call to change the default.

PR #3134, opened 2023-07-14
openthread_border_router: update OTBR and universal-silabs-flasher

agners:
Use the latest OTBR version which allows to use the DELETE method to
the /node endpoint to erase all persistent information of the node.


marcelveldt (approved):
nice!

emontnemery (approved):

PR #3154, opened 2023-07-31
silabs-multiprotocol: Update to Gecko SDK v4.3.1

agners:


puddly (approved):
Everything works fine for me!

MattWestb:
Great work our **Stefan** _and_ **Puddly** i testing it later today or tomorrow and reporting back if getting problems with my Billys RCPs.

MattWestb:
Have updating my 3 test system and all is getting the came error then starting Zigbeed:
```
Logger: bellows.ezsp.protocol
Source: runner.py:179
First occurred: 15:22:12 (1 occurrences)
Last logged: 15:22:12

Couldn't set EzspConfigId.CONFIG_PACKET_BUFFER_COUNT=255 configuration value: EzspStatus.ERROR_INVALID_ID
```
If i remember right is is 255 allocation all free RAM as buffers instead having one fix value.
 My test system is:
NanoPi Neo armbian Ubuntu 32 bits RCP only Zigbeed.
RK3228 armbian Ubuntu 32 bits RCP OTBR and Zigbeed.
RK3318 armbian Ubuntu 64 bits RCP only Zigbeed.

All having one IKEA Billy RCP (MG1P) with latest firmware and looks working OK  for over 10 minutes now.

puddly:
https://github.com/zigpy/bellows/pull/576 should take care of the warning (by turning it to a debug message) but this type of breaking change shouldn't be in a .1 release. Perhaps for Zigbeed it doesn't make sense to expand this, since RAM is assumed to be effectively limitless?

MattWestb:
OK i is running all HA on 20232.08.0 and have skipping the 1 but shall updating to 2 then its being released then its also having some homekit fixes that i need for my EVE plugs that running on Thread.

Thanks for feedback all locks working OK so its great !!!

PR #3156, opened 2023-07-31
silabs_flasher: Update universal-silabs-flasher and firmwares

agners:
Update universal-silabs-flasher to v0.0.13 and EmberZNet firmwares for Home Assistant Yellow and SkyConnect to the latest version v7.3.1.0.


darkxst:
Fwiw, I took a different approach in my fork of this, and switched to python 3.11 to be able to keep using the HA wheels (as the cp310 wheels were all deleted recently) rather than building the missing wheels in the add-on.  In this case all arches built fine including i386 (although I havent done any run time testing of i386).

agners:
> 

Right, we are using the Alpine base images here, so we can do that. Thanks for the hint!

silabs_flasher/Dockerfile
```diff
@@ -10,11 +10,17 @@ RUN \
     && apk add --no-cache \
         py3-pip \
         python3 \
+    && apk add --no-cache --virtual .tmp-build-deps \
```
darkxst, line 13: You can drop these build-deps, the bump to python 3.11 is enough.

silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
```diff
@@ -51,7 +51,7 @@ if bashio::config.has_value 'firmware_url'; then
 else
     # Assume to run on Yellow if UART4 is mapped to ttyAMA1
     if [ -d /sys/devices/platform/soc/fe201800.serial/tty/ttyAMA1 ] && [ "${device}" == "/dev/ttyAMA1" ]; then
-        firmware="NabuCasa_Yellow_EZSP_v7.2.2.0_PA32_ncp-uart-hw_115200.gbl"
+        firmware="NabuCasa_Yellow_EZSP_v7.3.1.0_PA32_ncp-uart-hw_115200.gbl"
     else
         # Check device manufacturer/product information
         usb_device_path=$(realpath /sys/class/tty/$(readlink /sys/class/tty/$(basename ${device}))/../../../../)
```
darkxst, line 57: I ended up using the following here, so it correctly detects the USB device on all drivers (for e.g. CDC), although probably doesnt make much difference for the use case here, it does show up in the logs (here and multiprotocol) "checking device identifying "some USB Hub identifier".
```
usb_device_path=$(realpath /sys/class/tty/$(readlink /sys/class/tty/$(basename ${device}) | sed 's:/[^/]*tty[^/]*::g' )/../)
```

agners: Hm, using sed to cut the dev path seems a bit hacky too, but I guess its the best we can do :man_shrugging:

darkxst: yeh it feels like there should be a cleaner way to just get the usb device path, but I couldnt find one...

puddly (approved):

darkxst:
@agners You forgot to bump the version in `config.yaml`!

frankebobdk:
@agners 

I’m seeing this as well. 
getting this error when trying to fetch update for the addon:

Logger: homeassistant.components.hassio
Source: components/hassio/websocket_api.py:130 
Integration: Home Assistant Supervisor ([documentation](https://www.home-assistant.io/integrations/hassio), [issues](https://github.com/home-assistant/core/issues?q=is%3Aissue+is%3Aopen+label%3A%22integration%3A+hassio%22)) 
First occurred: 16.45.31 (1 occurrences) 
Last logged: 16.45.31

Failed to to call /store/addons/core_silabs_flasher - Addon core_silabs_flasher with version latest does not exist in the store

agners:
> @agners You forgot to bump the version in `config.yaml`!

Addressed with #3167.

PR #3152, opened 2023-07-27
letsencrypt: support DuckDNS DNS challenge

iranl:
Adds DuckDNS DNS challenge support.

CNAME for custom domain "test.example.com" with DuckDNS domain "abc.duckdns.org" should be set up as follows:

```commandline
_acme-challenge.test.example.com. 600 IN CNAME abc.duckdns.org.
```

Note: Alias / custom domain support for obtaining a Let's Encrypt certificate in the DuckDNS addon is broken (#2505) and devs have proposed to remove support for aliases altogether (#2662 #2964).

Proposed solution was to use the DuckDNS addon for keeping a dynamic IP up to date only and configure the Let's Encrypt addon to obtain a certificate (through a HTTP challenge or third party DNS provider that is supported by the Let's Encrypt addon).

This PR allows for DuckDNS DNS challenges including for custom domains and as such no ports need to be opened to allow a HTTP challenge and no secondary DNS provider is needed to complete the DNS challenge for custom domains.

PR #3157, opened 2023-08-01
letsencrypt: Add desec.io support DNS challenge 

henrikhierl:


mdegat01 (changes requested):
Missing a changelog and version bump. Looks good otherwise.

kmahelona:
+1 to getting this PR merged. 🙏🏽

agners:
@henrikhierl the add-on now got updated to the latest Alpine version etc. Can resolve the merge conflict and add the version bump to the changelog and config.yaml (similar to how #3152 was done)? Then we should be able to merge this soon.

agners (approved):

max-hh0:
Hi, I am wondering if there is a bug in the deSEC integration. The letsencrypt-addon publishes the TXT-record for „_acme-challenge.sub.domain.de“. But the record automatically disappears after very few seconds. So letsencrypt can not verify the request.
If I manually create a TXT-record for that domain it is persisted / does not automatically disappear. 

Could the problem be caused by the addon deleting the record too early?

Would be great to make this work with deSEC because this DNS provider supports DNSSEC (and is free).

Cheers

Issue #3139, opened 2023-07-17
Silicon Labs Multiprotocol: all of sudden, increased CPU usage HAOS


Anto79-ops:
### Describe the issue you are experiencing

Hello,

I just recently enabled multi-pan on my SkyConnect on HAOS/RPi and ZHA. 

Home Assistant 2023.7.2
Supervisor 2023.07.1
Operating System 10.3
Frontend 20230705.1 - latest
Silicon Labs Multiprotocol: 2.2.0
ZHA 
Open Thread Border Router
Thread
Matter  4.8.3

~~I don't actually have any thread or matter devices (I tried to install nanoleaf, but it simply would not work)~~ I have 2 thread devices now.

I do have 5 google nest hubs that are acting as a TBR of which I added my SkyConnect, (firmware 4.3) too so I have a single thread network with 6 thread border routers.

I had to change my ZHA channel from 20 to 16.

Every once in a while (almost every 2 days), I notice increased CPU usage of HA and glances tells me it's coming from the Silicon Labs Multiprotocol addon (see towards the end this CPU Plot)

![image](https://github.com/home-assistant/addons/assets/79593761/ab104cad-14cc-41e2-bcf3-ac053c07e999)

The logs for the addon (or at least what I could copy and paste are full of this):

```
otbr-agent[314]: 02:17:15.577 [N] MeshForwarder-:     src:[fe80:0:0:0:dca7:fa1:377c:5f11]:19788
otbr-agent[314]: 02:17:15.577 [N] MeshForwarder-:     dst:[fe80:0:0:0:e821:177b:8c9e:79e0]:19788
otbr-agent[314]: 02:17:16.569 [N] MeshForwarder-: Failed to send IPv6 UDP msg, len:125, chksum:2edc, ecn:no, to:ea21177b8c9e79e0, sec:no, error:NoAck, prio:net
otbr-agent[314]: 02:17:16.570 [N] MeshForwarder-:     src:[fe80:0:0:0:dca7:fa1:377c:5f11]:19788
otbr-agent[314]: 02:17:16.570 [N] MeshForwarder-:     dst:[fe80:0:0:0:e821:177b:8c9e:79e0]:19788
otbr-agent[314]: 02:17:17.397 [W] Mle-----------: Failed to process UDP: Duplicated
otbr-agent[314]: 02:17:17.525 [W] Mle-----------: Failed to process UDP: Duplicated
[...]

PR #3169, opened 2023-08-08
silabs-multiprotocol: update mDNSResponder to 1790.80.10

agners:
Update the mDNSResponder to the latest version 1790.80.10. This might help with high CPU usage as reported in #2910 and #3139.


puddly (approved):
Tested by multiple users for a few weeks and confirmed working.

PR #3181, opened 2023-08-17
Update Piper add-on to 1.3.2

synesthesiam:
- Add voices for Hungarian, Turkish, Portuguese, Slovak, and Czech
- Add `upgrade_voices` and `debug_logging` options
- Upgrade to Debian bookworm


piper/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
-## 1.1.0
+## 1.3.0
+
+- Add voices for Hungarian, Turkish, Portuguese, Slovak, and Czech
+- Add `upgrade_voices` and `debug_logging` options
+- Upgrade to Debian bookworm
+
+## 1.2.0
```
balloob, line 9: This was 1.1.0 before?

synesthesiam: Yeah, I made a mistake last time.

piper/translations/en.yaml
```diff
@@ -27,5 +27,21 @@ configuration:
...
+    description: >-
+      Number of Piper processes to run simultaneously (default is 1). Increase
+      if you need to quickly switch between multiple voices, but beware that
+      this will increase RAM usage by the add-on.
+  update_voices:
+    name: Update voices
+    description: >-
+      Download list of new voices automatically every time the add-on starts.
+      You will also need to reload the wyoming integration for Piper in Home
+      Assistant to see new voices.
```
frenck, line 41:
```suggestion
      Download the list of new voices automatically every time the add-on starts.
      You must also reload the Wyoming integration for Piper in Home Assistant
      to see new voices.
```

frenck (changes requested):

piper/Dockerfile
```diff
@@ -12,14 +12,16 @@ ARG PIPER_RELEASE
...
     apt-get update \
     && apt-get install -y --no-install-recommends \
-        netcat \
+        netcat-traditional \
         python3 \
         python3-pip \
+        python3-venv \
     \
-    && pip3 install --no-cache-dir -U \
+    && python3 -m venv .venv \
```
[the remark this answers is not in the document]
synesthesiam: The Python in Debian bookworm gives this error when trying to upgrade system packages:

```
$ pip3 install --upgrade pip
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.
```

There is likely a flag to override, but I figured it was best to go with the flow.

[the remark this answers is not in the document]
frenck: I don't agree, as this is not a normal user runtime.

I'd suggest restoring the original and set the `PIP_BREAK_SYSTEM_PACKAGES=1` on the environment.

piper/config.yaml
```diff
@@ -19,14 +19,18 @@ options:
...
   speaker: int
   length_scale: float
   noise_scale: float
   noise_w: float
   max_piper_procs: int
+  debug_logging: bool
+  update_voices: bool
```
balloob, line 33: Any reason we make this an option and not always on ?

piper/Dockerfile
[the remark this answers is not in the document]
synesthesiam: Removed the venv

piper/config.yaml
synesthesiam: It unfortunately prints a line each time the healthcheck hits (which asks for available voices). So you'd end up with quite a few log messages over time with this on.

frenck (approved):
✅ Approve from my end, not sure if the above open discussions need to be resolved before merging this one in.

../Frenck

piper/config.yaml
```diff
@@ -19,14 +19,18 @@ options:
   noise_scale: 0.667
   noise_w: 0.333
   max_piper_procs: 1
+  debug_logging: false
+  update_voices: false
```
agners, line 23:
```suggestion
  update_voices: true
```

piper/config.yaml
agners: @synesthesiam  I think @balloob was referring to the lower option (`update_voices`). I agree it might be a good idea to do this by default on startup, but maybe there are reasons we don't want it to be enabled by default?

piper/config.yaml
synesthesiam: OK, let me make the loading a bit more robust and I'll enable it by default.

synesthesiam:
@agners Ready to merge.

piper/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Changelog
 
-## 1.1.0
+## 1.3.0
```
agners, line 3:
```suggestion
## 1.3.2
```

agners (approved):

frenck (approved):

PR #2881, opened 2023-02-16
Allow pre-hashed passwords in mosquitto logins config.

scolby33:
I would like to not hard-code visible passwords in my configuration files. It is possible to pass a pre-hashed password to mosquitto, and this PR adds the ability to use such a value.

This change adds the ability to pass already-hashed passwords in the logins array. If the password begins `PBKDF2$sha512$` and has length 134, it is not hashed again and is incorporated directly into the `pw` file. Otherwise, the existing logic is used.

I have tested these changes locally and found that they worked as expected.

Things I considered but did not implement:

- using a new `password_hash` key in the configuration: I didn't do this because the schema doesn't let you specify mutually-exclusive or "exactly one of these" options, so instead I decided to re-use the existing password key
- doing more in-depth checks against the hash: it seems unlikely that someone would choose a password that matches the format of a PBKDF2 hash and is so long
- checking for other hashing algorithms: it is my understanding from https://github.com/home-assistant/addons/blob/35d99c4f8b484802bebe8c72bb53c12d6f8e0a24/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl#L19 that only PBKDF2 can be used

PR #3183, opened 2023-08-20
Allow pre-hashed passwords in mosquitto logins config.

scolby33:
I am creating a new PR based on #2881 since it was closed as stale, even though it is still relevant. Can someone take the time to review this eight-line changeset that has been languishing without review comment for over 96 working days? If there are improvements to be made, I would be happy to make them. I would really appreciate the ability to use this feature.

---

I would like to not hard-code visible passwords in my configuration files. It is possible to pass a pre-hashed password to mosquitto, and this PR adds the ability to use such a value.

This change adds the ability to pass already-hashed passwords in the logins array. If the password begins `PBKDF2$sha512$` and has length 134, it is not hashed again and is incorporated directly into the `pw` file. Otherwise, the existing logic is used.

I have tested these changes locally and found that they worked as expected.

Things I considered but did not implement:

- using a new `password_hash` key in the configuration: I didn't do this because the schema doesn't let you specify mutually-exclusive or "exactly one of these" options, so instead I decided to re-use the existing password key
- doing more in-depth checks against the hash: it seems unlikely that someone would choose a password that matches the format of a PBKDF2 hash and is so long
- checking for other hashing algorithms: it is my understanding from https://github.com/home-assistant/addons/blob/35d99c4f8b484802bebe8c72bb53c12d6f8e0a24/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl#L19 that only PBKDF2 can be used


mdegat01 (changes requested):
You're right about the lack of mutually exclusive schema option, something we need to work on (https://github.com/home-assistant/supervisor/issues/2640 has been open for a while though not entirely sure that covers this particular gap). However the prefix at the start of the password value poses two issues to me:
1. Technically that could be someone's password value. It's extremely unlikely but not impossible
2. It's very unintuitive and makes it appear to users like they can change the type of hash

I think instead it would be better to add a `password_pre_hashed` boolean option to the config of each login. This field would be optional and let us know if how whether to hash the password value or use it as is in the code. And then remove the prefix.

In addition you need to bump the version in the addon's config and add an entry to the changelog. Thanks though, this is a good idea, sorry it languished so long.

agners (approved):
With the YAML issue addressed, LGTM.

scolby33:
I adjusted the YAML line lengths! (Might have been a bit too conservative looking at it, but it should pass the linter now.)

mdegat01 (approved):

mdegat01:
Thanks @scolby33 👍

PR #3198, opened 2023-08-30
Expose alternative ports

jkolo:
I have added alternative ports on which listening can be added via customisation, e.g. for an alternative certificate, disabled http/2 or yet another configuration. Ports are not forwarded by default, so they do not affect standard configurations.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.5.1
+
+- Added alternative ports 
+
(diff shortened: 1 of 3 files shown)


mdegat01:
The nginx proxy addon is intended to be an extremely simple way to get Home Assistant listening externally over SSL while still being able to use non-SSL locally. It is not intended to be a fully configurable reverse proxy.

If you want a more fully configurable reverse proxy, consider using the Nginx Proxy Manager addon in the community addon store. Or deploying your own Nginx, Caddy or the reverse proxy of your choice.

PR #3199, opened 2023-08-30
letsencrypt: Update DOCS.md

asheroto:
I am using the Home Assistant operating system on Windows via VMWare. I assumed that the same or similar steps applied to other platforms as well. I made many changes but I think it will help folks understand what to do. I was struggling with some of the configuration at first since I'm new to HA, but once I figured it out it made since, so I thought I'd tweak the documentation. Let me know what you think or if you need me to change anything. 😊

- Move some sections for improved readability.
- Add detailed explanations.
- Add troubleshooting section.
- Add screenshots.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -534,7 +534,7 @@ Previously, Cloudflare’s "Global API Key" was used for authentication, however
 
 ## Certificate Files
 
-The certificate files will be available within the `/ssl` share after successful request of the certificates. By default other addons are referring to the correct path of the certificates. Note that the `Let's Encrypt` log files may tell you that the certificate files are located under `/data/letsencrypt/live/` but that is NOT true. They are under the `/ssl` directory.
+The certificate files will be available within the `/ssl` share after successful request of the certificates. By default other addons are referring to the correct path of the certificates. If you need to see more complete logs then are visible in the UI you can find them in the `/ssl` directory. Or by doing `ha addons logs core_letsencrypt` from the CLI.
 
 After clicking **Save** on the configuration, go back to the **Info** tab and click **Start**. That will engage Certbot and start requesting the certificate. You can view the status under the **Log** tab.
 
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -7,8 +7,8 @@ Additional Let's Encrypt setup instructions can be [found here](https://www.home
 
 Follow these steps to get the add-on installed on your system:
 
-1. Navigate in your Home Assistant frontend to **Settings** → **Add-ons** → **Add-on store**.
-2. Find the **Let's Encrypt** add-on and click it.
+1. Navigate in your Home Assistant frontend to [Settings → Add-ons → Add-on store](https://my.home-assistant.io/redirect/supervisor_store/).
+2. Find the [Let's Encrypt](https://my.home-assistant.io/redirect/supervisor_addon/?addon=core_letsencrypt) add-on and click it.
 3. Click on the **Install** button.
 
 ## How to use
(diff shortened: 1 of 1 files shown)


asheroto:
Ah, thank you. Just signed.

letsencrypt/DOCS.md
```diff
@@ -513,39 +532,68 @@ on the DNS zone to be used for authentication.
...
-dns-inwx
 ```
+http:
+  ssl_certificate: /ssl/fullchain.pem
+  ssl_key: /ssl/privkey.pem
+```
+
+You may want to change additional settings, but the two above are required for SSL certificates to work correctly.
+
+[Full HTTP section documentation](https://www.home-assistant.io/integrations/http/)
```
mdegat01, line 555: If you want to go into this part here then we should do the recommended way. Adding your SSL certificate to the HA configuration is not recommended outside of a core or container install (since those have no other options). If addons are available the recommended approach is to install the [nginx proxy add-on](https://my.home-assistant.io/redirect/supervisor_addon/?addon=core_nginx_proxy) and add your SSL certificate there.

The problem with adding the certificates to HA directly is you make LAN access difficult or impossible for most users. HA can now only be accessed via SSL even on LAN, there is no HTTP-only option. And that means if the user tries to go to `homeassistant.local:8123` (or whatever they use for their LAN URL) they get certificate errors that they now have to figure out how to deal with and their mobile app no longer works on LAN.

So if we want to recommend how they use these options in conjunction with HA we need to describe the recommended approach. Use a reverse proxy (the nginx proxy addon is by far the simplest), do not add them directly to HA config.

mdegat01 (changes requested):
Biggest change required is definitely in the section on how to use the SSL certs with HA. As I noted we should not be recommending they configure the SSL options under HTTP. That's not actually the easiest way and ends up creating a bunch of new issues they have to deal with in the form of invalid certs on LAN.

Your note about the easy option being split DNS makes me think you host your own DNS server. That's great if you do but most users do not. And I can tell you that DNS issues (particularly from people that do host their own DNS servers) dominate issue reports around the HA repos. So this isn't actually an option we recommend to people either as its often replacing 1 issue with 10 new ones for users that don't know DNS very well (i.e. most users).

letsencrypt/DOCS.md
asheroto: Hey sorry for the delay I got sick with Covid 😷 but am better now.

That makes sense.

So what should I do instead since this is for the Let's Encrypt doc? Is there any other reason to use LE other than to get your own SSL cert?

letsencrypt/DOCS.md
```diff
@@ -513,39 +532,68 @@ on the DNS zone to be used for authentication.
...
+
+Done!
+
+# Troubleshooting
+
+If PortChecker says your HA port is open, but you cannot connect from within your own network, keep in mind that a lot of basic routers won't let you access your public IP address from within your own local network. For instance, if you've set up port forwarding on your router to send traffic on port 8123 to your Home Assistant at local IP 192.168.1.150, trying to connect to your public IP (say, 1.2.3.4) from another local device (like one at 192.168.1.140) might not work. **This is because most routers only handle port forwarding for incoming traffic from the internet, not from within your home network.**
+
+If you run into this problem, you have a few options:
+- Some routers can get around this if you turn on or configure a feature called "hairpin NAT" or "NAT loopback". Look up how to enable that. It basically tells your router to allow accessing your public IP from a private IP address.
+- The simplest fix is usually to set up what's called "split DNS". Depending on your DNS setup, this makes it so when you're at home, your subdomain directs to your local IP, but when you're out and about, it'll point to your public IP.
```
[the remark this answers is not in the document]
asheroto: That's true, I'm used to working with other IT people and kind of jumped to the conclusion that anyone with HA would be too, but I suppose not all are.

You think I should just remove these lines? Or maybe say "for advanced users only" ?

Issue #3186, opened 2023-08-21
VLC add on needs config to expose/set password to work with vlc-telnet integration in case discovery doesn't work/happen


BobC76:
### Describe the issue you are experiencing

The add on documentation says simply:

> How to use
> Start the add-on.
> Add the VLC Telnet integration to Home Assistant, see documentation: https://www.home-assistant.io/integrations/vlc_telnet

As noted in a comment below, this brief set of instructions assumes one can set up the integration via the discovery information that the add-on sends to Home Assistant Core. But this doesn't always happen, and you can't always just confirm the discovery under Devices & Services in the GUI Settings..

As the latest vlc_telnet expects a password, and there's no way to set it on the add on, if for some reason discovery doesn't happen, it can't work as described.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

VLC

### What is the version of the add-on?

0.2.0

### Steps to reproduce the issue

1. Follow the instructions at https://github.com/home-assistant/addons/blob/master/vlc/DOCS.md - they don't work because the vlc_telnet integration requires a password, and no-where does the vlc documentation state where that can be set/seen

## System Information

version | core-2023.8.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.4
os_name | Linux
os_version | 6.1.39
arch | x86_64
timezone | Europe/London
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.32.1
Stage | running
Available Repositories | 1272
Downloaded Repositories | 43

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 10.4
-- | --
update_channel | beta
supervisor_version | supervisor-2023.08.1
agent_version | 1.5.1
docker_version | 23.0.6
disk_total | 30.8 GB
disk_used | 8.3 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
[...]

PR #3200, opened 2023-09-01
Update DOCS.md

BobC76:
Addressing some confusion I had when I had already installed the integration and then re-installed the add-on, and auto discovery didn't just work. See https://github.com/home-assistant/addons/issues/3186

### DIFF
--- a/vlc/DOCS.md
+++ b/vlc/DOCS.md
@@ -17,10 +17,7 @@ Follow these steps to get the add-on installed on your system:
 
 ## Configuration
 
-Add-on configuration:
-
-```yaml
-```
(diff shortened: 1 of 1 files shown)


vlc/DOCS.md
```diff
@@ -17,10 +17,7 @@ Follow these steps to get the add-on installed on your system:
 
 ## Configuration
 
-Add-on configuration:
-
-```yaml
-```
+Note that automatic discovery is expected when the add-on is being installed, and if that is not the case the user should make sure to remove the Telnet integration and reinstall the add-on.
```
agners, line 20:
```suggestion
Note that automatic discovery is expected when the add-on is being installed. If no "VLC media player via Telnet" integration is being discovered, make sure to remove any stale "VLC media player via Telnet" integrations and reinstall the add-on to trigger discovery again.
```

PR #3201, opened 2023-09-02
nginx_proxy: Improve domain description

garyt72:
### Option: `domain` (required)

The `domain` configuration description was not clear, stating the value should be the domain name (example.com) instead of the fully qualified domain name of the sever (server.example.com).  This was shown in the example yaml configuration, but not clear here

change:
   The domain name to use for the proxy.
to: 
   The server's fully qualified domain name to use for the proxy.


frenck (approved):
Thanks, @garyt72 👍

../Frenck

PR #3244, opened 2023-10-05
duckdns: Support wildcard certs for subdomains in DuckDNS addon (resolves #336)

### LINKED ISSUE #336: DuckDNS Addon cannot create wildcard certificate
*@JAAlperin commented on Jun 19, 2018, 5:25 PM UTC:*

{
"lets_encrypt": {
"accept_terms": true,
"certfile": "fullchain.pem",
"keyfile": "privkey.pem"
},
"token": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
"domains": [
"xxx.duckdns.org",
"*.xxx.duckdns.org"
],
"seconds": 1800
}

results in this:

OKOKOKOK + Responding to challenge for xxx.duckdns.org authorization...

*   Cleaning challenge tokens...
    OKOKOKOK + Challenge validation has failed :(
    ERROR: Challenge is invalid! (returned: invalid) (result: {
    "type": "dns-01",
    "status": "invalid",
    "error": {
    "type": "urn:ietf:params:acme:error:unauthorized",
    "detail": "Incorrect TXT record "gYvyuKSEz-Sa58khnK9KUc0MUu3C5WDf_7qRDwlDAcM" found at _acme-challenge.xxx.duckdns.org",
    "status": 403
    },
    "url": "[https://acme-v02.api.letsencrypt.org/acme/challenge/szX3vuPHMvfMYF-heQcbgmGA4sYZzSVXMMLbjhDm3_Q/5197664255](https://acme-v02.api.letsencrypt.org/acme/challenge/szX3vuPHMvfMYF-heQcbgmGA4sYZzSVXMMLbjhDm3_Q/5197664255)",
    "token": "MxeILOKZ8GYrxuHkWPQYTu3_ZXjbexpDp_miok76grY"
    })

If I understand the many internet posts regarding this issue, it has to do with HOOK_CHAIN=yes, the tokens in the single TXT record, and has been fixed already in certbot via a changed script.

*This issue was moved by [frenck](https://github.com/frenck) from home-assistant/home-assistant.github.io/issues/5574.*

kyle-cackett:
This PR resolves the issue described in [this comment](https://github.com/home-assistant/addons/issues/336#issuecomment-1234475344) on issue #336 (copied below for convenience) which occurs when the domains array contains entries of the form `"*.xxx.duckdns.org > xxx.duckdns.org"`. This type of entry is used to obtain wildcard certificates from Let's Encrypt / dehydrated.

>> There's a quick workaround if you only need to use subdomains of "xxx.duckdns.org". Configure the addon with:
"domains": [
  "*.xxx.duckdns.org > xxx.duckdns.org"
]
This will create a wildcard certificate for "*.xxx.duckdns.org" with "xxx.duckdns.org" as an alias (required by dehydrated). The certificate will not be valid for "xxx.duckdns.org", but only for its subdomains.

>It doesn't work propertly. It generate wildcard certificate but doesn't update IP on DuckDNS.org.
In log only empty warnings:
[18:52:58] WARNING: 

The `sed` substitution `'s/[^,]*[ > ]//g'` matches all occurrences of strings which do not contain a `,` (so that after the `join(",")` at most one match is found per entry in the domain array) that end with the characters ` > ` and removes them leaving only valid domains according to the [DuckDNS API](https://www.duckdns.org/spec.jsp). Note that this transformation will not impact Let's Encrypt because the `$DOMAINS` variable is only used by DuckDNS. The `le_renew` function grabs the domains from the config directly on line 23:

https://github.com/home-assistant/addons/blob/72fac56dae57fe3573a1e568b248fa304cca2b1b/duckdns/data/run.sh#L23C1-L24C1

### Testing

1. Wrote the following script to test the sed substitution:

```

TEST_DOMAINS=$(bashio::config 'test_domains | join(",")' | sed 's/[^,]*[ > ]//g')
echo $TEST_DOMAINS
```
Added the following field to the json object in `/tmp/.bashio/addons.self.options.config.cache` 
```
"test_domains": ["*.example.duckdns.org > example.duckdns.org","test.duckdns.org","*.sample.duckdns.org > sample.duckdns.org","subdomain.duckdns.org","demo.duckdns.org"],
```
Ran the script and received the following output:
```
example.duckdns.org,test.duckdns.org,sample.duckdns.org,subdomain.duckdns.org,demo.duckdns.org
```
2. Modified my IP address on duckdns.org and ran the new version of this `run.sh` from within the add-on. IP was correctly reset to my actual IP.

### DIFF
--- a/duckdns/data/run.sh
+++ b/duckdns/data/run.sh
@@ -10,7 +10,7 @@ LE_UPDATE="0"
 if bashio::config.has_value "ipv4"; then IPV4=$(bashio::config 'ipv4'); else IPV4=""; fi
 if bashio::config.has_value "ipv6"; then IPV6=$(bashio::config 'ipv6'); else IPV6=""; fi
 TOKEN=$(bashio::config 'token')
-DOMAINS=$(bashio::config 'domains | join(",")')
+DOMAINS=$(bashio::config 'domains | join(",")' | sed 's/[^,]*[ > ]//g') # Handle domains of the form "*.xxx.duckdns.org > xxx.duckdns.org"
 WAIT_TIME=$(bashio::config 'seconds')
 ALGO=$(bashio::config 'lets_encrypt.algo')
 


agners:
Is this still required since #3152 added DuckDNS support in the Let's Encrypt add-on?

From what I can see from https://github.com/home-assistant/addons/pull/2662#discussion_r1084602655 the preferred way is to use Let's Encrypt for these type of situation :thinking:

kyle-cackett:
Looks like #3152 was merged ~30 minutes ago so I haven't yet had a chance to evaluate whether switching add-ons and using the latest build of Let's Encrypt would be an alternative solution to the current limitation in the DuckDNS add-on described in this PR.

> Is this still required since https://github.com/home-assistant/addons/pull/3152 added DuckDNS support in the Let's Encrypt add-on?

Following this line of questioning, if DuckDNS support is now included in the Let's Encrypt add-on does that make the DuckDNS add-on redundant? If so perhaps the DuckDNS add-on should be deprecated and eventually discontinued entirely and users should be migrated to Let's Encrypt.

If the DuckDNS plugin is here to stay, including the fix proposed in this PR would be a nice since based on #336 some people are using this configuration which is causing some problems / frustrations.

Unrelated note: this fix could also be implemented without `sed` by using bash parameter expansion. E.g. `${DOMAINS//[^,]*[ > ]/}`

vdemidov:
> Is this still required since #3152 added DuckDNS support in the Let's Encrypt add-on?
I'm not sure. Will Let's Encrypt add-on support domains like *.xxx.duckdns.org? I will check after  Let's Encrypt add-on release. 
Anyway I will prefer one  DuckDNS add-on instead two addons.

iranl:
https://github.com/dehydrated-io/dehydrated/issues/554 and [the dehydrated troubleshooting guide](https://github.com/lukas2511/dehydrated/blob/master/docs/troubleshooting.md#dns-invalid-challenge-since-dehydrated-060--why-are-dns-challenges-deployed-first-and-verified-later) both specifically mention that API/DNS implementations such as employed by DuckDNS are considered broken and are not properly supported by dehyrated.

As such I consider the Let's encrypt implementation using dehydrated in the DuckDNS addon inherently broken and agree with @mdegat01 in https://github.com/home-assistant/addons/pull/2662#discussion_r1084602655 that support for aliases (including custom domains and wildcard DuckDNS domains) should be removed from the DuckDNS addon. 

I would even go one further and remove SSL support from the DuckDNS addon entirely and use the addon only for keeping a dynamic IP up to date, but maybe it's more sensible to keep the implementation only for the most basic of cases (only for xxx.duckdns.org certs).

I created #3152 in an effort to move DuckDNS DNS challenges to the Let's Encrypt addon as certbot and the [certbot_duckdns plugin](https://github.com/infinityofspace/certbot_dns_duckdns) properly support the (bad) API implementation at DuckDNS. It should support both domains such as (\*.)xxx.duckdns.org and (\*.)custom.com

Now that #3152 is merged maybe it's time to close this and reopen #2964?

PR #3248, opened 2023-10-06
Bump zwave-js to 12.2.0 and add config option for safe mode

raman325:
In some cases, users may need to adjust the length of timeouts or number of attempts configured in the driver. We therefore add a new hidden option to enable a safe mode, which, when enabled, will set some defaults for the `timeouts` and `attempts` options.

I have not validated these values with @AlCalzone yet

### DIFF
--- a/zwave_js/rootfs/etc/cont-init.d/config.sh
+++ b/zwave_js/rootfs/etc/cont-init.d/config.sh
@@ -84,7 +84,7 @@ done
 # flushed to disk
 if [[ ${flush_to_disk:+x} ]]; then
     bashio::log.info "Flushing config to disk due to creation of new key(s)..."
-    bashio::addon.options > "/data/options.json"
+    bashio::addon.options >"/data/options.json"
 fi
 
 s0_legacy=$(bashio::config "s0_legacy_key")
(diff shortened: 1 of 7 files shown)


AlCalzone (changes requested):

AlCalzone:
As discussed, the driver should expose some presets (including for the safe mode). I'll hold off on implementing that though until I have refined the [unresponsive controller recovery mechanism](https://github.com/zwave-js/node-zwave-js/issues/6402), which might require changes to the timeouts and presets.

MartinHjelmare:
Please undraft when the driver is ready and the PR is ready for review.

raman325:
everything appears to be working!

zwave_js/rootfs/usr/share/tempio/zwave_config.conf
```diff
@@ -14,5 +14,5 @@
         "S2_Authenticated": "{{ .s2_authenticated }}",
         "S2_Unauthenticated": "{{ .s2_unauthenticated }}"
     },
-    "enableSoftReset": {{ .soft_reset }}
+    "enableSoftReset": {{ .soft_reset }}{{ .presets }}
```
MartinHjelmare, line 17: Why do we concatenate the presets including the key instead of writing the key in the template? We know what they key is in advance, right?

raman325: we can change this, with the original implementation we didn't know which keys we were going to include hence this hack

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -119,13 +119,12 @@ else
     bashio::log.info "Soft-reset disabled by user"
 fi
 
-presets=""
+safe_mode=""
```
MartinHjelmare, line 122: I'd still use only one value for all preset values as we would need to build a formatted JSON list of all presets before updating the template.

MartinHjelmare (approved):

PR #3256, opened 2023-10-13
Bump zwave-js to 12.1.1

raman325:
Mostly config file changes although a bug fix for multi-target firmware updates which has been an issue for a while

- [Z-Wave JS 12.1.0](https://github.com/zwave-js/node-zwave-js/releases/tag/v12.1.0)
- [Z-Wave JS 12.1.1](https://github.com/zwave-js/node-zwave-js/releases/tag/v12.1.1)

CC @AlCalzone

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## 0.1.98
+
+### Bug fixes
+
+- Z-Wave JS: Fixed an issue with multi-target firmware updates that prevented updates from being applied correctly
+
+### Config file changes
+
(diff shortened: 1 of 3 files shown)


MartinHjelmare (approved):

AlCalzone:
Should we make sure the repair issues are skippable before merging this?

raman325:
they are confirmed not skippable if there is a repair flow. We can hold off on updating the add-on, however keep in mind that `zwave-js-ui` users are already seeing this.

AlCalzone:
I don't mind either way. Is the repair clear enough for users to understand that they don't have to do it?

raman325:
Yes. I'd suggest we merge if the firmware update issue is a big problem. Does it cause devices to brick?

raman325:
I'm going to go ahead and merge so that people can get the benefits of the config file updates and the firmware update fixes. There are enough users on zwave-js-ui that have not complained about that particular annoyance so I think it's ok to push forward.

AlCalzone:
The problematic firmware updates are disabled and will stay so until the driver version with the fix has significant usage.

PR #3258, opened 2023-10-13
Update README.md

DJJo14:
update url to lower case. github is case sensitive


frenck (approved):
Thanks, @DJJo14 👍 

../Frenck

PR #3270, opened 2023-10-19
fix typo openwakeword

chpego:
when you click on the link in the addon, you get an error on GitHub due to the name of the addon not being written in all lower case letters


frenck (approved):
Thanks, @chpego 👍 

../Frenck

PR #3273, opened 2023-10-20
Update Terminal & SSH Documentation to use ECDSA Key as RSA is no longer supported

MesserschmittX:
I had no success trying to get SSH to work with an RSA key like the linked documentation suggests. I found this [blog post](https://community.home-assistant.io/t/cannot-get-ssh-web-terminal-add-on-to-accept-authorized-keys/424600/2) which helped me get it to work with an ECDSA key.

### DIFF
--- a/ssh/DOCS.md
+++ b/ssh/DOCS.md
@@ -40,8 +40,7 @@ You can then connect to the port specified, using the username `root`. Please no
 
 If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is, therefore, generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
 
-> [!IMPORTANT]  
-> While following the instructions select ECDSA as **Type of key to generate** instead of RSA. RSA is no longer supported.
+> **Note**: While following the instructions above, select ECDSA as `Type of key to generate` instead of RSA. RSA is no longer supported.
 
 Enabling login via password will disable key-based login. You can not run both variants at the same time.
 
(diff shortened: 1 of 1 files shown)


ssh/DOCS.md
```diff
@@ -40,6 +40,9 @@ You can then connect to the port specified, using the username `root`. Please no
 
 If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is, therefore, generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
 
+> [!IMPORTANT]
```
frenck, line 43: This is GitHub specific syntax, which won't work inside Home Assistant itself.

frenck (changes requested):

ssh/DOCS.md
```diff
@@ -40,6 +40,8 @@ You can then connect to the port specified, using the username `root`. Please no
 
 If you enable connecting to the SSH Server using an SSH client, you are strongly recommended to use private/public keys to log in. As long as you keep the private part of your key safe, this makes your system much harder to break into. Using passwords is, therefore, generally considered a less secure mechanism. To generate private/public SSH keys, follow the [instructions for Windows][keygen-windows] and [these for other platforms][keygen].
 
+> **Note**: While following the instructions above, select ECDSA as `Type of key to generate` instead of RSA. RSA is no longer supported.
```
agners, line 43: I don't think that a block quote (`> `) is appropriate here, so:

```suggestion
**Note**: While following the instructions above, select ECDSA as `Type of key to generate` instead of RSA. RSA is no longer supported.
```

MesserschmittX: As this is basically the end of the instructions and it tells you to go to another page right before, I felt the need to emphasize it a bit more than just plain text. I just did this, and I, for one, went straight to the "instructions for Windows" instead of reading further. I also don't actually care how exactly this looks, I just hoped it would help others save time in finding this.

agners: Yeah but block-quote has a meaning, and it is not the right way to emphasize something. IMHO, the note is just bellow the link, and with it being bold it should be emphasized enough. Your change is already a big help for users :+1:

agners (approved):

frenck (approved):
Looks good to me as well 👍

PR #3274, opened 2023-10-21
nginx_proxy: Adding support for TCP Proxy Protocol

miguelrjim:
This allows Nginx to support the [TCP proxy protocol](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/), this is useful if you still want to TLS terminate on the home-assistant host but you have another entrypoint for all your TLS connections (i.e. traefik is your entrypoint and you redirect from it to home-assistant based on the HTTPS domain being requested, this makes sure that any kind of network request to home-assistant remains encrypted).

What this allows is that home-assistant will know the true ip of the user making requests (making this change because I saw in the logs that there was a failed log-in but the ip pointed to my traefik instance).

Tested this by installing these changes as a local add-on on home-assistant.

## Summary by CodeRabbit

- **New Features**
	- Introduced support for TCP Proxy Protocol, improving handling of TCP traffic.
	- Added `real_ip_from` configuration option, allowing specification of trusted IP addresses for real IP forwarding.

- **Bug Fixes**
	- Enhanced accuracy in logging client IPs through updated proxy settings.

- **Documentation**
	- Updated documentation to include new configuration options and their usage.

- **Chores**
	- Incremented version number to 3.10.0 and improved configuration management structure for maintainability.


bvli:
I'd really like to see this PR merged  👍

miguelrjim:
Hi @agners, just wondering what would be the next step to get the PR merged?

nginx_proxy/config.yaml
```diff
@@ -26,16 +26,23 @@ options:
     active: false
     default: nginx_proxy_default*.conf
     servers: nginx_proxy/*.conf
+  proxyProtocol:
```
agners, line 29: Multi-word configs in other add-ons usually use Python style naming with underlines, so:

```suggestion
  proxy_protocol:
```

agners (changes requested):
I am not very familiar with the various proxy options/configurations, but wouldn't the first option documented in https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/ to get the originating IP address more generic? (E.g. not using the RealIP module? :thinking: ).

nginx_proxy/config.yaml
```diff
@@ -26,16 +26,23 @@ options:
...
   cloudflare: bool
   customize:
     active: bool
-    default: str
-    servers: str
+    default: str?
+    servers: str?
+  proxyProtocol:
+    enabled: bool
+    realIpFrom: str?
```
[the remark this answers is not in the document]
miguelrjim: What field were you thinking?

miguelrjim:
> I am not very familiar with the various proxy options/configurations, but wouldn't the first option documented in https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/ to get the originating IP address more generic? (E.g. not using the RealIP module? 🤔 ).

Hmm it could be? Tbh i've been using [$proxy_protocol_addr](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_proxy_protocol_addr) as that was being used in a couple of other examples with proxy protocol.

nginx_proxy/config.yaml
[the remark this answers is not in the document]
agners: You can ask the user for a list, the Samba add-on has an example for it's allowed host list, syntax would be:

```
  allow_hosts:
    - str
```

I am not sure how to make this optional though. :sweat_smile:

agners:
> Hmm it could be? Tbh i've been using [$proxy_protocol_addr](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_proxy_protocol_addr) as that was being used in a couple of other examples with proxy protocol.

There is also this note:

> The Real‑IP modules for [HTTP](https://nginx.org/en/docs/http/ngx_http_realip_module.html) and [Stream TCP](https://nginx.org/en/docs/stream/ngx_stream_realip_module.html) are not included in NGINX Open Source by default; see [Installing NGINX Open Source](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/) for details. No extra steps are required for NGINX Plus.

:thinking: 

Again, I am really not an expert in NGINX/Proxy setups. But I think if things can work without the RealIP module, it seems the better approach to me.

nginx_proxy/config.yaml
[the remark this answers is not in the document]
miguelrjim: There's definitely some improvements that can be done into how this addon is setup; will definitely take a look later to see about these other kind of changes as well.

nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/nginx/run
```diff
@@ -56,6 +56,23 @@ sed -i "s/%%HA_PORT%%/$HA_PORT/g" /etc/nginx.conf
 [ -n "$HSTS" ] && HSTS="add_header Strict-Transport-Security \"$HSTS\" always;"
 sed -i "s/%%HSTS%%/$HSTS/g" /etc/nginx.conf
 
+if bashio::config.true 'proxy_protocol.enabled'; then
+    REAL_IP_FROM=$(bashio::config 'proxy_protocol.real_ip_from')
+    if [ -z "$REAL_IP_FROM" ]; then
+        echo "real_ip_from under proxy_protocol has to be provided"
+        exit 1
+    fi
+    sed -i "s#%%REAL_IP_FROM%%#$REAL_IP_FROM#g" /etc/nginx/serverProxyProtocol.conf
```
agners, line 65: It seems that camel case is not common for nginx config files. Also, includes are typically stored in `conf.d` sub directory (see https://docs.nginx.com/nginx/admin-guide/basic-functionality/managing-configuration-files/). Can you rename these include files and move them to `conf.d` directory?

agners (changes requested):

nginx_proxy/config.yaml
```diff
@@ -37,5 +40,9 @@ schema:
...
   customize:
     active: bool
-    default: str
-    servers: str
+    default: str?
+    servers: str?
+  proxy_protocol:
+    enabled: bool
+    real_ip_from: str?
+
```
agners, line 48:
```suggestion
```

miguelrjim:
@agners just wondering if we're good to merge the pr after your comments had been taken care of

nginx_proxy/config.yaml
```diff
@@ -37,5 +40,8 @@ schema:
   cloudflare: bool
   customize:
     active: bool
-    default: str
-    servers: str
+    default: str?
+    servers: str?
```
agners, line 44: These should remain mandatory. If a default value is set, they automatically become mandatory settings. The user doesn't need to provide them, since there is a default, but from a config perspective, they are mandatory.

```suggestion
    default: str
    servers: str
```

The docs are a bit unclear, but I tested the behavior (and also extended the docs, see https://github.com/home-assistant/developers.home-assistant/pull/2253)

nginx_proxy/config.yaml
```diff
@@ -26,6 +26,9 @@ options:
     active: false
     default: nginx_proxy_default*.conf
     servers: nginx_proxy/*.conf
+  proxy_protocol:
+    enabled: false
+    real_ip_from: null
```
agners, line 31: Since `real_ip_from` is optional, there must be no default value, otherwise it comes mandatory again. With this current config, the user is requested to set a `real_ip_from` even when `enabled: false` is set :see_no_evil: 

```suggestion
```

agners (changes requested):
Sorry for the late review :cry: 

I actually wanted to test this, and I finally came around. During testing, I realized one thing: Instead of using the sub config we could just promote the `real_ip_from` as top level config. The sub-config makes sense if there are multiple configs underneath. The enable property made sense in the past when there were no optional configs.

But today, we have optional configs, so promoting `real_ip_from` to top level, we can just check if it is set, and if so enable the proxy protocol. Use `bashio::config.exists` to check if the user has set a value:

```
if bashio::config.exists 'real_ip_from'; then
```

The current approach if the user sets proxy protocol to enable, the add-on will only tell at runtime through logs that `real_ip_from` is missing. With the single config approach, all requirements are checked at config time. So I'd prefer this approach for that reason.

If you prefer the current approach, that is fine too, but we have to remove the default value of `real_ip_from`.

Also, in any case the top-level config needs translation strings. So please also add name/description strings for the top-level config in `translations/en.yaml`.

miguelrjim:
@agners No worries, thanks for taking your time reviewing it!

I've included your suggested changes, I agree this makes it easier and makes sense! I've tested the latest changes locally as well and it's working as intended.

nginx_proxy/config.yaml
```diff
@@ -39,3 +39,4 @@ schema:
     active: bool
     default: str
     servers: str
+  real_ip_from: str?
```
agners, line 42: I see that the example docs has multiple IP addresses: http://nginx.org/en/docs/http/ngx_http_realip_module.html

Can we maybe make a list, essentially revisit what I've suggested [here](https://github.com/home-assistant/addons/pull/3274#discussion_r1520945900)?

I don't think it can be optional then, but we can consider empty list as not enabled.

nginx_proxy/translations/en.yaml
```diff
@@ -24,6 +24,11 @@ configuration:
     name: Customize
     description: >-
       See the Documentation tab for more information about these options.
+  real_ip_from:
+    name: RealIpFrom
```
agners, line 28: This can have spaces. And maybe give a hint what the consequences are in the name of the property already :thinking: 
```suggestion
    name: Real IP from (enables PROXY protocol)
```

agners (changes requested):
Thanks for the update. This looks quite good, however I have yet another review/request :see_no_evil:

nginx_proxy/config.yaml
miguelrjim: Updated the PR to handle a list; I switched to use tempio to make it easier, I'll write a PR later to update this addon to use templating for most of the other operations we're currently doing in bash directly.

miguelrjim: Tested this changes when there are IPs specified, and when the list is empty; both are working as expected.

miguelrjim:
@agners Made more substantial changes to move replacement logic by taking advantage of templating; tested the changes locally by specifying different values in the ui options (changing the boolean values & text inputs) and the generated `nginx.conf` was as expected; nginx started successfully and was processing requests as intended, currently using this latest version directly in my home assistant instance.

agners (approved):
LGTM, thanks for the contribution and your patience!

What would be nice if this add-on makes use of the new add-on config directory instead of share for the configuration. But definitely something for another day :smile:

mpreitz:
This feature breaks the funcionality of the plugin. Setting a "real_ip_from" is mandantory, otherwise the plugin won't start. If setting the local network as allowed the plugin starts, but no connection is possible.

Would be nice to read a word in the changelog if breaking changes are part of the release.

agners:
Making `real_ip_from` mandatory was not really intended :cry: 

This should address the issue: https://github.com/home-assistant/addons/pull/3725.

mad-tunes:
Might this be related to some having trouble getting to HA after updating to 3.10.0?
https://discord.com/channels/330944238910963714/672220642099200000/1273278057511387169
Thanks

mad-tunes:
It seems that adding real_ip_from: [] to settings yaml, or setting ' ' or '*' in the GUI and then removing it, is letting people at HA thorugh nginx with 3.10.0

miguelrjim:
Sorry! This was definitely not intended to be a breaking change, thanks for the quick fix @agners

PR #3282, opened 2023-10-25
matter_server: Add beta flag

agners:
Add a new flag which allows to install the latest pre-release of the Python Matter Server. This is useful for testing new (pre-)releases before bumping the add-on.

Previously we tried different approaches like a separate beta version of the add-on. This has the problem that it has it's separate data store (means no more devices in case the non-beta add-on has been used previously) and it gets chaotic as the Matter integration is still trying to start the non-beta version of the add-on.

With this attempt we can reuse the same add-on for testing. The downside is once new OS dependencies are required, those won't be added. In that case a regular add-on version bump would be necessary first.


frenck (approved):

matter_server/config.yaml
```diff
@@ -20,8 +20,10 @@ image: homeassistant/{arch}-addon-matter-server
 init: false
 options:
   log_level: info
+  beta: false
 schema:
   log_level: list(debug|info|warning|error|critical)
+  beta: bool
```
MartinHjelmare, line 26: Make sure to make it optional.

frenck:
```suggestion
  beta: bool?
```

PR #3286, opened 2023-10-28
Document that symlinks may not work

marcin-osowski:
I've spent some time trying to set the reverse proxy up. Symlinks don't seem to work here (perhaps docker blocks it?), yet symlinks are commonly used to manage certificates.

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -12,7 +12,7 @@ Follow these steps to get the add-on installed on your system:
 
 The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario.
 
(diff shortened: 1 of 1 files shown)


agners:
Yes this is definitely caused by Docker, where a bind mount is used. A symlink to another place on the host system isn't valid inside the add-on.

In a full Home Assistant OS the certificates are managed by the Let's Encrypt add-on. For Supervised installations deploying certificates can be done that way, but we expect that this type of quirks are known to the user. Otherwise we have to document that in many other places as well (e.g. /config directory).

Maybe instead of documenting (which often doesn't get read anyways :cry: ), pasting the error you got from the add-on log when using the symlink here in this thread helps people find the issue quickly?

marcin-osowski:
I figured it out by guessing that this may be the case, and installing the certificate and key as regular files in `/ssl`, then observing that :443 serves the right cert. There was no suggestive error in the logs, at least I didn't find any. I think it just said that `open()` failed.

PR #3304, opened 2023-11-11
ssh: Add config symlink for docs and backwards compat

frenck:
This adds a symlink from `/homeassistant` to `/config` and also links it as a common folder.

This maintains backward compatibility, matches our docs, and doesn't interfere with the muscle memory of a user.


frenck:
No longer needed.

pvizeli (approved):

PR #3316, opened 2023-11-16
Add support for addon_configs to zwave_js

raman325:
This will let users:
1. Access their cache files
2. ~~Use custom device config files~~ Removed
3. Optionally log to file (and specify the max number of files/days of logs to keep)

For the last one, we can opt to make this enabled by default with no configuration option, but I like the idea of giving the user the option so that they can avoid unnecessary I/O if they don't plan to access the log files.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Add-On: Allow custom device config files to be manually installed in `/addon_configs/core_zwave_js/custom_device_configs`
 - Add-On: Provide access to Z-Wave JS cache files for debugging in `/addon_configs/core_zwave_js/cache`
-- Add-On: Add configuration option to log to file. When enabled, logs will be written to `/addon_configs/core_zwave_js/zwave_js.log`
+- Add-On: Add configuration option to log to file. When enabled, logs will be written to `/addon_configs/core_zwave_js` with the `.log` file extension
 
 ## 0.3.0
 
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.3.1
+
+### Features
+
+- Add-On: Allow custom device config files to be manually installed in `/addon_configs/core_zwave_js/custom_device_configs`
+- Add-On: Provide access to Z-Wave JS cache files for debugging in `/addon_configs/core_zwave_js/cache`
+- Add-On: Add configuration option to log to file. When enabled, logs will be written to `/addon_configs/core_zwave_js/zwave_js.log`
+
 ## 0.3.0
 
 ### Features
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.3.1
+## 0.4.0
 
 ### Features
 
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 ### Features
 
-- Add-On: Allow custom device config files to be manually installed in `/addon_configs/core_zwave_js/custom_device_configs`
 - Add-On: Provide access to Z-Wave JS cache files for debugging in `/addon_configs/core_zwave_js/cache`
 - Add-On: Add configuration option to log to file. When enabled, logs will be written to `/addon_configs/core_zwave_js` with the `.log` file extension
 
--- a/zwave_js/DOCS.md
+++ b/zwave_js/DOCS.md
@@ -140,8 +140,8 @@ the Supervisor.
 
 ### Optional `log_to_file`
 
-When this option is enabled, logs will be written to the `/addon_configs/core_zwave_js` folder.
-Log filenames will begin with `zwave_` and will end with `.log`.
+When this option is enabled, logs will be written to the `/addon_configs/core_zwave_js`
+folder with the `.log` file extension.
 
 ### Option `soft_reset`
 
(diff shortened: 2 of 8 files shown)


zwave_js/DOCS.md
```diff
@@ -176,6 +181,12 @@ the introduction of S2 security inclusion in zwave-js, this option has been
 deprecated in favor of `s0_legacy_key`. If still set, the `network_key` value will be
 migrated to `s0_legacy_key` on first startup.
 
+## Custom device config files
```
MartinHjelmare, line 184: I'm not in favor of adding this feature. The risk of introducing bugs that we'll waste time to troubleshoot in support is high. We've already seen that happen for users using Z-Wave JS UI. Why do you need custom device config?

MartinHjelmare (dismissed):

zwave_js/DOCS.md
raman325: if a user wants to add support for a device that doesn't have a config file, they would need something like this to do so. Same for if there is a mistake with the existing config file and they need to test a fix.

With that being said, definitely understand the concern about increasing support costs and introducing bugs that would be hard to find on our side.

Since device config files primarily affect:
1. manufacturer/product name
2. Manuals/reset instructions
3. Config parameters

Would those problem spaces still be a concern from a troubleshooting perspective?

AlCalzone: FWIW driver logs tell if a user-provided config file was loaded. That requires the log to include the startup sequence though, which is kinda difficult to capture in HA currently.

raman325: Being able to log to file will help with that

raman325: There is also the `isEmbedded` property of the device config which would be exposed in the device diagnostics. We could also consider adding an alert to the device info page when a user is using a user provided config (e.g. `node.is_embedded == False`)

raman325: thinking more about this I think adding an alert for this to the device info page is a good idea in general, thoughts?

MartinHjelmare: If you aren't able to set up a development environment I don't think you should change your device config either.

raman325: I personally think the concern is more than the actual risk, but this should be a feature that is rarely used, so I am OK to remove it 👍🏾

zwave_js/DOCS.md
```diff
@@ -138,6 +138,17 @@ This option sets the log level of Z-Wave JS. Valid options are:
 If no `log_level` is specified, the log level will be set to the level set in
 the Supervisor.
 
+### Option `log_to_file`
+
+When this option is enabled, logs will be written to the `/addon_configs/core_zwave_js`
```
MartinHjelmare, line 143: How can the user access this folder?

zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.3.1
+
+### Features
+
+- Add-On: Provide access to Z-Wave JS cache files for debugging in `/addon_configs/core_zwave_js/cache`
```
MartinHjelmare, line 7: The cache files aren't documented. Should we do that?

zwave_js/DOCS.md
raman325: shouldn't it just be available? https://developers.home-assistant.io/blog/2023/11/06/public-addon-config/

zwave_js/CHANGELOG.md
raman325: another thing to help with troubleshooting. We can remove it if desired

zwave_js/DOCS.md
raman325: shouldn't it just be available? https://developers.home-assistant.io/blog/2023/11/06/public-addon-config/

MartinHjelmare: Well, the user will typically need another add-on to access files, like the file editor, samba, or ssh add-ons. There's no user interface in this add-on to access those files. Do we need to explain that?

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -145,13 +145,18 @@ else
...
+log_max_files=$(bashio::config "log_max_files")
+
 # Generate config
 bashio::var.json \
     s0_legacy "${s0_legacy}" \
     s2_access_control "${s2_access_control}" \
     s2_authenticated "${s2_authenticated}" \
     s2_unauthenticated "${s2_unauthenticated}" \
     log_level "${log_level}" \
+    log_to_file "${log_to_file}" \
```
MartinHjelmare, line 158: Is the log to file option in zwave-js compatible with the other log transports that we use? It sounds like it should be as the docs use the phrase "in addition":

https://zwave-js.github.io/node-zwave-js/#/api/driver?id=logconfig

zwave_js/CHANGELOG.md
MartinHjelmare: I mean should we document it?

zwave_js/rootfs/etc/cont-init.d/config.sh
raman325: Ah I misread, yes let me update the docs good callout

MartinHjelmare: Have we confirmed that logging to file is compatible with the other log transports the server uses?

raman325: I tested the UI functionality in HA and it works.

When we set `logToFile`, we do it when creating the driver. The server only starts after the driver has started, and when we need to use our own log transport, we push it into the transports list which would preserve the existing transports, which in this case is the file transport. So I don't see how the two could come in conflict

MartinHjelmare: Ok.

zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.3.1
```
MartinHjelmare, line 3: I'd bump minor version since we're adding new features.

MartinHjelmare (approved):
Thanks!

PR #3312, opened 2023-11-15
samba: Add support for accessing add-on configs

frenck:
## 12.1.0

- Use the new Home Assistant folder for the `config` share
- Add support for accessing public add-on configurations

PR #3322, opened 2023-11-20
samba: Add documentation for addon_configs

frenck:
SSIA

As mentioned here:

<https://github.com/home-assistant/addons/pull/3312#issuecomment-1813327616>

### DIFF
--- a/samba/DOCS.md
+++ b/samba/DOCS.md
@@ -25,7 +25,7 @@ This addon exposes the following directories over smb (samba):
 Directory | Description
 -- | --
 `addons` | This is for your local add-ons.
-`addon_configs` | This is for your the configuration files or your add-ons.
+`addon_configs` | This is for the configuration files of your add-ons.
 `backup` | This is for your backups.
 `config` | This is for your Home Assistant configuration.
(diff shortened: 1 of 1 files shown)


samba/DOCS.md
```diff
@@ -25,6 +25,7 @@ This addon exposes the following directories over smb (samba):
 Directory | Description
 -- | --
 `addons` | This is for your local add-ons.
+`addon_configs` | This is for your the configuration files or your add-ons.
```
agners, line 28:
```suggestion
`addon_configs` | This is for your add-on's configuration files.
```

frenck: That implies singular, which it isn't?

agners: Right, I think this should do it.

```suggestion
`addon_configs` | This is for your add-ons' configuration files.
```

agners: Or I guess the of form which might be nicer? Either way works for me.

```suggestion
`addon_configs` | This is for the configuration files of your add-ons.
```

agners (approved):

PR #3324, opened 2023-11-21
Update Certbot Direct Admin

InhouseMedia:
Direct Admin DNS is not working because of the pre-release version that was still used.

Fixed issue with directadmin_name being used before has been allocated

https://github.com/cybercinch/certbot-dns-directadmin/releases


agners:
Closing as it is superseded by #3325.

PR #3325, opened 2023-11-22
letsencrypt: Update Alpine, Python and Certbot

agners:
Update to Alpine 3.18 and Python version 3.11. Also update Certbot to the latest release 2.7.4 along with all DNS authenticator plug-ins.


letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
```diff
@@ -6,9 +6,7 @@ mkdir -p /data/workdir
 mkdir -p /data/letsencrypt
 
 # Setup Let's encrypt config
-echo -e "dns_cloudxns_api_key = $(bashio::config 'dns.cloudxns_api_key')\n" \
-      "dns_cloudxns_secret_key = $(bashio::config 'dns.cloudxns_secret_key')\n" \
-      "dns_digitalocean_token = $(bashio::config 'dns.digitalocean_token')\n" \
+echo -e "dns_digitalocean_token = $(bashio::config 'dns.digitalocean_token')\n" \
       "certbot_dns_directadmin:directadmin_url = $(bashio::config 'dns.directadmin_url')\n" \
```
sairon, line 10: Shouldn't this and the following also be changed from the `dist_name:plugin_name` just to `plugin_name`?

sairon (commented):
Looks good, checked the Certbot release notes and haven't found any gotcha. Just one thing in comments.

letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
agners: Good catch! There are a couple of old style configs still, I've updated them all now :+1:

sairon (approved):
Looks good now :+1:

frenck (approved):

Issue #3335, opened 2023-11-26
Let's Encrypt: dns_directadmin_url is required error


defaultsecurity:
### Describe the issue you are experiencing

The addon won't start anymore. The log says "dns_directadmin_url is required" The parameter is correctly set in the config. The addons was working before the update to 5.0.1.

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.0.1

### Steps to reproduce the issue

1. Start addon
2. Addons stops

### System Health information

-

### Anything in the Supervisor logs that might be useful for us?

```txt
-
```

### Anything in the add-on logs that might be useful for us?

```txt
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
cont-init: info: running /etc/cont-init.d/file-structure.sh
cont-init: info: /etc/cont-init.d/file-structure.sh exited 0
s6-rc: info: service legacy-cont-init successfully started
s6-rc: info: service legacy-services: starting
services-up: info: copying legacy longrun lets-encrypt (no readiness notification)
s6-rc: info: service legacy-services successfully started
[23:21:06] INFO: Selected DNS Provider: dns-directadmin
[23:21:06] INFO: Use propagation seconds: 60
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Renewing an existing certificate for ****.*********.***
{}: dns_directadmin_url is required
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
s6-rc: info: service legacy-services: stopping
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service legacy-cont-init: stopping
s6-rc: info: service legacy-cont-init successfully stopped
s6-rc: info: service fix-attrs: stopping
s6-rc: info: service fix-attrs successfully stopped
s6-rc: info: service s6rc-oneshot-runner: stopping
s6-rc: info: service s6rc-oneshot-runner successfully stopped
```

### Additional information

Config:

```
domains:
  - ****.******.***
email: *****@*******.***
keyfile: privkey.pem
certfile: fullchain.pem
[...]

PR #3336, opened 2023-11-27
letsencrypt: Fix DirectAdmin DNS challenge support

agners:
It seems that the DirectAdmin Certbot plug-in does not use the "dns-" suffix like other DNS based challenge plug-ins use. Make sure the add-on config remains backwards compatible and just pass the correct arguments to use the DirectAdmin plug-in with Certbot.

Fixes: #3335

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.0.2
+
+- Fix DirectAdmin DNS challenge support
+
 ## 5.0.1
 
 - Add DuckDNS DNS challenge support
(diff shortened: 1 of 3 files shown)


Veine-Hagen (approved):

0xFelix:
Looks to be still broken in version `5.0.3` for me. Any ideas?

```
[09:49:49] INFO: Selected DNS Provider: dns-directadmin
[09:49:49] INFO: Use propagation seconds: 60
usage: 

certificate. 
certbot: error: unrecognized arguments: --directadmin-credentials /data/dnsapikey --directadmin-propagation-seconds 60
```

agners:
Uh I see, testing manually turns out that the documentation at https://certbot-dns-directadmin.readthedocs.io/en/latest/ is actually wrong :see_no_evil: 

I'll try to fix it in another attempt.

PR #3340, opened 2023-12-05
Fix DirectAdmin DNS challenge support by using dns prefix

agners:
It seems that the current DirectAdmin certbot plug-in documentation is a bit outdated. The options as well as the configuration require the "dns" prefix just like any other plug-ins.

Fixes: #3336


frenck (approved):
Thanks, @agners 👍 

../Frenck

Issue #3132, opened 2023-07-13
nginx_proxy: Host header causes problems with ESPHome addon


mancontr:
### Describe the issue you are experiencing

As discussed [here](https://community.home-assistant.io/t/esphome-web-ui-black-screen-for-validate-install-and-log/337578/16), when using the NGINX SSL Proxy addon together with the ESPHome addon, the websocket requests fail with this error:
```Cross origin websockets not allowed```

The reason is that the Origin and Host field differ, because the former includes the port, but the latter doesn't.

The fix would be very simple: on the default config, replace $host with $http_host, so it includes the port, on the `proxy_set_header Host` [rule](https://github.com/home-assistant/addons/blob/master/nginx_proxy/rootfs/etc/nginx.conf#L63C13-L63C34).

I tried this fix and worked correctly for me, and didn't seem to break any of my other addons, but further testing would be nice.

I can send a PR if this fix looks reasonable.

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Debian

### Which add-on are you reporting an issue with?

NGINX Home Assistant SSL proxy

### What is the version of the add-on?

3.5.0

### Steps to reproduce the issue

1. Configure the NGINX and ESPHome addons
2. Access the ESPHome page through ingress, and use the "validate" button
3. The screen stays black, and an error is reported in the supervisor logs

## System Information

version | core-2023.7.1
-- | --
installation_type | Home Assistant Supervised
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.4
os_name | Linux
os_version | 5.10.0-22-amd64
arch | x86_64
timezone | Europe/Madrid
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.32.1
Stage | running
Available Repositories | 1272
Downloaded Repositories | 6

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Debian GNU/Linux 11 (bullseye)
-- | --
update_channel | stable
supervisor_version | supervisor-2023.07.1
agent_version | 1.5.1
docker_version | 23.0.6
disk_total | 217.6 GB
[...]

PR #3356, opened 2023-12-13
nginx_proxy: Fix issue #3132

mancontr:
Issue #3132, affecting nginx_proxy, has been causing other addons to break for a long time. For example, part of the esphome addon is not usable through the nginx_proxy because of this bug. All the details are in the issue.

The fix is very simple: replace $host with $http_host, so the port gets included and the socket connection is not detected as cross-origin.

I've had this fix applied on a custom config since the issue was opened, without any problems, but I think it would be nice to fix it here for everyone.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.6.0
+
+- Add port to Host header to fix origin issues affecting ESPHome and other addons
+
 ## 3.5.0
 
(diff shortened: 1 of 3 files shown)


agners:
Nice catch!

Without understanding too much here, from this [StackOverflow thread](https://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx?rq=3) it seems that another difference is that $http_host is from the client request. Can it be trusted? :thinking:

agners:
Please also bump the version in config.yaml and add your change to `CHANGELOG.md`.

PR #3357, opened 2023-12-14
openthread_border_router: Add NAT64 and TREL support

agners:
Add NAT64 network address translation support to allow Thread devices connect to IPv4 hosts. This is optional and disabled by default.

Also add TREL (Thread Radio Encapsulation Link) support which essentially allows two Border Router to forward Thread packets through the backbone (WiFi/Ethernet) link instead of through the radio network. This can be beneficial especially in larger setups with multiple Border Routers.

Also update OTBR to the latest upstream version and some smaller cleanups.


openthread_border_router/Dockerfile
```diff
@@ -31,18 +30,19 @@ ENV OTBR_UNUSED_DEBS libavahi-client3 avahi-daemon rsyslog
...
     && apt-get update \
     && apt-get install -y --no-install-recommends \
-       git \
        iproute2 \
        python3 \
        python3-pip \
        lsb-release \
        netcat-openbsd \
        sudo \
+       git \
```
marcelveldt, line 45: is there a reason this is no longer alphabetically sorted ?

openthread_border_router/Dockerfile
```diff
@@ -63,25 +63,30 @@ RUN \
...
         -DOT_LINK_RAW=1 \
         -DOTBR_VENDOR_NAME="Home Assistant" \
         -DOTBR_PRODUCT_NAME="OpenThread Border Router" \
         -DOTBR_WEB=ON \
         -DOTBR_BORDER_ROUTING=ON \
         -DOTBR_REST=ON \
         -DOTBR_BACKBONE_ROUTER=ON \
+        -DOTBR_TREL=ON \
+        -DOTBR_NAT64=ON \
+        -DOT_POSIX_NAT64_CIDR="192.168.255.0/24" \
```
marcelveldt, line 80: should this be configurable if users running on the 10.x private subnet ? Or is that unrelated ?

openthread_border_router/Dockerfile
```diff
@@ -63,25 +63,30 @@ RUN \
...
+        -DOTBR_DNS_UPSTREAM_QUERY=ON \
         && cd build/otbr/ \
         && ninja \
         && ninja install) \
-    && pip install universal-silabs-flasher==${UNIVERSAL_SILABS_FLASHER} \
+    && pip install gpiod==1.5.4 universal-silabs-flasher==${UNIVERSAL_SILABS_FLASHER} \
```
marcelveldt, line 85: is this here by accident ?

openthread_border_router/rootfs/etc/s6-overlay/scripts/otbr-agent-configure.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/with-contenv bashio
+# ==============================================================================
+# Configure OTBR depending on add-on settings
+# ==============================================================================
+
+if bashio::config.true 'nat64'; then
+    bashio::log.info "Enabling NAT64."
+    ot-ctl nat64 enable
```
marcelveldt, line 8: long term we maybe want to finetune firewall rules to determine which device can do outbound communications ?

openthread_border_router/Dockerfile
agners: No, it is a temporary work around. See https://github.com/home-assistant/addons/pull/3357/commits/210baf31ebc61468455c3013bc077c07cf0aab14.

openthread_border_router/rootfs/etc/s6-overlay/scripts/otbr-agent-configure.sh
agners: Uh sure, why not :sweat_smile: 

But I don't think this fits into the add-on configuration system. It probably should go through the REST API. Ideally we'd add support for such a feature in upstream OTBR.

openthread_border_router/Dockerfile
agners: It should not matter if you use a 10.x.x.x private subnet. It could be possible problematic when using 192.168.255.x in a network, maybe :thinking: FWIW, the value is just the default value of OTBR. If I don't configure it explicitly here, the build system would use the same value implicitly.

It is possible to change the setting at runtime, see https://github.com/openthread/openthread/blob/main/src/cli/README.md#nat64-cidr. 

However, I am not sure if it is necessary at this point. I'd suggest that we leave that static default currently and see if someone runs into problems.

PR #3362, opened 2023-12-16
Add configuration instructions to README.md

AJediIAm:
Include instructions on the parameters and how to select the right URL


agners (changes requested):
The configuration is already documented largely in `DOCS.md` (which is shown in the documentation tab in Supervisor).

Can you move the additional notes about firmware source/URL there too? Thanks!

Issue #3355, opened 2023-12-13
Whisper can not be run


proasnet:
### Describe the issue you are experiencing

After start Whisper on x86 machine, Intel CPU, 3GB RAM, HASOS, I getting in cyclic
INFO: Service exited with code 256 (by signal 4)

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Whisper

### What is the version of the add-on?

1.0.0

### Steps to reproduce the issue

1. Start addon
2.
3.
...

## System Information

version | core-2023.12.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.63-haos
arch | x86_64
timezone | Europe/Prague
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4995
Installed Version | 1.33.0
Stage | running
Available Repositories | 1419
Downloaded Repositories | 5

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 11.2
-- | --
update_channel | stable
supervisor_version | supervisor-2023.11.6
agent_version | 1.6.0
docker_version | 24.0.7
disk_total | 116.7 GB
disk_used | 39.6 GB
healthy | true
supported | true
board | generic-x86-64
supervisor_api | ok
version_api | ok
installed_addons | File editor (5.7.0), Terminal & SSH (9.8.1), Mosquitto broker (6.4.0), Samba share (12.2.0), RPC Shutdown (2.4), MariaDB (2.6.1), phpMyAdmin (0.9.0), Paradox Alarm Interface (3.4.0), ESPHome (2023.9.3), MQTT Explorer (browser-1.0.1), Samba Backup (5.2.0), Piper (1.4.0), openWakeWord (1.8.2), Whisper (1.0.0)

</details>

<details><summary>Dashboards</summary>

dashboards | 9
-- | --
resources | 2
views | 14
mode | storage

</details>

<details><summary>Recorder</summary>

[...]

Issue #3035, opened 2023-05-09
Whisper container not working: Service exited with code 256 (by signal 4)


agittins:
### Describe the issue you are experiencing

(Note: whisper doesn't appear in the add-on list in the bug report form, I hope I'm reporting in the right place)

Whisper does not appear via wyoming autodiscovery (piper has shown up OK), and logs show the service inside the whisper container repeatedly exiting with "INFO: Service exited with code 256 (by signal 4)"

After starting the whisper container in add-ons, the logs show:
```
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
s6-rc: info: service legacy-cont-init successfully started
s6-rc: info: service whisper: starting
s6-rc: info: service whisper successfully started
s6-rc: info: service discovery: starting
[18:51:16] INFO: Service exited with code 256 (by signal 4)
[18:51:18] INFO: Service exited with code 256 (by signal 4)
[18:51:20] INFO: Service exited with code 256 (by signal 4)
[18:51:22] INFO: Service exited with code 256 (by signal 4)
```

The Service exited line repeats infinitum (for well over 24 hours, anyway).

On the vm running HAOS, the last few lines of `journalctl -e` show:
```
May 09 08:55:00 homeassistant addon_core_whisper[463]: [18:55:00] INFO: Service exited with code 256 (by signal 4)
May 09 08:55:02 homeassistant audit[1799253]: ANOM_ABEND auid=4294967295 uid=0 gid=0 ses=4294967295 subj=docker-default pid=1799253 comm="python3" exe="/usr/bin/python3.9" sig=4 res=1
May 09 08:55:02 homeassistant audit: BPF prog-id=133583 op=LOAD
May 09 08:55:02 homeassistant audit: BPF prog-id=133584 op=LOAD
May 09 08:55:02 homeassistant audit: BPF prog-id=133585 op=LOAD
May 09 08:55:02 homeassistant systemd[1]: Started Process Core Dump (PID 1799274/UID 0).
May 09 08:55:02 homeassistant systemd-coredump[1799275]: Process 1799253 (python3) of user 0 dumped core.
May 09 08:55:02 homeassistant systemd[1]: systemd-coredump@44474-1799274-0.service: Deactivated successfully.
[...]

PR #3368, opened 2023-12-20
whisper: Exit with error on unsupported CPU (missing AVX)

agners:
As discussed in https://github.com/home-assistant/addons/issues/3035 CTranslate2 requires the AVX extension. Check if the extension is supported by the CPU before running the add-on. This should give the user a better clue as to why the add-on is not working.

Refers to: #3035, #3355

### DIFF
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.1
+
+- Handle unsupported CPU configurations
+
 ## 1.0.0
 
 - Upgrade to `wyoming-whisper` 1.0.1
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.0.0
+version: 1.0.1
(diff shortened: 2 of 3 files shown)


sairon (commented):
Looks good to me from the technical perspective. Just an ACK from @synesthesiam will be nice, so he knows why "his" add-on got updated :)

Btw. we might also consider adding a "known issues" section to the README of the add-on, so people are aware of this limitation.

sairon (approved):

synesthesiam:
This looks good to me :+1:   Thanks!

BertHankes:
101: still Fatal no AVX support on intel cpu

agners:
@BertHankes yes, this does not intend to support CPU's without AVX. It just lets you know what the problem of your system is...

iainfogg:
It just keeps looping on mine with an error every second, rather than erroring then stopping, which would seem a better approach

BertHankes:
after running it is indeed looping. it should stop. It can only be stopped bij delete the addon.

PR #3372, opened 2023-12-23
vlc: Fix race condition at startup

agners:
Fixes: https://github.com/home-assistant/addons/issues/3186#issuecomment-1868098475

### DIFF
--- a/vlc/CHANGELOG.md
+++ b/vlc/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.3.0
+
+- Prevent race condition reading secrets on startup
+- Update to Alpine 3.19
+
 ## 0.2.0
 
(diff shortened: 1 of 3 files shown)


frenck (approved):
Thanks, @agners 👍 

../Frenck

upsuper:
@agners Thanks for fixing it.

Out of curiosity, could you explain a bit how this would fix the race condition of creating the secret? It seems to me the only change in this PR is adding an empty `base` file, which isn't very obvious how it can affect the execution order between secret reading and writing.

frenck:
It ensures the base of the S6-Overlay runs first (which should always be set up first).
The empty file denotes a dependency.

upsuper:
I see. Thanks for explaining!

Issue #3289, opened 2023-11-01
NGINX SSL Proxy add-on fails to work but no logs to debug


bucklevision:
### Describe the issue you are experiencing

Add-on shows as started, then 20 seconds late it stops itself.
At no point does 443: get opened on the machine that i can see via netstat.
Browser reports connection refused.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

NGINX Home Assistant SSL proxy

### What is the version of the add-on?

3.5.0

### Steps to reproduce the issue

DuckDNS installed and working, certs checked and fine on domain.
Install NGINX SSL Proxy add-on and configure with FQDN used by duckdns (https://X.duckdns.org) and operating on port 443
Forward port on router 443 to the HAOS machine 443.
Enable NGINX proxy - no response at X.duckdns.org:443

### System Health information

No repairs, system healthy. No button/pop-up.

### Anything in the Supervisor logs that might be useful for us?

```txt
Nothing visible.
```

### Anything in the add-on logs that might be useful for us?

```txt
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
s6-rc: info: service legacy-cont-init successfully started
s6-rc: info: service nginx: starting
s6-rc: info: service nginx successfully started
s6-rc: info: service crond: starting
s6-rc: info: service crond successfully started
s6-rc: info: service legacy-services: starting
s6-rc: info: service legacy-services successfully started
sed: bad option in substitution expression
[17:20:58] INFO: Service nginx exited with code 1 (by signal 0)
s6-rc: info: service legacy-services: stopping
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service crond: stopping
[17:20:58] INFO: Service crond exited with code 256 (by signal 15)
s6-rc: info: service crond successfully stopped
s6-rc: info: service nginx: stopping
s6-rc: info: service nginx successfully stopped
s6-rc: info: service legacy-cont-init: stopping
s6-rc: info: service legacy-cont-init successfully stopped
s6-rc: info: service fix-attrs: stopping
s6-rc: info: service fix-attrs successfully stopped
s6-rc: info: service s6rc-oneshot-runner: stopping
[...]

PR #3373, opened 2023-12-26
nginx_proxy: Update documentation

tkurki:
- Make the difference between global configuration and add-on configuration clear.
- Include reference to 400 bad request (symptom of missing configuration)
- Add reference to Let's Encrypt.
- Demote the note about no longer relevant configuration options.

See also #3289.

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -10,14 +10,10 @@ Follow these steps to get the add-on installed on your system:
 
 ## How to use
 
-The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario.
+The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) and/or the [Let's Encrypt](https://github.com/home-assistant/addons/tree/master/letsencrypt) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario.
 
-1. The certificate to your registered domain should already be created via the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on or another method. Make sure that the certificate files exist in the `/ssl` directory.
-2. In the `configuration.yaml` file, some options in the `http:` section are no longer necessary for this scenario, and should be commented out or removed:
-   - `ssl_certificate`
-   - `ssl_key`
-   - `server_port`
-3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set).
+1. The certificate to your registered domain should already be created via [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns), [Let's Encrypt](https://github.com/home-assistant/addons/tree/master/letsencrypt) or another method. Make sure that the certificate files exist in the `/ssl` directory.
+2. You must add the following section to your [Home Assistant configuration.yaml](https://www.home-assistant.io/docs/configuration/). If the `http` section is using the `ssl_certificate`, `ssl_key` or `server_port` keys, make sure to remove them.
 
    ```yaml
    http:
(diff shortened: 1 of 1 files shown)


nginx_proxy/DOCS.md
```diff
@@ -10,28 +10,29 @@ Follow these steps to get the add-on installed on your system:
 
 ## How to use
 
-The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario.
+The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario. Another option for certificate management is [Let's Encrypt](https://github.com/home-assistant/addons/tree/master/letsencrypt) add-on.
```
agners, line 13:
```suggestion
The NGINX Proxy add-on is commonly used in conjunction with the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) and/or the [Let's Encrypt](https://github.com/home-assistant/addons/tree/master/letsencrypt) add-on to set up secure remote access to your Home Assistant instance. The following instructions covers this scenario.
```

nginx_proxy/DOCS.md
```diff
@@ -10,28 +10,29 @@ Follow these steps to get the add-on installed on your system:
...
 
-1. The certificate to your registered domain should already be created via the [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns) add-on or another method. Make sure that the certificate files exist in the `/ssl` directory.
-2. In the `configuration.yaml` file, some options in the `http:` section are no longer necessary for this scenario, and should be commented out or removed:
-   - `ssl_certificate`
-   - `ssl_key`
-   - `server_port`
-3. And you need to add the `trusted_proxies` section (requests from reverse proxies will be blocked if these options are not set).
+1. The certificate to your registered domain should already be created via [Duck DNS](https://github.com/home-assistant/addons/tree/master/duckdns), [Let's Encrypt](https://github.com/home-assistant/addons/tree/master/letsencrypt) or another method. Make sure that the certificate files exist in the `/ssl` directory.
+2. You must add the following section to your [Home Assistant global configuration.yaml](https://www.home-assistant.io/docs/configuration/), otherwise the Home Assistant server process will respond with `400 Bad Request` to requests arriving via Nginx add-on reverse proxy.
```
agners, line 16: I think we should still tell users to remove any HA specific SSL options in this step, so:

```suggestion
2. You must add the following section to your [Home Assistant configuration.yaml](https://www.home-assistant.io/docs/configuration/), otherwise the Home Assistant Core process will respond with `400 Bad Request` to requests arriving via this NGINX Porxy add-on. If the `http` section is using the `ssl_certificate`, `ssl_key` or `server_port` keys, make sure to remove them.
```


The question is if is really worthwhile to document what happens when not following this process? It might be more confusing? :thinking:

nginx_proxy/DOCS.md
```diff
@@ -10,28 +10,29 @@ Follow these steps to get the add-on installed on your system:
...
+4. Leave all other options as-is.
+5. Save configuration.
+6. Start the add-on.
+7. Have some patience and wait a couple of minutes.
+8. Check the add-on log output to see the result.
+
+Note: some options in the `http:` section of `configuration.yaml` are no longer necessary for this scenario, and should be commented out or removed:
+   - `ssl_certificate`
+   - `ssl_key`
+   - `server_port`
```
agners, line 34:
```suggestion
```

nginx_proxy/DOCS.md
tkurki: I intentionally included the 400 response and why you get that so that people seeing that would understand what is happening. I think it is a good idea to include the 400 so you can go "oh, that's where I am, here is the fix I am looking for", the note about HA core is not so valuable.

How long do you think the note about the `http` sections needs to be there? I moved it to the bottom so that it might eventually go away, as a fresh installation like mine does not contain that and I I have no idea how prevalent that is.

Your suggestion is 👍 as it stands, just that there's a typo in `Porxy`.

agners: Whoops, fixed the "Porxy" typo.

> I think it is a good idea to include the 400 so you can go "oh, that's where I am, here is the fix I am looking for", the note about HA core is not so valuable.

I guess you could miss any one of these numbers, and you'll run into another issue. IMHO, these type of what happen if you did not follow on every step would hurt the flow of the instructions very much. I'd prefer to have them in a Troubleshooting section further down.

> How long do you think the note about the http sections needs to be there?

Yeah good question not sure. These are still valid options, and people might migrate from a non-Proxy setup to a Proxy setup. This could also be the case for new installations, e.g. if people followed instructions using Core SSL implementation.

tkurki:
- added Troubleshooting section and moved 400 response comment there
- squashed and force pushed

agners (approved):

Issue #3303, opened 2023-11-11
DHCP server quits on launch due to dhcpd.conf file parsing error


FreelancerJay:
### Describe the issue you are experiencing

I have just set up a new Home Assistant OS installation in a VM, and installed the DHCP Server addon.

I changed the IP ranges from the defaults, set the interface to run on and added a domain.
When I started the addon, it ran for a couple of seconds and then quit, reporting an issue on line 3 of dhcpd.conf, expecting a semicolon. This is the line that has the default-lease-time setting.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

DHCP Server

### What is the version of the add-on?

1.3.0

### Steps to reproduce the issue

1. Install Home Assistant OS 11.1
2. Use `ha cli` to set static IP on network interface DHCP server will be running on
3. Install DHCP Server
4. Start addon

### System Health information

Nothing

### Anything in the Supervisor logs that might be useful for us?

```txt
Nothing
```

### Anything in the add-on logs that might be useful for us?

```txt
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] done.
[services.d] starting services
[services.d] done.
[21:28:17] INFO: Creating DHCP configuration...
[21:28:19] INFO: Starting DHCP server...
Internet Systems Consortium DHCP Server 4.4.2-P1
Copyright 2004-2021 Internet Systems Consortium.
All rights reserved.
For info, please visit https://www.isc.org/software/dhcp/
/etc/dhcpd.conf line 3: semicolon expected.
default-lease-time 
 ^
Configuration file errors encountered -- exiting

If you think you have received this message due to a bug rather
than a configuration issue please read the section on submitting
bugs on either our web page at www.isc.org or in the README file
before submitting a bug.  These pages explain the proper
process and the information we find helpful for debugging.

exiting.
[cmd] /run.sh exited 1
[cont-finish.d] executing container finish scripts...
[cont-finish.d] done.
[s6-finish] waiting for services.
[s6-finish] sending all processes the TERM signal.
[s6-finish] sending all processes the KILL signal and exiting.
```

### Additional information

My config:
```
default_lease: 86400
dns: []
ntp: []
[...]

PR #3374, opened 2023-12-27
dhcp_server: Don't fail if no DNS is provided

agners:
Fixes: #3303


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3376, opened 2023-12-28
letsencrypt: Add option to change key type

Fufs:
Adds option to change the Private Key type. This might be necessary as some devices do not support ECDSA keys (like Tasmotas with MQTTS).

As some services require one or the other type, in the future it might be beneficial for generating both types (or perhaps generating multiple different certificates).

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -125,6 +125,15 @@ dreamhost_api_key: ''
 
 </details>
 
+<details>
+  <summary>Changing the key type</summary>
+  
+  Starting with Certbot version 2.0.0 (aaddon version 5.0.0), ECDSA keys are now the default. They utilize a more secure cryptography algorithm, however, they are not supported everywhere yet. For instance, Tasmota does not support MQTTS with an ECDSA key. If your use case does not support ECDSA keys, you can change them with the `keytype` parameter.
+
+  ```yaml
+  keytype: rsa
+  ``` 
+
 
 ## Example Configurations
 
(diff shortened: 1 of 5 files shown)


letsencrypt/config.yaml
```diff
@@ -30,6 +31,7 @@ schema:
   domains:
     - str
   email: email
+  keytype: list(ecdsa|rsa)
```
agners, line 34: Since this is a new option, I think it should be optional. Also, certbot only supports a single parameter as `--key-type`, so this should be a str.

Unless you want to support multiple keytypes as you alluded in the intro, but in this case you'd need to invoke certbot multiple times.

```suggestion
  keytype: str?
```

Fufs: There's only 2 options available in certbot, so I set it to 'list(ecdsa|rsa)` just like the `challenge:` parameter did, as it seems to make it a select form input in the UI where only one option can be picked. I assumed `list(a|b)` means "either a or b" but if it's wrong then str will work. 

I agree about the optional thing but I'm gonna have to edit the run script to allow the keytype parameter to not be present.

So perhaps `keytype: list(ecdsa|rsa)?` or as you suggest `keytype: str?` depending on the first thing

agners: > I assumed `list(a|b)` means "either a or b" but if it's wrong then str will work.

Yes, right, that is a list of valid values. Which is correct. Sorry about that.

Making it just optional is fine. :+1:

letsencrypt/DOCS.md
```diff
@@ -125,6 +125,15 @@ dreamhost_api_key: ''
 
 </details>
 
+<details>
+  <summary>Changing the key type</summary>
+  
+  Starting with Certbot version 2.0.0 (aaddon version 5.0.0), ECDSA keys are now the default. They utilize a more secure cryptography algorithm, however, they are not supported everywhere yet. For instance, Tasmota does not support MQTTS with an ECDSA key. If your use case does not support ECDSA keys, you can change them with the `keytype` parameter.
```
agners, line 131:
```suggestion
  Starting with Certbot version 2.0.0 (add-on version 5.0.0 and newer), ECDSA keys are now the default. These keys utilize a more secure cryptography algorithm, however, they are not supported everywhere yet. For instance, Tasmota does not support MQTTS with an ECDSA key. If your use case does not support ECDSA keys, you can change them with the `keytype` parameter.
```

agners (commented):
Can you also update the changelog and increment the version in the config.

agners (approved):

PR #3384, opened 2024-01-04
openthread_border_router: Update to OpenThread RCP v4.4.0.0

agners:
Update the Home Assistant SkyConnect and Yellow firmwares to the latest version built from Gecko SDK v4.4.0.0. These variants also come with a built-in watchdog in case the RCP firwmare crashes.

Also update to the latest version of the Universal Silicon Labs flasher.


frenck (approved):
Thanks, @agners 👍 

../Frenck

Issue #3386, opened 2024-01-05
silabs_multiprotocol: Z2M broken after upgrading to 2.4.0


n-prat:
### Describe the issue you are experiencing

See https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/issues/555

```
[13:29:31] INFO: Preparing to start...
[13:29:31] INFO: Socat not enabled
[13:29:32] INFO: Starting Zigbee2MQTT...
Zigbee2MQTT:info 2024-01-05 13:29:34: Logging to console and directory: '/config/zigbee2mqtt/log/2024-01-05.13-29-34' filename: log.txt
Zigbee2MQTT:info 2024-01-05 13:29:34: Starting Zigbee2MQTT version 1.35.0 (commit #unknown)
Zigbee2MQTT:info 2024-01-05 13:29:34: Starting zigbee-herdsman (0.30.0)
Assertion failed: Command (setConfigurationValue) returned unexpected state: [object Object]
Assertion failed: Command (setValue) returned unexpected state: 55
Error: {"address":0,"clusterId":32770,"sequence":2} after 10000ms
at Timeout._onTimeout (/app/node_modules/zigbee-herdsman/src/utils/waitress.ts:64:35)
at listOnTimeout (node:internal/timers:569:17)
at processTimers (node:internal/timers:512:7)
```

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Silicon Labs Multiprotocol

### What is the version of the add-on?

2.4.0

### Steps to reproduce the issue

1. Upgrade Core,OS,Z2M
2. Check everything is OK
3. Upgrade SiLabs to 2.4.0
4. ZM2 can't start anymore
...

## System Information

version | core-2024.1.0
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.70-haos
arch | x86_64
timezone | Europe/Paris
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.33.0
Stage | running
Available Repositories | 1372
Downloaded Repositories | 7
HACS Data | ok

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
[...]

PR #3388, opened 2024-01-05
silabs-multiprotocol: Revert to Gecko SDK v4.3.1

agners:
The Gecko SDK v4.4.0 update requires a minimum Core version of 2024.1, this was missed in the metadata.

The idea is to push two Silicon Labs Multiprotocol update: 1. Downgrade with current Core requirements. 2. (Re)upgrade with proper Core requirements.

Note that it seems that Zigbee2MQTT is not compatible with Gecko SDK 4.4.0 currently as well (see #3386). However, we do not have a way to define cross add-on dependencies today. So the second update will break Z2M setups still :cry:


silabs-multiprotocol/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.4.1
+
+- Revert back to Silicon Labs Gecko SDK 4.3.1
```
puddly, line 5: I'd still like to receive feedback regarding stability of this release (with ZGP removed). Maybe we can extend the message?

```suggestion
- Revert back to Silicon Labs Gecko SDK 4.3.1 to maintain compatibility with Home Assistant Core 2023.12.0 and Zigbee2MQTT. If you are running 2024.1.0 with ZHA, you don't have to downgrade.
```

agners: Yeah make sense :+1:

puddly (approved):

PR #3389, opened 2024-01-05
letsencrypt: fix docs broken by missing closing tag

sairon:
Closing tag was missing, causing Example Configurations and everything after that failing to show in rendered view.


frenck (approved):
Thanks, @sairon 👍 

../Frenck

PR #3396, opened 2024-01-06
openthread_border_router: Fix TREL and enable channel manager

agners:
Enable TREL by default on the primary network interface (which we consider as infrastructure interface).

Enable the channel manager, but leave it off by default. This allows to use `ot-ctl channel monitor start` and `ot-ctl channel monitor` to see which IEEE 802.15.4 channel is least busy.

Remove the CoAP API support as they are not required for the Home Assistant use case.

### DIFF
--- a/openthread_border_router/Dockerfile
+++ b/openthread_border_router/Dockerfile
@@ -83,6 +83,8 @@ RUN \
         -DOT_POSIX_NAT64_CIDR="192.168.255.0/24" \
         -DOTBR_DNS_UPSTREAM_QUERY=ON \
         -DOT_CHANNEL_MONITOR=ON \
+        -DOT_COAP=OFF \
+        -DOT_COAPS=OFF \
         && cd build/otbr/ \
         && ninja \
         && ninja install) \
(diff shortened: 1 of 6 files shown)


openthread_border_router/Dockerfile
```diff
@@ -79,6 +82,9 @@ RUN \
         -DOTBR_NAT64=ON \
         -DOT_POSIX_NAT64_CIDR="192.168.255.0/24" \
         -DOTBR_DNS_UPSTREAM_QUERY=ON \
+        -DOT_CHANNEL_MONITOR=ON \
+        -DOT_COAP=OFF \
+        -DOT_COAPS=OFF \
```
marcelveldt, line 87: this wasn't mentioned in the PR description, what does it do ?

marcelveldt (approved):
Looks fine to me!
Had one comment that you should maybe update the PR description.

openthread_border_router/Dockerfile
agners: Added a note! :+1:

PR #3399, opened 2024-01-07
nginx_proxy: Support longer domains name

samuelef95:
Add support for longer domain name, modifying `server_names_hash_bucket_size` from 64 to 128

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.7.0
+
+- Modify `server_names_hash_bucket_size` to add support for longer domain names
+  
 ## 3.6.0
 
 - Add port to Host header to fix origin issues affecting ESPHome and other addons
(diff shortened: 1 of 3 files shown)


agners:
In what case is this problematic? I guess if you use a long domain name, but what is the outcome in that case?

samuelef95:
Yes, exactly when i use a long domain name.

The error I find in the log Is:

INFO: Running nginx...
nginx: [emerg] could not build server_names_hash, you should increase server_names_hash_bucket_size: 64
[19:27:51] INFO: Service nginx exited with code 1 (by signal 0)
s6-rc: info: service legacy-services: stopping
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service crond: stopping
[19:27:51] INFO: Service crond exited with code 256 (by signal 15)

I didn't find a way to increase the value using custom nginx configuration because it's already defined in "http" method and can be defined only only there.

As a temporary solution I'm using *.duckdns.org.

agners (approved):

Issue #3403, opened 2024-01-09
Whisper crash after version 1.0.1.


andregoncalvespires:
### Describe the issue you are experiencing

Use HASSOS on a proxmox virtual machine. After upgrading Whisper to version 1.0.1, the error "FATAL: Your CPU does not support the Advanced Vector Extensions required by Whisper." started to be displayed in the addon log.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Whisper

### What is the version of the add-on?

1.0.1

### Steps to reproduce the issue

1. Use HASSOS in VM proxmox
2. Instal or Update 1.0.1 whisper addon version
3. Start addon

## System Information

version | core-2024.1.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.70-haos
arch | x86_64
timezone | America/Sao_Paulo
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 4481
Installed Version | 1.33.0
Stage | running
Available Repositories | 1381
Downloaded Repositories | 38

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 11.3
-- | --
update_channel | stable
supervisor_version | supervisor-2023.12.0
agent_version | 1.6.0
docker_version | 24.0.7
disk_total | 30.8 GB
disk_used | 17.6 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
installed_addons | Advanced SSH & Web Terminal (17.0.2), ESPHome (2023.12.5), Grafana (9.1.2), Home Assistant Google Drive Backup (0.105.2), InfluxDB (5.0.0), MQTT Explorer (browser-1.0.1), Piper (1.4.0), Samba share (12.2.0), Studio Code Server (5.15.0), Zigbee2MQTT (1.35.1-1), openWakeWord (1.8.2), Matter Server (5.0.2), Mosquitto broker (6.4.0), Whisper (1.0.1)

</details>

[...]

PR #3405, opened 2024-01-09
Fix AVX check for Whisper

emwjacobson:
Commit 314ec9f54966ae14099bb874f1e5547b8cdd46fa prevented any non-AVX compatible CPUs from running Whisper despite it running okay, if just very slowly, before.

Related Issue: #3403 

I believe that rather than completely excluding non-AVX CPUs, that a warning should be output in the logs so that users can still use Whisper.


agners:
@emwjacobson thanks for the PR! Can you also bump the version in config.yaml and add an entry to the `CHANGELOG.md`? Thanks!

emwjacobson:
Done :)

whisper/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.0.2
+
+- Added warning for CPUs not supporting AVX instructions
```
agners, line 5:
```suggestion
- Convert error to warning for CPUs not supporting AVX instructions
```

PR #3409, opened 2024-01-13
letsencrypt: Add support for Hurricane Electric

dotdoom:


letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -169,6 +169,12 @@ elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-dreamhost" ]; th
     bashio::config.require 'dns.dreamhost_api_key'
     PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--dns-dreamhost-credentials" "/data/dnsapikey")
 
+# Hurricane Electric
+elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-he" ]; then
+    bashio::config.require 'dns.he_user'
+    bashio::config.require 'dns.he_pass'
+    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--dns-he-credentials" "/data/dnsapikey")
```
agners, line 176: I'd suggest to use the standard argument set here, which also includes propagation seconds:

```suggestion
    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" "/data/dnsapikey" "--${DNS_PROVIDER}-propagation-seconds" "${PROPAGATION_SECONDS}")
```

dotdoom:
Thank you for the review.  While verifying that these arguments work for this plugin, I've found another one implementing the same, and it seems a bit more active.  Currently trying to decide between [certbot-dns-henet](https://github.com/gentoo-root/certbot-dns-henet) and [certbot-dns-he](https://github.com/tsaaristo/certbot-dns-he), let me get back to this in the next day or two.

dotdoom:
Done.  Stayed with dns-he since that has a pip package, resulting in a uniform install method and dependency management.

PR #3414, opened 2024-01-16
matter_server: Bump Python Matter server to 5.2.1

agners:


marcelveldt (approved):

irichter:
@agners, I see that this commit assumes I will always have my matter devices on the first interface.

This change might break my [currently working] matter setup because my HA is connected to multiple VLANs, but my Matter devices aren't connected to the primary VLAN. Can you confirm whether upgrading my matter server addon will break this configuration?

Are there any plans to support having devices on alternate interfaces, or across several interfaces? Is there a configuration option I can use to choose an interface?

agners:
> This change might break my [currently working] matter setup because my HA is connected to multiple VLANs, but my Matter devices aren't connected to the primary VLAN. Can you confirm whether upgrading my matter server addon will break this configuration?

Not generally. The only place where we use the primary interface in the Python Matter server is when receiving a link-local address from the Android companion app (see https://github.com/home-assistant-libs/python-matter-server/pull/501 and https://github.com/home-assistant/android/pull/4069). We started to pass the IP address because otherwise we run into commissioning issues *sometimes* (see https://github.com/home-assistant-libs/python-matter-server/issues/463).

Now I admit, the whole situation is not ideal. I am currently checking if there is another option to use the Android API which is more reliable then just using pin code.

> Are there any plans to support having devices on alternate interfaces, or across several interfaces? Is there a configuration option I can use to choose an interface?

As you've discovered the Matter SDK already listens for mDNS responses on all interfaces. I intended not to break this.

To avoid unnecessary traffic and complexity, we might limit the interfaces we listen on in the future. But I'd like to support advanced network setup as much as possible. See also this recent, somewhat related discussion on the OTBR add-on https://github.com/home-assistant/addons/pull/3416#pullrequestreview-1833105096.

Issue #3124, opened 2023-06-29
Multiprotocol consistently mismatches channels


Jonathanese:
### Describe the issue you are experiencing

Addon: Silicon Labs Multiprotocol
Version:  2.2.0
Device: Sonoff Zigbee Dongle-E
Firmware: RCP Multi-Protocol

Every time I set up the SiLabs multiprotocol plugin and integrations, I immediately get an error saying the channels are mismatched. It sets the Zigbee channel to 25, and the Thread channel to 15.

I try to change the Thread channel by running Form on the OTBR page, setting it to 25. This tends to get rid of the error, but I can't pair anything via Matter. It finds the device, connects, generates the credentials, sends the credentials, then checks the connection. And it always fails at that last step, which I assume is a Thread issue because of a new network which perhaps the rest of the system hasn't detected.

I try to change the Zigbee channel, but rather than letting me change it, it says "This is configured by the Multiprotocol addon and must be changed in the hardware page". But there's no hardware page to set this. So it basically configures the channels wrong and prevents me from fixing it.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Silicon Labs Multiprotocol

### What is the version of the add-on?

2.2.0

### Steps to reproduce the issue

1. Flash Dongle-E using the online flashing tool. Use the RCP Multiprotocol firmware.
2. Install the Multiprotocol addon. 460kbaud. Software flow control. OTBR enabled. Ports enabled.
3. Point ZHA to the multiprotocol virtual device
4. Thread and OTBR are auto-detected
5. Reboot Home Assistant OS to re-run detection
6. <Mismatched Channel" error usually occurs here as a settings notification>
7. In ZHA configuration, try to change channel. Prompt tells you you can't do so. References a page in "Hardware" that doesn't exist.

**Try matter**
8. Install Matter addon. Autodetected/Auto-Added to integrations.
9. On phone, try to add Nanoleaf A12 bulb.
10. Sequence: Finding Device -> Connecting to device -> Generating Matter Credentials -> Testing connection -> Connection failed

**Try Change Channel**
11. Open OTBR page
12. Go to "Form", and configure network for channel 25 with the name home-assistant. Defaults elsewhere
13. Click "Form"
[...]

PR #3416, opened 2024-01-17
silabs-multiprotocol: Make the OTBR infrastructure network interface configurable

PaulSD:
The OpenThread Border Router agent currently only supports routing between a single "infrastructure" network interface and a single Thread network interface per agent process.  (Multiple "infrastructure" interfaces can be specified, but this is for fail-over purposes and only one will be used at a time.)

Thread device commissioning in HA currently requires an Android device which can communicate both with the Thread device via Bluetooth and with the OpenThread Border Router "infrastructure" interface via IPv6 with Router Advertisements and mDNS.

Prior to this commit, the first HA host interface with an associated default route was always used as the OTBR "infrastructure" interface.  This caused problems if the HA host has multiple interfaces and the Android device needed for commissioning is on a different network than the automatically-selected interface.

To address that, this commit permits the user to configure the "infrastructure" interface.

(Side note: Historically, the "infrastructure" interface was sometimes conflated with the "backbone" interface, but those have now been unified under the term "infrastructure": https://github.com/openthread/openthread/pull/9638 )

### DIFF
--- a/silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
+++ b/silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
@@ -5,40 +5,40 @@
 
 . /etc/s6-overlay/scripts/otbr-agent-common
 
-declare otbr_infra_if
 declare device
 declare baudrate
 declare flow_control
 declare otbr_log_level
 declare otbr_log_level_int
+declare otbr_infra_if
 declare otbr_rest_listen
 declare otbr_rest_listen_port
 
 otbr_log_level=$(bashio::string.lower "$(bashio::config otbr_log_level)")
 case "${otbr_log_level}" in
     debug)
-        otbr_log_level_int="7"
+        otbr_log_level_int='7'
         ;;
     info)
-        otbr_log_level_int="6"
+        otbr_log_level_int='6'
         ;;
     notice)
-        otbr_log_level_int="5"
+        otbr_log_level_int='5'
         ;;
     warning)
-        otbr_log_level_int="4"
+        otbr_log_level_int='4'
         ;;
     error)
-        otbr_log_level_int="3"
+        otbr_log_level_int='3'
         ;;
     critical)
-        otbr_log_level_int="2"
+        otbr_log_level_int='2'
         ;;
     alert)
-        otbr_log_level_int="1"
+        otbr_log_level_int='1'
         ;;
     emergency)
-        otbr_log_level_int="0"
+        otbr_log_level_int='0'
         ;;
     *)
         bashio::exit.nok "Unknown otbr_log_level: ${otbr_log_level}"
@@ -58,26 +58,26 @@ else
     fi
 fi
 
-mkdir -p /data/thread && ln -sft /var/lib /data/thread || bashio::exit.nok "Could not create directory /var/lib/thread to store Thread data."
+mkdir -p /data/thread && ln -sft /var/lib /data/thread || bashio::exit.nok 'Could not create directory /var/lib/thread to store Thread data.'
 
 if bashio::config.true 'otbr_firewall'; then
-    bashio::log.info "Setup OTBR firewall..."
+    bashio::log.info 'Setup OTBR firewall...'
     ipset create -exist otbr-ingress-deny-src hash:net family inet6
     ipset create -exist otbr-ingress-deny-src-swap hash:net family inet6
     ipset create -exist otbr-ingress-allow-dst hash:net family inet6
     ipset create -exist otbr-ingress-allow-dst-swap hash:net family inet6
 
     ip6tables -N $otbr_forward_ingress_chain
-    ip6tables -I FORWARD 1 -o $thread_if -j $otbr_forward_ingress_chain
+    ip6tables -I FORWARD 1 -o "${thread_if}" -j $otbr_forward_ingress_chain
 
-    ip6tables -A $otbr_forward_ingress_chain -m pkttype --pkt-type unicast -i ${thread_if} -j DROP
+    ip6tables -A $otbr_forward_ingress_chain -m pkttype --pkt-type unicast -i "${thread_if}" -j DROP
     ip6tables -A $otbr_forward_ingress_chain -m set --match-set otbr-ingress-deny-src src -j DROP
     ip6tables -A $otbr_forward_ingress_chain -m set --match-set otbr-ingress-allow-dst dst -j ACCEPT
     ip6tables -A $otbr_forward_ingress_chain -m pkttype --pkt-type unicast -j DROP
     ip6tables -A $otbr_forward_ingress_chain -j ACCEPT
 
     ip6tables -N $otbr_forward_egress_chain
-    ip6tables -I FORWARD 2 -i $thread_if -j $otbr_forward_egress_chain
+    ip6tables -I FORWARD 2 -i "${thread_if}" -j $otbr_forward_egress_chain
     ip6tables -A $otbr_forward_egress_chain -j ACCEPT
 else
     # Make sure ip6tables (as used by Docker) allow IP forwarding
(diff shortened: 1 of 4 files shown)


silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -5,73 +5,79 @@
...
 declare otbr_rest_listen
 declare otbr_rest_listen_port
 
-backbone_if="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
-
 otbr_log_level=$(bashio::string.lower "$(bashio::config otbr_log_level)")
 case "${otbr_log_level}" in
     debug)
-        otbr_log_level_int="7"
+        otbr_log_level_int='7'
```
agners, line 20: Is there a particular reason to use this style?

silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -5,73 +5,79 @@
...
+if bashio::config.has_value 'otbr_infra_if'; then
+    otbr_infra_if="$(bashio::config 'otbr_infra_if')"
+    bashio::log.info "Using configured network interface ${otbr_infra_if} for otbr_infra_if."
+else
+    otbr_infra_if="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+    if [ -n "${otbr_infra_if}" ]; then
+        bashio::log.info "Using primary network interface ${otbr_infra_if} for otbr_infra_if."
+    else
+        bashio::log.warning 'No primary network interface found!  Using static eth0 for otbr_infra_if.'
+        otbr_infra_if='eth0'
```
agners, line 57: This is actually a bit legacy. `eth0` is almost never the right interface. It would be better if we exit here with an error, e.g.
```suggestion
        bashio::exit.nok  'No primary network interface found!'
```

silabs-multiprotocol/translations/en.yaml
```diff
@@ -25,15 +25,22 @@ configuration:
...
   otbr_log_level:
     name: OpenThread Border Router agent log level
     description: >-
       Set logging level of the OpenThread Border Router agent (otbr-agent).
+  otbr_infra_if:
+    name: OpenThread Border Router infrastructure interface
+    description: >-
+      HA host network interface name to bind the "infrastructure" side (vs the
+      "Thread" side) of the OpenThread Border Router to.  By default, the
+      first interface with an associated default route is used.
```
agners, line 38: We use what is considered primary by NetworkManager. Is the first interface with an associated default what NetworkManager considers primary? :thinking:

agners (commented):
In general, I am a bit hesitant about this PR. I see that this is useful for multi-homed Home Assistant installation, but if we go down this road, we'll have to add a similar option to more places (e.g. the OpenThread Border Router add-on, the Matter Server add-on...).

A feature I have in the back of my mind to add is macvlan for add-ons in general. Currently we are on the host network, which is not very nice anyways. A macvlan would have separate, but L2 access to a particular network interface.

Assuming we can configure a mavclan somewhere in HA, and then assign this to one or multiple add-ons would solve this much more elegant.

silabs-multiprotocol/translations/en.yaml
PaulSD: I said "first" here because we used this code: `bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface
'`

However, I think (but am not entirely sure) that NetworkManager will only ever have one primary interface, and I think that is the most recently activated interface that has an associated default route (either IPv4 or IPv6 default route).

silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
PaulSD: Agreed.  I thought this seemed like an odd choice for a default, but I didn't want to break anyone if this is actually used somewhere.  I will change it.

silabs-multiprotocol/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
PaulSD: Single quote does not interpolate; Double quote interpolates.
So, the idea is to avoid interpolation when it isn't needed.

Obviously this is a minor nit, so if you don't like this change then I can drop it.

PaulSD:
> A feature I have in the back of my mind to add macvlan support for add-ons in general. Currently we are on the host network, which is not very nice anyways. A macvlan would have separate, but L2 access to a particular network interface.
> Assuming we can configure a mavclan somewhere in HA, and then assign this to one or multiple add-ons would solve this much more elegant.

I agree that would be a cleaner solution, but it would require adding settings and UI to HA Core to facilitate creating macvlan networks and attaching them to host interfaces and add-ons ... And being new to HA and unfamiliar with all the use cases and internals, that is not something I'm comfortable tackling myself (at least not without some significant guidance/direction).

Creating macvlan networks could go on the System -> Network page ... But where would the associations to add-ons be configured?  (Maybe in each add-on's Info or Configuration tab?) ... And how would we determine/configure which add-ons should use this vs which add-ons should use the host network (I imagine some add-ons will need to be able to access multiple physical networks, so macvlan wouldn't work for those) - Should the add-on specify whether host network or macvlan should be used, or should that be configurable by the user for all add-ons, or ...?

> In general, I am a bit hesitant about this PR. I see that this is useful for multi-homed Home Assistant installation, but if we go down this road, we'll have to add a similar option to more places (e.g. the OpenThread Border Router add-on, the Matter Server add-on...).

Yeah, I see you just added similar code to the Matter Server add-on here, which should probably have similar configuration: https://github.com/home-assistant/addons/pull/3414/files
Is this something we could add temporarily until a more generalized solution (ala macvlan) can be implemented?

irichter:
Replying to @agners
> A feature I have in the back of my mind to add macvlan support for add-ons in general. Currently we are on the host network, which is not very nice anyways. A macvlan would have separate, but L2 access to a particular network interface.
> 
> Assuming we can configure a mavclan somewhere in HA, and then assign this to one or multiple add-ons would solve this much more elegant.

I'm not sure that using macvlan is the correct solution here, since it's not that a multi-homed HA needs multiple MACs on each interface, but rather that add-ons should not make assumptions about what interfaces to use where it's necessary or desirable to pick one. Definitely some add-ons might benefit from their own L2 MAC and associated IP, but I don't think that's the main use case, and it potentially requires assigning many more external IPs to HA than necessary. I can see the benefits to assigning add-ons distinct internal IPs to permit inter-add-on communication, but not so much for external use.

I think something inspired by [swconfig](https://openwrt.org/docs/techref/swconfig) or [DSA](https://www.kernel.org/doc/html/latest/networking/dsa/dsa.html) to assign interfaces to add-ons makes more sense than just being able to give add-ons their own macvlan(s). (It would still be possible to create an external macvlan interface and assign it to a single add-on.) As an example use case, consider a dnsmasq add-on that serves DHCP clients on some interfaces, and uses a separate upstream interface for DNS. This could also solve the issue of Matter knowing which interface to used when passed a link-local IPv6 address; it could also be leveraged to reduce broadcast traffic to interfaces that do not need it (e.g.: mDNS requests for IoT devices being sent on a non-IoT VLAN).

agners:
Sorry for the delay. I wanted to expand on this a bit more, so I made an architecture discussion at https://github.com/home-assistant/architecture/discussions/1034. Let's continue the more general discussion over there.

As for the add-on specific/this PR at hand:

> Yeah, I see you just added similar code to the Matter Server add-on here, which should probably have similar configuration: https://github.com/home-assistant/addons/pull/3414/files

TBH, I am not quite happy with the interface selection we had to add to the Matter Python server. It makes assumption on which network the commissioning phone and the Matter server runs on. Ideally the Matter add-on should just listen on any interface where the user runs Matter devices on. But anyhow, that is a different story.

> Is this something we could add temporarily until a more generalized solution (ala macvlan) can be implemented?

Since the OTBR indeed only supports a single infrastructure link we would even need such a selection when we support multiple `macvlan`'s :sweat_smile: 

Since this will be an optional option, I am not against adding it.

On a bit a different note: For many people trying to actively use the Thread side of the Silicon Labs Multi-protocol add-on lead to issues quite quickly. Did you observe this too? Maybe you are aware, but we recently shifted our focus more on the pure Thread variant (see also the [SkyConnect section of the Matter blog post](https://www.home-assistant.io/blog/2024/01/25/matter-livestream-blog/#using-home-assistant-yellow-or-home-assistant-skyconnect)).

Or in different words, I rather prefer to see this added to the [OpenThread Border Router add-on](https://github.com/home-assistant/addons/tree/master/openthread_border_router) :innocent:

PaulSD:
> Or in different words, I rather prefer to see this added to the [OpenThread Border Router add-on](https://github.com/home-assistant/addons/tree/master/openthread_border_router) 😇

Ok, I can add this in both the Multiprotocol and OTBR add-ons.  I'll work on that.

> On a bit a different note: For many people trying to actively use the Thread side of the Silicon Labs Multi-protocol add-on lead to issues quite quickly. Did you observe this too? Maybe you are aware, but we recently shifted our focus more on the pure Thread variant (see also the [SkyConnect section of the Matter blog post](https://www.home-assistant.io/blog/2024/01/25/matter-livestream-blog/#using-home-assistant-yellow-or-home-assistant-skyconnect)).

Depending on what you mean by "quite quickly" ... Yes, I have been observing some issues every few days:
* About once every few days, I start getting errors like `otbr-agent[...]: ... [W] Platform------: [netif] Failed to transmit, error:Parse` and all or most of my Thread devices drop offline.  Restarting the Multiprotocol add-on usually fixes this.
* About once a week, the radio hardware seems to stop working and I get errors like `Info : Failed to connect, secondary seems unresponsive`.  Unplugging and replugging the USB port for the radio then restarting the Multiprotocol add-on usually fixes this.

I will try switching to the pure Thread variant and see if things improve for me.

BTW, I've been keeping [this comment](https://github.com/home-assistant/addons/issues/3124#issuecomment-1872996807) updated with a list of issues I encountered (including the above) and how I dealt with them.  Is there a better place that I should be posting this information?

agners:
> Depending on what you mean by "quite quickly" ... Yes, I have been observing some issues every few days:

Probably depends on the amount of Thread traffic, but from what I've seen within a day.

> I will try switching to the pure Thread variant and see if things improve for me.
> 
> BTW, I've been keeping [this comment](https://github.com/home-assistant/addons/issues/3124#issuecomment-1872996807) updated with a list of issues I encountered (including the above) and how I dealt with them. Is there a better place that I should be posting this information?

Thanks for putting together the list! Technically, I guess each problem should have it's own issue. But before opening all separately, check if issues for the same problem already exist. I am quite sure that some of the Multiprotocol issues are already tracked in other issues. Some issues I think we have fixed/worked around (e.g. for SkyConnect we have a firmware which comes with a watchdog to workaround the firmware crash).

The problem really is, there are a bit too many unreliable layers here, which makes it hard to reliably know what is happening and what's at fault, especially on the higher levels. Also we really shifted our focus on the pure Thread variant. So ideally retest your setup with the pure Thread variant, and then create separate issues for what is left. Issues which are related to the OTBR belong in this issue tracker, Matter issues we collect in the Core issue tracker.

PR #3421, opened 2024-01-18
letsencrypt ClouDNS bug fix and build fix

Kaurin:
In case a user wants to use the ClouDNS integration 
certbot errors out because there are more than 1 type of ClouDNS
auth entries. All three possible auth methods are currently always
populated, making the ClouDNS integration unusable.

This fix essentially removes the auth options user didn't provide.
This fix still allows the user to mistakenly use more than one
auth type. In this case certbot will error out - but this is desired.

Documented examples show to the user the three possible ways
they can use the letsencrypt addon with ClouDNS.

### DIFF
--- a/letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
+++ b/letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
@@ -53,9 +53,6 @@ echo -e "dns_desec_token = $(bashio::config 'dns.desec_token')\n" \
       "dns_inwx_password = $(bashio::config 'dns.inwx_password')\n" \
       "dns_inwx_shared_secret = $(bashio::config 'dns.inwx_shared_secret')\n" \
       "dns_google_domains_access_token = $(bashio::config 'dns.google_domains_access_token')\n" \
-      "dns_cloudns_auth_id = $(bashio::config 'dns.cloudns_auth_id')\n" \
-      "dns_cloudns_sub_auth_id = $(bashio::config 'dns.cloudns_sub_auth_id')\n" \
-      "dns_cloudns_sub_auth_user = $(bashio::config 'dns.cloudns_sub_auth_user')\n" \
       "dns_cloudns_auth_password = $(bashio::config 'dns.cloudns_auth_password')\n" \
       "dns_dreamhost_baseurl = $(bashio::config 'dns.dreamhost_baseurl')\n" \
       "dns_dreamhost_api_key = $(bashio::config 'dns.dreamhost_api_key')\n" \
@@ -64,6 +61,20 @@ echo -e "dns_desec_token = $(bashio::config 'dns.desec_token')\n" \
 
 if bashio::config.exists 'dns.google_domains_zone'; then
       echo -e "dns_google_domains_zone = $(bashio::config 'dns.google_domains_zone')\n" >> /data/dnsapikey
+fi
+
+# ClouDNS
+# Only a single non-empty auth option must be in /data/dnsapikey when using ClouDNS to avoid a certbot error
+if bashio::config.exists 'dns.cloudns_auth_id'; then
+      echo -e "dns_cloudns_auth_id = $(bashio::config 'dns.cloudns_auth_id')\n" >> /data/dnsapikey
+fi
+
+if bashio::config.exists 'dns.cloudns_sub_auth_id'; then
+      echo -e "dns_cloudns_sub_auth_id = $(bashio::config 'dns.cloudns_sub_auth_id')\n" >> /data/dnsapikey
+fi
+
+if bashio::config.exists 'dns.cloudns_sub_auth_user'; then
+      echo -e "dns_cloudns_sub_auth_user = $(bashio::config 'dns.cloudns_sub_auth_user')\n" >> /data/dnsapikey
 fi
 
 chmod 600 /data/dnsapikey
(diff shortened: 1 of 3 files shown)


Kaurin:
If you prefer to have two separate commits - one for the letsencrypt build fix, and one for the letsencrypt functional fix, let me know. I can flex my git-rebase-foo

letsencrypt/build.yaml
```diff
@@ -8,23 +8,3 @@ build_from:
 codenotary:
   signer: notary@home-assistant.io
   base_image: notary@home-assistant.io
-args:
```
agners, line 11: Passing the version numbers through this metadata is a feature really. If you build using `docker build` manually just use `--build-arg` for each one of these.

Since we use them in all add-ons we should keep them here too.

letsencrypt/Dockerfile
```diff
@@ -36,8 +35,9 @@ RUN \
         musl-dev \
         openssl-dev \
         cargo \
+        py3-pip \
```
agners, line 38: Why is this suddenly necessary? Since we already have add-on deployed, pip definitely worked there.

letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
```diff
@@ -80,3 +80,17 @@ fi
 if bashio::config.exists 'dns.cloudxns_secret_key'; then
       bashio::addon.option 'dns.cloudxns_secret_key'
 fi
+
+# ClouDNS
+# Only a single non-empty auth option must be in /data/dnsapikey.
+# User should pick just one (as per examples). If they pick more than one by accident, they will get an error
```
agners, line 86: I think this is already in the documentation. If you want to do something more elaborate, it would be better to proactively in https://github.com/home-assistant/addons/blob/668e6fad5d1eae56d009c733bfc0d7281a8446d1/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L167-L168. E.g. we can use `bashio::config.has_value`.

agners (changes requested):

letsencrypt/build.yaml
Kaurin: My iteration method was as described in the [Local aad-on testing document](https://developers.home-assistant.io/docs/add-ons/testing). 
* I checked out the addons repo
* VSCode + devcontainer
* Tasks: Start Homeassistant
* I would manually make changes to the letsencrypt addon
* Comment out the image from `config.yaml` (during iteration)
* Reinstall addon in the bespoke homeassistant

During this iteration process, the args would not be injested, nor would pip be available to the built-in build process (which I haven't gotten into). 

Please let me know if there is anything else I need to be aware of during the iteration process

letsencrypt/Dockerfile
Kaurin: See above comment as it is related

letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
Kaurin: ClouDNS is properly documented but currently all three "auth type" keys are left in /data/dnsapikey (values empty by default). Certbot halts if more than one key for ClouDNS exist /data/dnsapikey regardless if only has has a value. Note that this only happens if you use any of the ClouDNS auth types. 

I'm unsure what you meant by "more elaborate" as this is a bug fix for the current implementation (duct-tape as it may be). I suppose I could move the logic to the `run` file. It's my first time contributing so I don't yet have a good feel of the project.

Kaurin:
Undid build changes unrelated to the bugfix

letsencrypt/Dockerfile
Kaurin: I will remove it

letsencrypt/build.yaml
Kaurin: I will trustfall to @agners and undo pipeline changes

letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
Kaurin: If you look a few lines above my added code, there is already some post-processing to the `/data/dnsapikey` being done here, which is why I felt this is the appropriate place to do so. I would appreciate some guidance to as why it would be more suitable for this logic to be moved to the `run` file

letsencrypt/rootfs/etc/cont-init.d/file-structure.sh
```diff
@@ -82,3 +82,16 @@ fi
 if bashio::config.exists 'dns.cloudxns_secret_key'; then
       bashio::addon.option 'dns.cloudxns_secret_key'
 fi
+
+# ClouDNS
+# Only a single non-empty auth option must be in /data/dnsapikey when using ClouDNS to avoid a certbot error
+# Code below removes unused auth types from /data/dnsapikey
+if ! bashio::config.has_value 'dns.cloudns_auth_id'; then
+      sed -i '/.*dns_cloudns_auth_id.*/d' /data/dnsapikey
```
agners, line 90: Instead of deleting, how about removing it from the echo above and only set it if the config exists/has_value? Essentially similar to `dns.google_domains_zone` above.

agners (approved):

Kaurin:
Hold for final tests please, thank you. Testing now

Kaurin:
@agners tests pass.

I also tested **not** following the example and putting in two auth options which gives me the expected error. 
The error is very descriptive and explains to the user they must choose only one of the auth options, so I'd say this is good to go.

agners:
Thanks for your fix and testing!  🤩

MarcMuc:
Thank you all for your effort!

PR #3422, opened 2024-01-19
silabs-multiprotocol: Enable SWD debugging for Yellow and automatically generate coredumps

puddly:
This build augments the multiprotocol addon for the Yellow with debug firmware (built with `-O0`). It will be flashed on top of the existing Gecko SDK 4.4.0 firmware on startup.

OpenOCD with EFR32 patches connects over SWD to the Zigbee module and GDB is set up to dump RAM and registers upon any any assertion being hit. These text core dumps are written to `/share/silabs_multiprotocol_crash_dumps/`.

This probably will be broken out into a separate addon, with the regular multiprotocol addon downgraded to the more stable previous Gecko SDK release.


puddly:
@zalatnaicsongor If you'd like to continue hacking with multi-PAN and have a Yellow handy, try out this branch of the addon. I used it to collect crash information from the firmware itself in a format that can be remotely "debugged".

Issue #3418, opened 2024-01-17
LetsEncrypt addon: Add option to request secp384r1 certificate


DanAE111:
### Describe the issue you are experiencing

I run a private CA with an internal ACME server when the addon requests an ecdsa certificate from my CA it is defaulting to a secp256r1 certificate. Would it be possible to add additional parameters for the elliptic curve type and key size so that they can be specified as part of advanced configuration?

My internal CA kicks back the cert request because the key size is below 384.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.0.9

### Steps to reproduce the issue

1. Have an ACME server that only accepts secp384r1 certificate requests.
2. Configure the addon to use said ACME server.
3. Request a certificate.
4. Certificate request is rejected because the key size requested is too small (secp256r1 instead of secp384r1).

## System Information

version | core-2023.12.3
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.63-haos
arch | x86_64
timezone | Australia/Sydney
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.33.0
Stage | running
Available Repositories | 1383
Downloaded Repositories | 22

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 11.2
-- | --
update_channel | stable
supervisor_version | supervisor-2023.12.1
agent_version | 1.6.0
docker_version | 24.0.7
disk_total | 30.8 GB
disk_used | 10.5 GB
healthy | true
supported | true
board | ova
supervisor_api | ok
version_api | ok
installed_addons | Terminal & SSH (9.8.1), File editor (5.7.0), ESPHome (2023.12.7), NGINX Home Assistant SSL proxy (3.5.0), Samba share (12.2.0), TubesZB TI CC2652 FW Flasher (0.2.2), SQLite Web (4.0.0), Let's Encrypt (5.0.9)

[...]

Issue #3380, opened 2024-01-02
Let's encrypt from RSA to ECDSA message in logs - certificate not generated


jmcollin78:
### Describe the issue you are experiencing

While trying to renew the certificate, I've got this message in logs: 
```
[10:45:47] INFO: Selected DNS Provider: dns-ovh
[10:45:48] INFO: Use propagation seconds: 60
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Are you trying to change the key type of the certificate named home.clouderial.fr from RSA to ECDSA? Please provide both --cert-name and --key-type on the command line to confirm the change you are trying to make.
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
s6-rc: info: service legacy-services: stopping
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service legacy-cont-init: stopping
s6-rc: info: service legacy-cont-init successfully stopped
s6-rc: info: service fix-attrs: stopping
s6-rc: info: service fix-attrs successfully stopped
s6-rc: info: service s6rc-oneshot-runner: stopping
s6-rc: info: service s6rc-oneshot-runner successfully stopped
```

EDIT: the new certificates are not generated. This will cause an major issue because the site will not be visible (in one week for my case).

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Almond

### What is the version of the add-on?

5.0.9

### Steps to reproduce the issue

1. just start the addon to generate a new certificate
2. look at the journal logs

## System Information

version | core-2023.12.4
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.58-haos-raspi
arch | aarch64
timezone | Europe/Paris
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.33.0
Stage | running
Available Repositories | 1373
Downloaded Repositories | 43

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

[...]

PR #3424, opened 2024-01-20
letsencrypt: Default to RSA and allow ECDSA curve selection

WillCodeForCats:
Rework of my previous PR #3130 to address problems raised in issue #3380 and issue #3418.

Summary:

* The default key type will continue to be ECDSA (new certificates, renewals, or new by changing the first domain your domain list).
* The add-on will look for existing certificates. If you have an RSA certificate it will continue to be renewed as RSA. Your certificate will not be migrated to ECDSA in the background if you've been expecting RSA.
* You can set the `key_type` value to upgrade an RSA certificate to ECDSA (or go from ECDSA to RSA). Setting `key_type` disables the auto-detect function.
* For any case, once you set the `key_type` all new certificates and renewals will use the set value of `rsa` or `ecdsa`. To re-enable the auto-detect function you would delete `key_type` from the YAML config.
* New `elliptic_curve` setting to choose from `secp256r1` or `secp384r1` for ECDSA keys (default p384).
* Removes previous `keytype` config setting to use the auto-detect behavior by default.


agners:
Thanks for this PR and raising awareness of this issue.

Ideally, we should have done this with the update to Certbot 2.0.0 (leave RSA as default and make ECDSA as option). However, now that we have started with ECDSA as default, I wonder if with this change, we run into the same problem for folks which started off with ECDSA now? :thinking:  I guess this is far fewer people but still...

Ideally, I think we should detect what type of cert it is, and continue using RSA in that case. At the same time allow explicit configuration for folks who want to migrate.

agners:
> we run into the same problem for folks which started off with ECDSA now? 🤔 

Ah I guess with the explicit `--cert-name` and `--key-type` argument those would get migrated back to RSA.


How about the opposite, force people to migrate to ECDSA? With the major version bump of the add-on/Certbot 2.0.0 we kinda signal a major change... Not sure which situations these days would prefer a RSA over ECDSA, but if folks have troubles with the ECDSA based certificates they could still migrate back...

WillCodeForCats:
Personally I moved to ECDSA certs 3+ months ago without any issues. I'm not currently using this add-on since I wanted to start trying ECDSA certs in the HAOS environment (which is what I tried to do with my previous PR which I rolled into a custom add-on instead). I planned to shift back to the official add-on once it caught up.

There's no way to detect if someone is using something that doesn't work with ECDSA (like already reported for Tasmota) other than tell people to go back to RSA if they have suddenly have problems. I would think that eventually everything that's actively being maintained should support ECDSA eventually. But there will guaranteed breaking for someone somewhere if the key algo changes without their knowledge.

We could do a shell call to try and detect that as part of the setup: make the key_type setting optional where if unset we use auto-detect otherwise if it is set then use that algo and don't run auto-detect.

New installs defaulting to ECDSA I feel is ok because if someone is setting up for the first time they're actively engaged in checking that things are working properly. Existing installs should probably stay with existing key type to avoid breaking stuff that runs automated in the background.

letsencrypt/config.yaml
```diff
@@ -20,7 +20,7 @@ options:
   domains:
     - null
   email: null
-  keytype: ecdsa
+  key_type: ecdsa
```
agners, line 23: This is an unnecessary breaking change. If people already use that option, we'd need to migrate it. But I feel it is not worth it, I'd suggest to just go with `keytype`.

WillCodeForCats: I was using my code from PR #3130 which predated the other naming.

agners:
> We could do a shell call to try and detect that as part of the setup: make the key_type setting optional where if unset we use auto-detect otherwise if it is set then use that algo and don't run auto-detect.

Yeah I think that would be the best way to go here.

A poor man solution would be to just see if the/a certification file already exist. But better if we can check what algorithm it is using.

letsencrypt/translations/en.yaml
```diff
@@ -37,5 +37,11 @@ configuration:
   dns:
     name: DNS
     description: DNS Provider configuration
+  key_type:
+    name: Key Type
+    description: Select the certificate key type.
```
agners, line 42: This is obsolete now.

```suggestion
```

letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 5.0.12
+
+- Allow ECDSA curve selection.
+- Check for existing certificates at startup.
```
agners, line 6:
```suggestion
- By default, choose key type based on existing certificates at startup.
```

letsencrypt/DOCS.md
```diff
@@ -206,13 +206,16 @@ If your custom ACME server uses a certificate signed by an untrusted certificate
 </details>
 
 <details>
-  <summary>Changing the key type</summary>
 
-Starting with Certbot version 2.0.0 (add-on version 5.0.0 and newer), ECDSA keys are now the default. These keys utilize a more secure cryptography algorithm, however, they are not supported everywhere yet. For instance, Tasmota does not support MQTTS with an ECDSA key. If your use case does not support ECDSA keys, you can change them with the `keytype` parameter.
+  <summary>Selecting the Key Type</summary>
+
+  By default the RSA key type is used. You can choose to use an ECDSA key instead.
```
agners, line 212: This needs some updating.

letsencrypt/DOCS.md
```diff
@@ -206,13 +206,16 @@ If your custom ACME server uses a certificate signed by an untrusted certificate
...
 
-Starting with Certbot version 2.0.0 (add-on version 5.0.0 and newer), ECDSA keys are now the default. These keys utilize a more secure cryptography algorithm, however, they are not supported everywhere yet. For instance, Tasmota does not support MQTTS with an ECDSA key. If your use case does not support ECDSA keys, you can change them with the `keytype` parameter.
+  <summary>Selecting the Key Type</summary>
+
+  By default the RSA key type is used. You can choose to use an ECDSA key instead. 
 
   ```yaml
-  keytype: rsa
-  ``` 
+  key_type: 'ecdsa'
```
agners, line 215:
```suggestion
  keytype: 'ecdsa'
```

agners (changes requested):

letsencrypt/DOCS.md
WillCodeForCats: I'll get back to the docs after I test it some more.

letsencrypt/translations/en.yaml
WillCodeForCats: This should still be valid (other than fixing name).

WillCodeForCats:
Any objecting to adding `secp521r1` to the curve list for completeness?

WillCodeForCats:
Whoops sorry about the review request, didn't mean to click that.

letsencrypt/DOCS.md
WillCodeForCats: Updated docs.

letsencrypt/DOCS.md
WillCodeForCats: Updated docs.

letsencrypt/translations/en.yaml
WillCodeForCats: Keeping this for the help text in the config page.

WillCodeForCats:
> Any objecting to adding `secp521r1` to the curve list for completeness?

I'll add it since I think we should have it to be complete if the other two are options, but this can be easily reverted if necessary.

WillCodeForCats:
Reverting the p512 curve since Let's Encrypt doesn't support it in testing:

`Error finalizing order :: invalid public key in CSR: ECDSA curve P-521 not allowed`

WillCodeForCats:
I've created certs as a new install, renewed the certs back and forth between RSA and ECDSA with each option combination, and tested auto detect with both. All passed what I could think of to test.

WillCodeForCats:
I suppose an option for p512 could still be useful to someone running an internal ACME server, but the majority of users for this add-on will probably be using this with Let's Encrypt.

letsencrypt/translations/en.yaml
```diff
@@ -37,5 +37,11 @@ configuration:
   dns:
     name: DNS
     description: DNS Provider configuration
+  keytype:
```
agners, line 40: `keytype` already exist a bit further up. You can move it down to keep it nearby `elliptic_curve`, but we shouldn't have it twice.

WillCodeForCats: Missed, that, fixed now.

letsencrypt/translations/en.yaml
```diff
@@ -37,5 +32,14 @@ configuration:
...
+    name: Key Type
+    description: >-
+      Select the certificate key type. Leave this unset for auto-detect or
+      choose a key type to override the default behavior.
```
agners, line 39: It is kinda obvious that setting it will force it, but the default might be worth noting here:

```suggestion
      Select the certificate key type. If unset, will auto-detect based on
      the key type of the existing certificate or default to ecdsa.
```

letsencrypt/config.yaml
```diff
@@ -20,7 +20,6 @@ options:
   domains:
     - null
   email: null
-  keytype: ecdsa
```
agners, line 23: I think for all those add-ons which got started at least once, the default is already in the option store of the add-on.

But not much we can do about it. At least the ones which get started later will start with an empty default, which should lead to the auto-detection to work. For everyone else, they got auto-migrated to `ecdsa`.

agners (approved):
LGTM! Thanks for your work on this :pray:

letsencrypt/config.yaml
WillCodeForCats: I'll try testing that in a VM, but yeah we can't go back in time to change that.

WillCodeForCats: Yeah it's just stuck in there now if the config was saved while it was part of the default config.

We could go back to `key_type` and delete `keytype`. Because of auto-detect it wouldn't break anything if anyone had changed it to want RSA.

WillCodeForCats: Yeah I like the way it works clearing `keytype`. Nothing breaks and auto-detect will work for everyone.

agners (approved):
LGTM! Thanks for your work and the extensive testing! :star_struck:

PR #3426, opened 2024-01-20
openthread_border_router: Fix Thread netif route metric

agners:
Make sure the Thread interface's route metric is low enough to take precedence over routes learned through IPv6 Neighbor Discovery protocol. This prevents from packets getting forwarded to the other border router instead of routing the packet on its own.

Also bump the OTBR to the latest version.


openthread_border_router/Dockerfile
```diff
@@ -55,6 +56,7 @@ RUN \
     && ./script/bootstrap \
     && patch -p1 < /usr/src/0001-Avoid-writing-to-system-console.patch \
     && patch -p1 < /usr/src/0002-rest-support-deleting-the-dataset.patch \
+    && patch -p1 < /usr/src/0001-openthread-set-netif-route-metric-lower.patch \
```
marcelveldt, line 59: shouldnt this be 0003 ?

agners: It numbering only really matter if the patches touch the same files. I've renamed it anyways, makes it look cleaner.

marcelveldt (approved):
LGTM

PR #3428, opened 2024-01-22
Clean up Genie add-on

frenck:
This removes the, no longer functional, Genie/Almond add-on from our add-on store.


agners (approved):

bcutter:
Very interestingly for the last (I think it started with 2024.4.x) supervisor releases, for an unknown reason supervisor first removes and than downloads that addon again (which renders the system almost useless as it peaks load and cpu usage at max for 15 to 20 minutes) like today when updating to supervisor 2024.5.1 release:

```
2024-05-09 11:35:23.839 INFO (MainThread) [supervisor.docker.interface] Attaching to homeassistant/aarch64-addon-almond with version 3.0.0
2024-05-09 11:35:26.268 INFO (SyncWorker_3) [supervisor.docker.manager] Removing image homeassistant/aarch64-addon-almond with latest
2024-05-09 11:35:26.274 INFO (SyncWorker_3) [supervisor.docker.manager] Removing image homeassistant/aarch64-addon-almond with 3.0.0
2024-05-09 11:35:36.018 INFO (MainThread) [supervisor.docker.interface] Downloading docker image homeassistant/aarch64-addon-almond with tag 3.0.0.
```

Glad I found this PR for the confirmation this addon is actually dead. So I manually removed that orphaned addon now, just leaving this note here in case someone else runs into the same situation.

agners:
Hm, that is most likely related to https://github.com/home-assistant/supervisor/pull/4991 which changed how/when add-on images get downloaded.

This seems like an issue to me. Did this only happen once? Or on every start since the new Supervisor?

/cc @mdegat01

bcutter:
This happened two times: when updating Supervisor to 2024.4.4 (from 2024.4.0) and to 2024.5.1 (from 2024.4.4).

I can't judge if that referenced Supervisor change really is associated to this behavior.

I also won't be able to provoke this for the next Supervisor update, as I as mentioned removed that addon. It is legacy and was the only one during Supervisor update, which is why I expect to never see this "supervisor update consumes system resources for 15 minutes because it irrationally deletes and redownloads a legacy addon image" issue again.

PR #3429, opened 2024-01-23
silabs-multiprotocol: Downgrade GSDK back to v4.3.1

puddly:
It looks like GSDK 4.4.0 created more issues than it solved, unfortunately.

This PR downgrades the addon back to 4.3.1 but also disables Zigbee Green Power (the source of all the crashes I was able to reproduce) and implements a firmware watchdog.


puddly:
I've locally tested this addon on both the Yellow and SkyConnect, no problems so far.

silabs-multiprotocol/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.4.4
+- Revert back to Silicon Labs Gecko SDK 4.3.1 while 4.4.0 instability is investigated.
+- Backport firmware modifications for improved stability.
+
```
agners, line 6: I see that you've also bumped the SiKabs flasher, probably worth mentioning.

```suggestion
- Bump universal SiLabs flasher to 0.0.17

```

agners (approved):
LGTM

PR #3433, opened 2024-01-24
letsencrypt: Add netcup configuration example via DNS

Patrick-DE:
Hi,

To make requesting certificates for Netcup users more straightforward, I decided to add an example configuration.
This should increase the visibility for the mandatory `propagation_seconds` flag.
Without that flag letsencrypt will not be able to verify the DNS entry due to the slow DNS update of Netcup.

Kr Patrick


agners (approved):

agners:
LGTM, thanks!

PR #3438, opened 2024-01-27
openthread_border_router: Enable DNS resolver and increase TX power

agners:
Enable DNS when NAT64 is enabled.

The firmware's default TX power of 0dBm is rather low. Set a the default TX power to 6dBm. Initial tests show that link quality increases quite a bit. To avoid asymmetric link quality the TX power from the controller should not exceed that of what other Thread routers devices typically use.

Update to the latest Universal Silicon Labs flasher 0.0.17.

Lastly also bump to the latest upstream OTBR.

### DIFF
--- a/openthread_border_router/rootfs/etc/s6-overlay/scripts/otbr-agent-configure.sh
+++ b/openthread_border_router/rootfs/etc/s6-overlay/scripts/otbr-agent-configure.sh
@@ -7,3 +7,6 @@ if bashio::config.true 'nat64'; then
     bashio::log.info "Enabling NAT64."
     ot-ctl nat64 enable
 fi
+
+# Set higher transmit power
+ot-ctl txpower 6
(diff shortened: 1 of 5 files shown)


openthread_border_router/rootfs/etc/s6-overlay/scripts/otbr-agent-configure.sh
```diff
@@ -7,3 +7,6 @@ if bashio::config.true 'nat64'; then
     bashio::log.info "Enabling NAT64."
     ot-ctl nat64 enable
 fi
+
+# Set higher transmit power
```
PaulSD, line 12: Would probably make sense to add a comment here with the notes you put in the PR:
This increases link quality significantly.
`To avoid asymmetric link quality the TX power from the controller should not exceed that of what other Thread routers devices typically use.`

agners: Make sense, added that as a comment.

sairon (approved):

Issue #3427, opened 2024-01-21
DuckDNS update via IPv6 not possible because container has no IPv6 access


fzakfeld:
### Describe the issue you are experiencing

The official DuckDNS addon supports both legacy IPv4 and IPv6 updates. 

The config has an `ipv6` field that lets you specify an interface or a URL to retrieve the current IPv6 address of the host.

> If you specify a URL here, contents of the resource it points to will be fetched and used as the address. This enables getting the address using a service like https://api6.ipify.org/ or https://ipv6.text.wtfismyip.com/

But everything runs inside a container which doesn't have the regular host's network. In fact, the container only has Legacy IP networking available.

```
➜  ~ docker exec -it addon_core_duckdns bash
bash-5.1# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
52: eth0@if53: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP
    link/ether 02:42:ac:1e:21:02 brd ff:ff:ff:ff:ff:ff
    inet 172.30.33.2/23 brd 172.30.33.255 scope global eth0
       valid_lft forever preferred_lft forever
bash-5.1#
```

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

DuckDNS

### What is the version of the add-on?

1.15.0

### Steps to reproduce the issue

1. Set up the addon with the `ipv6` option using a URL, e.g. https://ipv6.text.wtfismyip.com/

## System Information

version | core-2024.1.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.11.6
os_name | Linux
os_version | 6.1.63-haos-raspi
arch | aarch64
timezone | Europe/Berlin
config_dir | /config

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 11.3
-- | --
update_channel | stable
[...]

PR #3442, opened 2024-01-30
DuckDNS Use host network

marvinKaster:
DuckDNS access can't make a connection via IPv6 as docker containers do not allow access. In consequence, it is impossible to update the ipv6 address.

Using the host network fixes this issue.

See issue #3427

### DIFF
--- a/duckdns/DOCS.md
+++ b/duckdns/DOCS.md
@@ -84,9 +84,12 @@ a service like https://api.ipify.org/ or https://ipv4.text.wtfismyip.com
 ### Option: `ipv6` (optional)
 
 By default, Duck DNS will auto detect your IPv6 address and use that.
-This option allows you to specify an interface name from which the IPv6 address should be retrieved.
-You can also override the auto-detection and specify an
-IPv6 address manually.
+This option allows you to specify an interface name from which the IPv6
+address should be retrieved. Use `default` to choose the primary network
+interface automatically.
+
+If you set this option to an IPv6 address, this static address will be used
+instead.
 
 Retrieving the IPv6 address over a URL is not supported.
 
--- a/duckdns/DOCS.md
+++ b/duckdns/DOCS.md
@@ -84,12 +84,11 @@ a service like https://api.ipify.org/ or https://ipv4.text.wtfismyip.com
 ### Option: `ipv6` (optional)
 
 By default, Duck DNS will auto detect your IPv6 address and use that.
-This option allows you to override the auto-detection and specify an
+This option allows you to specify an interface name from which the IPv6 address should be retrieved.
+You can also override the auto-detection and specify an
 IPv6 address manually.
 
-If you specify a URL here, contents of the resource it points to will be
-fetched and used as the address. This enables getting the address using
-a service like https://api6.ipify.org/ or https://ipv6.text.wtfismyip.com
+Retrieving the IPv6 address over a URL is not supported.
 
 ### Option: `token`
 
--- a/duckdns/data/run.sh
+++ b/duckdns/data/run.sh
@@ -65,8 +65,8 @@ fi
 # Run duckdns
 while true; do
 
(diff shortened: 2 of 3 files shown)


duckdns/data/run.sh
```diff
@@ -65,8 +65,8 @@ fi
 # Run duckdns
 while true; do
 
-    [[ ${IPV4} != *:/* ]] && ipv4=${IPV4} || ipv4=$(curl -s -m 10 "${IPV4}")
-    [[ ${IPV6} != *:/* ]] && ipv6=${IPV6} || ipv6=$(curl -s -m 10 "${IPV6}")
+    [[ ${IPV4} != *:/* ]] && ipv4=${IPV4} || ipv4=$(curl -s -m 10 "${IPV4}") || bashio::log.warning "Couldn't retrieve ipv4 from server"
+    [[ ${IPV6} != *:/* ]] && ipv6=${IPV6} || ipv6=$(curl -s -m 10 "${IPV6}") || bashio::log.warning "Couldn't retrieve ipv6 from server"
 
     # Get IPv6-address from host interface
```
agners, line 71: From a security perspective, it would be nicer if the add-on does not require host network.

Is the IPv6 address retrial from the host interface in the following lines not working for you?

marvinKaster: No, unfortunately, `curl api6.ipify.org` cannot connect to the ipv6 server without the host network. 
In that case, only `curl api.ipify.org` works for the ipv4 address.

agners: Yes I get that. And I guess this line should be deleted, as it can't work.

However, following this code, there is a section which gets the IP address via API from the Supervisor:
https://github.com/home-assistant/addons/blob/master/duckdns/data/run.sh#L71-L82

Is this code not working for you?

marvinKaster: I see. Thank you. Yes, It works now when I set ipv6 to `default` in the configuration. Maybe you can remove the information about the URL and IPv6 from the documentation: 
`If you specify a URL here, contents of the resource it points to will be fetched and used as the address. This enables getting the address using a service like https://api6.ipify.org/ or https://ipv6.text.wtfismyip.com/` 

Moreover, it would be helpful to get any warnings in the logs as to why the addon is not working. I updated the branch with more detailed warnings. Furthermore, I changed that the addon uses the default interface automatically when nothing is specified. Maybe it makes sense to include these changes.

duckdns/DOCS.md
```diff
@@ -84,12 +84,11 @@ a service like https://api.ipify.org/ or https://ipv4.text.wtfismyip.com
 ### Option: `ipv6` (optional)
 
 By default, Duck DNS will auto detect your IPv6 address and use that.
-This option allows you to override the auto-detection and specify an
+This option allows you to specify an interface name from which the IPv6 address should be retrieved.
+You can also override the auto-detection and specify an
 IPv6 address manually.
```
agners, line 89:
```suggestion
This option allows you to specify an interface name from which the IPv6
address should be retrieved. Use `default` to choose the primary network
interface automatically.

If you set this option to an IPv6 address, this static address will be used
instead.
```

duckdns/data/run.sh
```diff
@@ -16,91 +16,126 @@ ALGO=$(bashio::config 'lets_encrypt.algo')
 
 # Function that performs a renew
 function le_renew() {
-    local domain_args=()
-    local domains=''
-    local aliases=''
+  local domain_args=()
```
agners, line 19: We default to 4 spaces for most scripts, I'd prefer if we keep it that way.

agners (changes requested):

duckdns/data/run.sh
agners: Oh yes, I see, this documentation is really wrong :see_no_evil: 

Thanks for the updated PR! 

Probably a new PR would have been easier here :sweat_smile:  But if you stick with this one, can you also update the title and description?

marvinKaster: Thanks. I incorporated the changes you suggested and opened a new pull request #3460

PR #3454, opened 2024-02-09
configurator: Update to Alpine 3.19

frenck:
SSIA, this PR updates Alpine Linux used in the file editor add-on to version 3.19.


agners (approved):

configurator/Dockerfile
```diff
@@ -1,6 +1,8 @@
 ARG BUILD_FROM
 FROM $BUILD_FROM
 
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
```
agners, line 4: I was actually wondering how we should handle the new pip default behavior. It is the same on Debian 12/Bookworm.

I guess we did the same already anyways in the past, so :man_shrugging:

frenck: Agree, there are two possible paths:

- Handle it in our base image (global pip configuration we already have there)
- Use venvs

But that is more a choice we'd have to make as a project. For the community add-ons, I did the first. Mostly because I didn't want to break upstream project and I think the problem this feature tries to solve is less of an issue in the container world.

../Frenck

PR #3468, opened 2024-02-16
matter-server: Allow specifying a custom primary network interface

tripplet:
Allow specifying a custom primary network interface for advanced uses

### DIFF
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -30,15 +30,17 @@ if ! bashio::var.has_value "${server_port}"; then
     extra_args+=('--listen-address' "$(bashio::addon.ip_address)")
 fi
 
-if bashio::config.has_value "custom_primary_interface"; then
-  primary_interface=$(bashio::config 'custom_primary_interface');
-else 
-  primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')";
+primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')";
+
+# Try fallback method (e.g. in case NetworkManager is not available)
+if [ -z ${primary_interface} ]; then
+  primary_interface="ip --json route show default | jq --raw-output '.[0].dev'";
 fi
 
 if [ -z ${primary_interface} ]; then
     bashio::exit.nok "No primary network interface found!"
 fi
+
 bashio::log.info "Using '${primary_interface}' as primary network interface."
 
 # Send out discovery information to Home Assistant
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -34,7 +34,7 @@ primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.int
 
 # Try fallback method (e.g. in case NetworkManager is not available)
 if [ -z ${primary_interface} ]; then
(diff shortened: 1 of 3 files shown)


agners:
There was recently some discussion around that, the idea is to make such interface selection a bit more generic. See also #3416 and specifically https://github.com/home-assistant/addons/pull/3416#issuecomment-1924179143.

That said, I am not against adding such a feature as a temporary measure. However, I think we should clearly state what this exactly means. The Matter Server listens on all interfaces by default, and even specifying the primary interface doesn't change that. The primary interface is currently just used to scope link-local IPv6 addresses (see https://github.com/home-assistant-libs/python-matter-server/pull/501). In practice, the user should set the primary interface to the network the Android phone is running in, when using Android companion app for commissioning. In other usecases this option isn't even relevant.

tripplet:
For me the problem is I'm running a not supported supervised installation without NetworkManager.
So the matter server add-on does not work any more since 5.0.3 because of that.
I fully understand that this is a very niche (and shrinking) group of users.
I'm planing to move HA into a VM to get rid of all my workarounds but currently don't have the hardware and time to do that for the installations I maintain.

agners:
Hm, I see, so you looking more for a fallback for the non-working API call.

How about getting the network interface of the default gateway as a fallback? :thinking:  That might also be helpful for other situation, e.g. when the primary interface somehow is missing.

```
default_route=$(ip route show default 0.0.0.0/0)
default_interface=$(echo $default_route | grep -oP 'dev \K\S+')
```

tripplet:
Yeah or something like this
```bash
ip --json route show default | jq --raw-output '.[0].dev'
```
But I fully understand that NetworkManager is supposed to work and is the default and supported API.
But a fallback if the NetworkManager call gives no result might be a good idea (and also fixes my problem)

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -30,10 +30,17 @@ if ! bashio::var.has_value "${server_port}"; then
     extra_args+=('--listen-address' "$(bashio::addon.ip_address)")
 fi
 
-primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')";
+
+# Try fallback method (e.g. in case NetworkManager is not available)
+if [ -z ${primary_interface} ]; then
+  primary_interface="$(ip --json route show default | jq --raw-output '.[0].dev')";
```
agners, line 37: Let's add an info or warning message here, just so we know where the primary interface came from.

tripplet: Makes sense, I think warning is the appropriate level?
Because the intended mode of operation is that NetworkManger gives the correct primary interface

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -30,10 +30,18 @@ if ! bashio::var.has_value "${server_port}"; then
     extra_args+=('--listen-address' "$(bashio::addon.ip_address)")
 fi
 
-primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')";
+
+# Try fallback method (e.g. in case NetworkManager is not available)
+if [ -z ${primary_interface} ]; then
+  bashio::log.warn 'Trying fallback method to determine primary interface'
+  primary_interface="$(ip --json route show default | jq --raw-output '.[0].dev')";
```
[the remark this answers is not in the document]
tripplet: No, just a habit, I removed them

agners (approved):

PR #3471, opened 2024-02-16
openthread_border_router: Fix container shutdown during startup

agners:
When otbr-agent fails during startup properly shutdown again works best by using exit code 125 in the finish script. This indicates a permanent failure for this service. This makes sure that the container shuts down which is the behavior the Supervisor expects.


frenck (approved):
TIL!

../Frenck

PR #3476, opened 2024-02-18
nginx_proxy: Support integration with letsencrypt addon

mraerino:
This turns out pretty handy when both addons are used alongside each other.

The only gotcha that people still have to care about is that you need to run the letsencrypt addon on port 80 once to get the initial cert material and then change its port and install nginx, because nginx won't start without a cert. Not sure if that could be documented better.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -27,6 +27,7 @@ There are two options to obtain certificates.
 
 - Requires Port 80 to be available from the internet and your domain assigned to the externally assigned IP address
 - Doesn’t allow wildcard certificates (*.yourdomain.com).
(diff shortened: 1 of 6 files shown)


mraerino:
@agners should we roll this into the same release as the IPv6 change?

agners:
I rather prefer if we merge them separately with a separate version number, so we could easily roll-back (especially the IPv6 change).

Now in case the IPv6 change needs more discussions, maybe it is easier if we roll out this one first. You'd need to add changelog entries and bump the version in `config.yaml` here too, and then rebase the IPv6 change afterwards. Sorry for the inconvenience, the add-on core release process is not ideal currently :cry:

PR #3487, opened 2024-02-26
Update wyoming-piper to 1.5.0

synesthesiam:
## 1.5.0

- Add voices for Persian: amir, gyro
- Add voice for Slovenian: artur
- Add voice for Turkish: fettah
- Add voices for French: tom, mls
- Add voice for Dutch: mls
- Add voice for German: mls


balloob (approved):

Je-sse:
I did some tests with the available Dutch (nl_NL) TTS voices in the past, but no matter what I wrote down, I could not understand a word of what was said (using the service or the Try Voice button in the Voice Assistant configuration menu).
As a workaround I'm using the Dutch/Belgian (nl_BE) TTS voices. I don't have any problems with them.

As this update included a new Dutch voice, I gave it another try. But it's still impossible to understand a word.

I'm not sure if this belongs here or if I should open a Bug Report Form, but is this a known issue?

Je-sse:
Thank you for the explanation! I did some tests with longer sentences and they are still not fluent, but they are now normal spoken words.

synesthesiam:
Looking through my speech data, I realized that I have very little Dutch audio that's not nl_BE or from audiobooks. I guess we need to collect more data 😄

Je-sse:
Is this something I can help with? I found the [OpenSLR](https://openslr.org/94) and [Mozilla](https://commonvoice.mozilla.org/en/datasets) datasets, but I think you already know them..

PR #3493, opened 2024-02-29
matter_server: Bump Python Matter server to 5.7.0

agners:
Bump the Python Matter server to 5.7.0. Also add Matter SDK log level as separate (optional) configuration. This allows to select the Matter SDK log level separately from the server logging level.

### DIFF
--- a/matter_server/config.yaml
+++ b/matter_server/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 5.2.0
+version: 5.3.0
 slug: matter_server
 name: Matter Server
 description: Matter WebSocket Server for Home Assistant Matter support.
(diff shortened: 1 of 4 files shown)


marcelveldt (approved):
Looks good!

matter_server/config.yaml
```diff
@@ -20,9 +20,11 @@ image: homeassistant/{arch}-addon-matter-server
 init: false
 options:
   log_level: info
+  log_level_sdk: error
   beta: false
 schema:
   log_level: list(debug|info|warning|error|critical)
```
MartinHjelmare, line 26: Side note: Shouldn't this option be optional?

Issue #3491, opened 2024-02-28
openthread_border_router: crash at otbr::Mdns::PublisherMDnsSd::UnsubscribeService


agners:
### Describe the issue you are experiencing

During regular operation, the OTBR suddenly crashes with the following crash dump:

```
otbr-agent: ../../src/mdns/mdns_mdnssd.cpp:951: virtual void otbr::Mdns::PublisherMDnsSd::UnsubscribeService(const string&, const string&): Assertion `it != mSubscribedServices.end()' failed.
otbr-agent[184]: 10d.22:43:47.271 [C] Platform------: ------------------ BEGINNING OF CRASH -------------
otbr-agent[184]: 10d.22:43:47.272 [C] Platform------: *** FATAL ERROR: Caught signal: 6 (Aborted)
otbr-agent[184]: 10d.22:43:47.315 [C] Platform------: # 0: /usr/sbin/otbr-agent(+0x220ab8) [0x557efb0ab8]
otbr-agent[184]: 10d.22:43:47.315 [C] Platform------: # 1: /usr/sbin/otbr-agent(+0x220bfc) [0x557efb0bfc]
otbr-agent[184]: 10d.22:43:47.320 [C] Platform------: # 2: linux-vdso.so.1 __kernel_rt_sigreturn+0x0 [0x949957bc]
otbr-agent[184]: 10d.22:43:47.321 [C] Platform------: # 3: /lib/aarch64-linux-gnu/libc.so.6 gsignal+0xdc [0x944f0eac]
otbr-agent[184]: 10d.22:43:47.321 [C] Platform------: # 4: /lib/aarch64-linux-gnu/libc.so.6 abort+0x108 [0x944ddaa0]
otbr-agent[184]: 10d.22:43:47.321 [C] Platform------: # 5: /lib/aarch64-linux-gnu/libc.so.6(+0x2d478) [0x7f944ea478]
otbr-agent[184]: 10d.22:43:47.321 [C] Platform------: # 6: /lib/aarch64-linux-gnu/libc.so.6(+0x2d4dc) [0x7f944ea4dc]
[...]

PR #3494, opened 2024-02-29
openthread_border_router: OTBR bump and Debian Bookworm update

agners:
This updates the OTBR to the latest upstream git version. In particular, this should address issue #3491.

Also update the base image to Debian Bookworm.

Fixes: #3491

### DIFF
--- a/openthread_border_router/Dockerfile
+++ b/openthread_border_router/Dockerfile
@@ -90,7 +90,8 @@ RUN \
         && cd build/otbr/ \
         && ninja \
         && ninja install) \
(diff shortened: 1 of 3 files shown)


openthread_border_router/Dockerfile
```diff
@@ -90,7 +90,8 @@ RUN \
         && cd build/otbr/ \
         && ninja \
         && ninja install) \
-    && pip install universal-silabs-flasher==${UNIVERSAL_SILABS_FLASHER} \
+    && pip install --break-system-packages \
```
marcelveldt, line 93: what does this break-system-packages do ?

PR #3500, opened 2024-03-05
whisper: add prompt option

AtomToast:
I have a quite uncommon name, that whisper currently struggles to make any sense of. Adding a prompt, as suggested [here](https://github.com/openai/whisper/discussions/963) and [here](https://github.com/openai/whisper/discussions/328), made whisper actually capable of recognizing it. 

Unfortunately for me still only with the medium model, which doesn't really run well on my raspi 4, however I thought it is still worth sharing.

### DIFF
--- a/whisper/DOCS.md
+++ b/whisper/DOCS.md
@@ -55,6 +55,16 @@ Number of candidates to consider simultaneously during transcription (see [beam 
 
 Increasing the beam size will increase accuracy at the cost of performance.
 
+### Option: prompt
+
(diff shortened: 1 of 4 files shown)


synesthesiam:
Initial prompt will be added in this PR: https://github.com/home-assistant/addons/pull/3507

@agners This PR can be closed

agners:
> @agners This PR can be closed

Ok!

Thanks @AtomToast for bringing this up!

PR #3501, opened 2024-03-06
Bump zwave-js to 12.4.4

raman325:
Changelog: https://github.com/zwave-js/node-zwave-js/releases/tag/v12.4.4

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## 0.4.5
+
+### Bug fixes
+
+- NVM backups can now be restored onto 800 series controllers
+
+### Config file changes
+
+- Use Color Switch V2 for Inovelli LZW42
+- Correct Zooz ZEN1x timer config params
+
+### Detailed changelogs
+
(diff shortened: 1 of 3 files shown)


zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Changelog
 
+## 0.4.5
+
+### Bug fixes
+
+- NVM backups can now be restored onto 800 series controllers
```
MartinHjelmare, line 7: I think we should prefix the fixes and feature with the context of the change, ie `Z-Wave JS:` if it was done in the driver, `Add-on:` if it was done in the add-on, etc.

We did this some time ago but stopped doing it for some reason. It's hard to read the changelog without that and know when we changed the add-on and when we just bumped the driver or server.

MartinHjelmare (changes requested):

zwave_js/CHANGELOG.md
raman325: agreed, I seem to have forgotten to do it properly. Hopefully I caught everything

MartinHjelmare (approved):
Thanks!

PR #3507, opened 2024-03-10
Update to wyoming-faster-whisper 2.0.0

synesthesiam:
- Add more models for `model` option
- Add "custom" option for `model` that will use `custom_model` instead
- Add `custom_model` that can be a path to a model directory or HuggingFace Hub model ID
- Use faster-whisper PyPI package

### DIFF
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.0.2
+version: 2.0.0
 slug: whisper
 name: Whisper
 description: Speech-to-text with Whisper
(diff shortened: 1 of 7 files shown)


whisper/config.yaml
```diff
@@ -14,13 +14,18 @@ backup_exclude:
...
   model: tiny-int8
+  custom_model: ""
   language: en
   beam_size: 1
+  debug_logging: false
 schema:
-  model: list(tiny-int8|tiny|base|base-int8|small-int8|small|medium-int8)
+  model: |
+    list(tiny-int8|tiny|tiny.en|base-int8|base|base.en|small-int8|distil-small.en|small|small.en|distil-medium.en|medium-int8|medium|medium.en|large|large-v1|distil-large-v2|large-v2|large-v3|custom)
+  custom_model: str
```
agners, line 24: This probably should be an optional option:

```suggestion
  custom_model: str?
```

synesthesiam: I've set this to optional and added an error case in the run script so the user knows what went wrong if they selected "Custom" for `model` but don't set `custom_model` to anything.

synesthesiam:
@agners I have one more update before this can be merged

whisper/rootfs/etc/s6-overlay/s6-rc.d/whisper/run
```diff
@@ -12,6 +12,10 @@ model="$(bashio::config 'model')"
 if [ "${model}" = 'custom' ]; then
     # Override with custom model
     model="$(bashio::config 'custom_model')"
+    if [ -z "${model}" ]; then
+      bashio::log.fatal "Custom model is not set"
+      exit 1
```
agners, line 17: There is a shorthand for this:
```suggestion
      bashio::exit.nok "Custom model is not set"
```

synesthesiam: Great! I added that.

synesthesiam:
Good to go :+1:

agners (approved):

PR #3508, opened 2024-03-11
samba: Handle passwords with backslash correctly

agners:
Make sure backslash in a password is not interpreted as an escape sequence.


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3517, opened 2024-03-19
nginx_proxy: Add map_hash_bucket_size

DaAwesomeP:
The `map_hash_bucket_size` option must be specified before the first `map`. Since the first `map` comes before the user `include` in `nginx.conf`, it is not possible for the user to set this value. A larger bucket is required when matching long strings (such as user agents) in a `map`.


agners:
@DaAwesomeP we had to make an update to the add-on builder, as the old version is no longer supported by SigStore. Can you rebase this PR to current master?

DaAwesomeP:
@agners got it, just pushed again!

agners (approved):

PR #3520, opened 2024-03-20
Bump tellstick addon to alpine 3.15 and deprecate it

mdegat01:
This is the latest version of Alpine this addon builds on currently. The make file fails on every version of alpine >3.15 .

There is a PR with alpine 3.17 fixes (https://github.com/telldus/telldus/pull/21) but the project appears abandoned. Not clear how to handle this addon going forward.

**Decision:** This addon is deprecated. This will be the final update for the addon.

### DIFF
--- a/tellstick/CHANGELOG.md
+++ b/tellstick/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.0
+
+- Use Alpine 3.15
+
 ## 2.1.0
 
 - Update hardware configuration for Supervisor 2021.02.5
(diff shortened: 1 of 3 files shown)


agners:
> There is a PR with alpine 3.17 fixes ([telldus/telldus#21](https://github.com/telldus/telldus/pull/21)) but the project appears abandoned. Not clear how to handle this addon going forward.

Ugh I see. Yeah I tend to just deprecate/remove the add-on in that case.

@endor-force since you are the original author of that add-on, do you know if this is still in working order/in use? Is there an alternative to control these devices today?

agners (approved):

PR #3539, opened 2024-04-03
matter_server: Fix bashio::log.warn

tripplet:
Fixes bashio::log.warn
I have to admit I didn't test it after the review, not sure why I used log.*warn* instead of log.*warning*.


tripplet:
Done

agners (approved):

tripplet:
Thanks, works 😄

PR #3444, opened 2024-02-01
Add easyDNS provider for letsencrypt addon.

drachul:
This update adds support for easyDNS DNS challenge in the letsencrypt addon making use of the [easyDNS certbot plugin](https://github.com/easydns/certbot-dns-easydns), by way of the [easyDNS API](https://sandbox.rest.easydns.net:3001/).

PR #3556, opened 2024-04-11
letsencrypt: add Dynu DNS provider

McGiverGim:
This PR adds Dynu as DNS provider to Let's encrypt addon. It has been made in the same way than it was done here, so I hope it's correct: https://github.com/home-assistant/addons/pull/3444

It uses the https://pypi.org/project/certbot-dns-dynu/ addon for certbot.


mdegat01 (approved):
Looks good. I was initially concerned this was an beta plugin or something with the version `0.0.4` but the pip page says it is production/stable so I guess that's just what they choose.

mdegat01:
Ok I guess I have to take back my approval but its not the fault of your code. We can't add this right now because of this:
```
#7 24.97 The conflict is caused by:
#7 24.97     certbot-dns-dnsimple 2.7.4 depends on dns-lexicon>=3.14.1
#7 24.97     certbot-dns-dnsmadeeasy 2.7.4 depends on dns-lexicon>=3.14.1
#7 24.97     certbot-dns-dynu 0.0.4 depends on dns-lexicon==3.11.7
```
That's not going to work. `certbot-dns-dynu` is too restrictive in its dependencies. I don't know why it depends on `dns-lexicon==3.11.7` explicitly, general that's bad practice and a mistake. It should say a range of versions it supports, either >= like the others or an explicitly range from X to Y if there's a breaking change it can't handle.

Unfortunately this means we can't support this library in this state. You'll need to follow up with the developer and get its dependencies updated. Or else find an alternative library to use for dynu instead.

McGiverGim:
Ok, I understand. I will try to ask the developer to update it.

McGiverGim:
I've opened an issue to the developer, but there's a PR opened with the same change and it seems ignored. 
I've seen some forks that fix the issue, I will test them and update this PR with the one that seem to have more activity.

McGiverGim:
Modified the library for this one:  https://pypi.org/project/certbot-dns-dynu-dev/ 
I've seen that we have another option that is this: https://pypi.org/project/certbot-dns-dynudns/

I've some strange problem trying to execute the addons inside the docker container, so any extra test will be appreciated.

agners (approved):

PR #3562, opened 2024-04-12
MariaDB: Upgrade to Alpine 3.19

mdegat01:
Upgrade MariaDB addon to Alpine 3.19


frenck (approved):
Thanks, @mdegat01 👍 

../Frenck

PR #3567, opened 2024-04-16
mariadb: Enable auto-upgrade

agners:
After a major update, the mariadb-upgrade tool should be run (see https://mariadb.com/kb/en/mariadb-upgrade/). However, we don't have a specific post-upgrade script we can run.

Luckily, MariaDB has a feature to run the upgrade on it's own when needed. Enable the feature using the MARIADB_AUTO_UPGRADE environment variable.

### DIFF
--- a/mariadb/CHANGELOG.md
+++ b/mariadb/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.7.1
+
+- Enable MariaDB auto-upgrade
+
 ## 2.7.0
 
 - Update to Alpine 3.19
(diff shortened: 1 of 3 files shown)


agners:
The `MARIADB_AUTO_UPGRADE` environment variable is actually a feature of the official MariaDB Docker image, and doesn't work in our case (see also https://github.com/MariaDB/mariadb-docker/blob/master/10.6/docker-entrypoint.sh#L573).

It tuns out the errors I've been seeing were only printed on first startup, and are an artifact of order of execution (mariadb needs to get started before upgrading). See also https://github.com/home-assistant/addons/issues/3566#issuecomment-2059273520.

PR #3572, opened 2024-04-19
Silabs multiprotocol BT HCI support

zalatnaicsongor:
Silicon Labs has an example firmware that supports Multiprotocol (OpenThread+Zigbee+BLE) RCP. This PR adds the `bt_host_cpc_hci_bridge` app as per the `AN1333: Running Zigbee, OpenThread, and Bluetooth Concurrently on a Linux Host with a Multiprotocol Co-processor` application note to expose a HCI PTY device.

If the support is enabled and the correct firmware is uploaded to the coordinator stick, the following output should appear in the addon logs:
```
[20:25:52] INFO: Starting bt_host_cpc_hci_bridge...
[I] Silicon Labs | CPC-HCI bridge
[I] CPC successfully initialized with 'cpcd_0'
[I] CPC Bluetooth endpoint opened
[I] PTY device file opened at '/dev/pts/2'
```

The created PTY device can then be used by the host (I used the Advanced SSH & Web Terminal addon) by executing `hciattach /dev/pts/2 any 115200`. Home Assistant should automatically then recognise the bluetooth adapter.

bt


zalatnaicsongor:
If someone would like to run this as a separate addon, I can make one which only runs cpcd and `bt_host_cpc_hci_bridge`, also I can provide the BT capable firmware for the EasyIOT GW-04 1.2 stick.

agners:
This PR essentially does two things: Bump to Gecko SDK v4.4.2 and enables the BT HCI support.

Currently we avoid pouring more effort into the Multiprotocol effort as it didn't prove stable for a lot of people. Adding another variable (Bluetooth) certainly won't help here :see_no_evil:

zalatnaicsongor:
Yeah, technically this could work with the older version of the Gecko SDK (4.3.x) but v4.4.2 proved to be a lot more stable for me (the uptime is now an order of magnitude better than it was used to, currently the add-on has been running for 5 days).

BT HCI support is actually not very useful as EFRMG21 and MG1 chips only have one radio, and enabling BT will cause an extremely spotty and laggy zigbee connection, however if someone has spare coordinators available they could potentially use it as a makeshift BT dongle.

If there's no intention to pour more effort into the Multiprotocol addon, perhaps I could help maintaining it as for some weird reason I find it an interesting piece of work :D

agners:
> Yeah, technically this could work with the older version of the Gecko SDK (4.3.x) but v4.4.2 proved to be a lot more stable for me (the uptime is now an order of magnitude better than it was used to, currently the add-on has been running for 5 days).

The thing is, with the Gecko 4.4.0 release we got lots of reports that things broke for folks. So we've reverted in the past, see https://github.com/home-assistant/addons/pull/3429.

I remember did some testing with later 4.4.x releases, but we still encountered issue in our testing. So we never really move forward there.

We also have the problem that we don't have a good way to test things: There is no beta channel for add-ons, currently. So merging this would really deploy it to stable directly. According to [analytics](https://analytics.home-assistant.io/), we have still several thousands of installations of the SiLabs Multiprotocol add-on. Honestly, this is too high of a risk to break these folks installation.

> If there's no intention to pour more effort into the Multiprotocol addon, perhaps I could help maintaining it as for some weird reason I find it an interesting piece of work :D

Yeah I think transition this to a community effort makes sense at this point. I agree, it is interesting tech, and maybe it will become stable at one point/under certain combination.

It is fairly simple to create your own repository, see https://developers.home-assistant.io/docs/add-ons/repository. You might also want to have a look at Frenck's add-on development method (see https://github.com/hassio-addons/).

The community forum is probably a good place to then announce your repo to connect with folks interested in trying it out as well.

I am closing this PR as this is extending with Bluetooth is really out-of-scope for the official one.

Issue #3383, opened 2024-01-04
gandi dns not working


oweinsto68:
### Describe the issue you are experiencing

Hi, 
the letsencrypt addon is not working for gandi. Ive checked in the code and it only takes the sharing id parameter and doesn't read the gandi_api_key configuration parameter. By the way gandi_api_key is deprecated and should be replaced with dns_gandi_token

`# Gandi
elif [ "${DNS_PROVIDER}" == "dns-gandi" ]; then
    if bashio::config.exists 'dns.gandi_sharing_id'; then
        bashio::log.info "Use Gandi sharing ID"
        echo "dns_gandi_sharing_id = $(bashio::config 'dns.gandi_sharing_id')" >> "/data/dnsapikey"
    fi
    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" "/data/dnsapikey" "--${DNS_PROVIDER}-propagation-seconds" "${PROPAGATION_SECONDS}" )
`

the following code should fix it:
`elif [ "${DNS_PROVIDER}" == "dns-gandi" ]; then
    if bashio::config.exists 'dns.gandi_sharing_id'; then
        bashio::log.info "Use Gandi sharing ID"
        echo "dns_gandi_sharing_id = $(bashio::config 'dns.gandi_sharing_id')" >> "/data/dnsapikey"
    fi
    if bashio::config.exists 'dns.dns_gandi_token'; then
        bashio::log.info "Use Gandi dns_gandi_token"
        echo "dns_gandi_token = $(bashio::config 'dns.dns_gandi_token')" >> "/data/dnsapikey"
    fi
    if bashio::config.exists 'dns.gandi_api_key'; then
        bashio::log.info "Use Gandi gandi_api_key"
        echo "gandi_api_key = $(bashio::config 'dns.gandi_api_key')" >> "/data/dnsapikey"
    fi
    PROVIDER_ARGUMENTS+=("--authenticator" "${DNS_PROVIDER}" "--${DNS_PROVIDER}-credentials" "/data/dnsapikey" "--${DNS_PROVIDER}-propagation-seconds" "${PROPAGATION_SECONDS}" )
`

Can you please implement the fix?

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Almond

### What is the version of the add-on?

5.0.9

### Steps to reproduce the issue

[...]

PR #3581, opened 2024-04-25
letsencrypt: Fix and extend Gandi DNS challenge

agners:
Fix Gandi DNS challenge using an API key. Also add support for token authentication.

Fixes: #3383


sairon (approved):

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -115,6 +115,14 @@ elif [ "${DNS_PROVIDER}" == "dns-gandi" ]; then
         bashio::log.info "Use Gandi sharing ID"
         echo "dns_gandi_sharing_id = $(bashio::config 'dns.gandi_sharing_id')" >> "/data/dnsapikey"
     fi
+    if bashio::config.exists 'dns.gandi_token'; then
+        bashio::log.info "Use Gandi gandi_token"
+        echo "dns_gandi_token = $(bashio::config 'dns.gandi_token')" >> "/data/dnsapikey"
+    fi
+    if bashio::config.exists 'dns.gandi_api_key'; then
```
disconn3ct, line 122: Should this be an 'else' to avoid strange behavior? (Or an error if both are provided.) It looks like token is preferred according to https://api.gandi.net/docs/authentication/. Alternately maybe these should be reversed to give the token precedence.

agners: I guess we could be smarter here, but not sure what the correct smartness is. E.g. the sharing id I think is only valid when an API key is present.

The plug-in just seem to read all of them and forward: https://github.com/obynio/certbot-plugin-gandi/blob/master/certbot_plugin_gandi/main.py#L82-L85. So why not doing it here too :man_shrugging: :sweat_smile:

disconn3ct: It only forwards one here. Both clauses point to the same target, and if both are set the deprecated one is used because it is last.

agners: Not sure what you mean, if all three are set, it forwards all three, at least with this change:

https://github.com/home-assistant/addons/blob/master/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L114-L125.

Maybe it is easiest if you open a PR what you'd suggest to change, then we can discuss an actual suggestion.

PR #3599, opened 2024-05-16
matter_server: Bump Python Matter server to 6.0.0

agners:


marcelveldt (approved):

matter_server/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 5.6.0
+version: 6.0.0
```
marcelveldt, line 2: woohoo, we finally have the version number equal ;-)

PR #3601, opened 2024-05-18
Add abilty to config cache size for dnsmasq addon

dabacon:
This adds the ability to set the cache size for the Dnsmasq addon.

The default cache size is 150, which can be too small when one has a lot of local servers and devices constantly calling out to cloud services.   Here is a user request for adjusting the cache size https://community.home-assistant.io/t/dnsmasq-increase-cache-size/382020

## Summary by CodeRabbit

- **New Features**
  - Introduced a new option `cache_size` allowing users to set the size of the Dnsmasq cache, enhancing caching performance.

- **Updates**
  - Upgraded the underlying system to Alpine 3.19 for improved stability and performance.

### DIFF
--- a/dnsmasq/CHANGELOG.md
+++ b/dnsmasq/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.8.0
+
+- Add cache_size option
+
 ## 1.7.0
 
 - Update to Alpine 3.19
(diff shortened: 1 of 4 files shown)


sairon (approved):
Looks good, 150 is indeed the default in dnsmasq itself, so it wouldn't introduce any breaking changes.

ScottG489:
The option is [cache-size](https://github.com/imp/dnsmasq/blob/master/dnsmasq.conf.example#L575) not [cache](https://github.com/home-assistant/addons/pull/3601/files#diff-038b2c26a4e2891216ce70518fff1cc7a06a72705eb8f6ff93a1440d0cd5f47fR12).

Are we not even spinning up the add-on for a sanity check as part of our testing process?

PR #3605, opened 2024-05-22
Add distil-large-v3 model to whisper addon

jamesmyatt:
Latest distilled large Whisper model

* https://huggingface.co/distil-whisper/distil-large-v3
* https://huggingface.co/Systran/faster-distil-whisper-large-v3


sairon (approved):
Looks good, just updated the changelog so it can be merged straight away. Thanks!

PR #3615, opened 2024-05-27
Increase mqtt max_queued_messages to 8192

bdraco:
The default of 1000 was not sufficent when the user has more than 1000 entities and startup is so busy that is cannot read the buffer before it fills up and the broker starts dropping messages.

```
2024-05-26 21:33:43: New client connected from 172.30.32.1:49641 as 6RPEl0PbDg8Z5fNLHymeL4 (p2, c1, k60, u'homeassistant').
2024-05-26 21:33:43: Outgoing messages are being dropped for client 6RPEl0PbDg8Z5fNLHymeL4.
```

While we have increased the mqtt receive buffer in core, this only solves the problem on fast systems.

This value can still be overridden with a config file in `/share/mosquitto` if a higher value (or lower) is desired.

I didn't change it to 0 (unlimited) to handle the case where they expose the MQTT server to the public internet (this is unwise but someone will).  

I also didn't make it configurable since having the value match close to what Home Assistant can support made the most sense. Making it easier to configure would likely lead to more issues in the future since it would be hard to increase the default again without having to adjust any configuration that had been overridden manually.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration parameter `max_queued_messages` to manage the upper limit of queued messages during busy startup periods.

- **Updates**
  - Updated Mosquitto version from `6.4.0` to `6.4.1`.
  - Increased default value of `max_queued_messages` to 8192.

### DIFF
--- a/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
+++ b/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
@@ -13,6 +13,12 @@ log_timestamp_format %Y-%m-%d %H:%M:%S
 persistence true
 persistence_location /data/
 
+# Limits
+# max_queued_messages is effectively the upper limit of
+# the number of entities on Home Assistant is startup
+# is busy and cannot ready messages fast enough
+max_queued_messages 8192
+
 # Authentication plugin
 auth_plugin /usr/share/mosquitto/go-auth.so
 auth_opt_backends files,http
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.4.1
+
+- Increase default max_queued_messages to 8192 to fix dropped messages during Home Assistant startup
+
 ## 6.4.0
 
 - Update mosquitto to 2.0.18
--- a/mosquitto/config.yaml
+++ b/mosquitto/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 6.4.0
+version: 6.4.1
 slug: mosquitto
 name: Mosquitto broker
 description: An Open Source MQTT broker


mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
```diff
@@ -13,6 +13,12 @@ log_timestamp_format %Y-%m-%d %H:%M:%S
 persistence true
 persistence_location /data/
 
+# Limits
+# max_queued_messages is effectively the upper limit of
+# the number of entities on Home Assistant is startup
```
bdraco, line 18:
```suggestion
# the number of entities on Home Assistant if startup
```

mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
```diff
@@ -13,6 +13,12 @@ log_timestamp_format %Y-%m-%d %H:%M:%S
...
+# is busy and cannot ready messages fast enough
```
bdraco, line 19:
```suggestion
# is busy and cannot read messages fast enough
```

jbouwh (approved):
LGTM,
Thanx @bdraco

bdraco:
I need to do more testing with this but its too late here to finish up today

If someone wants to try it, testing instructions are here https://developers.home-assistant.io/docs/add-ons/testing

bdraco:
```
# docker exec -it addon_local_mosquitto /bin/bash
root@local-mosquitto:/# grep max /etc/mosquitto/mosquitto.conf 
# max_queued_messages is effectively the upper limit of
max_queued_messages 8192
```

Looks good on my other system as well

bdraco:
no drops in the log there either now

mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
```diff
@@ -13,6 +13,12 @@ log_timestamp_format %Y-%m-%d %H:%M:%S
 persistence true
 persistence_location /data/
 
+# Limits
+# max_queued_messages is effectively the upper limit of
+# the number of entities on Home Assistant if startup
+# is busy and cannot read messages fast enough
+max_queued_messages 8192
```
bdraco, line 20: https://mosquitto.org/man/mosquitto-conf-5.html

> The maximum number of QoS 1 or 2 messages to hold in the queue (per client) above those messages that are currently in flight. Defaults to 1000. Set to 0 for no maximum (not recommended). See also the queue_qos0_messages and max_queued_bytes options.

bdraco:
@coderabbitai the default for `max_queued_messages` is not 0, its 1024. Please review https://mosquitto.org/man/mosquitto-conf-5.html

bdraco:
https://github.com/coderabbitai the default for max_queued_messages is not 1000 not 1024

bdraco:
@coderabbitai the default for max_queued_messages is 1000 not 1024

bdraco:
Left this in a reload loop overnight.  No more issues in HA. No more dropped messages.

pvizeli (approved):
Need a version pump + changelog

bdraco:
Thanks

bdraco:
x86_64 is up, and I updated a few systems. All good. Issue is resolved.

PR #3663, opened 2024-06-25
letsencrypt: Fix i386 build

agners:
It seems that that the latest cloudflare package 3.0.x pulls in new dependencies, notable pydantic. This causes issues when building i386. Unbreak the build for now by explictly request the cloudflare package which has been selected previously according to logs (2.9.11).


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3677, opened 2024-07-01
openthread_border_router: Deassert DTR/RTS lines for USB devices using these for BSL

tl-sl:
It seems most USB devices based on TI CC2652 (except for ZBDongle-P which is not affected) have `DTR` and `RTS` lines directly connected to the `reset` and `bootloader activation` pins.  For these devices if these lines are asserted at startup then these devices fail to startup as they are in a stuck state.

This patch ensures both DTR and RTS are deasserted on startup for devices where hw flow control has been disabled. Over in Zigbee land, both zigpy and zigbee-herdsmen must be doing the same, as the dongles work without issues in both ZHA and Zigbee2MQTT.

## Summary by CodeRabbit

- **New Features**
	- Improved startup reliability for USB radio devices by ensuring DTR and RTS lines are deasserted when flow control is disabled.
- **Version Update**
	- Updated the version number to 2.9.0.
- **Documentation**
	- Added details of the new patch and its impact in the changelog.
- **Configuration**
	- Incorporated the new patch file into the Dockerfile.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.9.0
+
+- Deassert DTR/RTS lines for USB devices using these for BSL
+
 ## 2.8.0
 
 - Bump to OTBR POSIX version 41474ce29a (2024-06-21 08:41:31 -0700)
@@ -15,6 +19,7 @@
 - Add support for network sockets using socat
 
 ## 2.5.1
+
 - Support Home Assistant Connect ZBT-1.
 
 ## 2.5.0
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.9.0
 
-- Deassert DTR/RTS lines for USB devices using these for BSL
+- Avoid triggering reset/boot loader on TI CC2652 based devices
 
 ## 2.8.0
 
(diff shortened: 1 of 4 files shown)


agners:
Thanks for your patch!

Ideally I'd see this patch upstream in https://github.com/openthread/ot-br-posix/. I try to push pretty much everything I have to patch on the OTBR upstream to lower maintenance long-term. I guess this particular instance could be a bit problematic as it could have unwanted side effects for other controllers :thinking: . But then again, if hardware flow control is not required, it should be safe. Maybe it would be better to be behind a command line flag, not sure.

In any case, before merging it here, I'd like to see at least a PR to the upstream repository :pleading_face: .

darkxst:
I have posted it upstream, but need to sort out the CLA there still.
https://github.com/openthread/openthread/pull/10454

openthread_border_router/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.9.0
+
+- Deassert DTR/RTS lines for USB devices using these for BSL
```
agners, line 5:
```suggestion
- Avoid triggering reset/boot loader on TI CC2652 based devices
```

tl-sl: Yes, good point, I was a bit technical there ;)

agners (approved):
Also checked against Connect ZBT-1, seems to work fine still :+1:

tl-sl:
This has been merged upstream also (with slightly different implementation to improve cross-platform support)
https://github.com/openthread/openthread/commit/cdbb9e5342bd8e3fa88d3b9bde7e7c23e5504ac0

agners:
Perfect, thanks for the update!

PR #3681, opened 2024-07-03
matter_server: Bump Python Matter server to 6.2.1

agners:
Bump Python Matter Server to [6.2.1](https://github.com/home-assistant-libs/python-matter-server/releases/tag/6.2.1)

This adds a new configuration option enable_test_net_dcl which enables PAA root certificates and other device information from test-net DCL.

## Summary by CodeRabbit

- **New Features**
  - Expanded logging capabilities with new `log_level_sdk` option.
  - Introduced `beta` option to install the latest beta version on startup.
  - Added `enable_test_net_dcl` option to use test-net DCL for development and testing.

- **Chores**
  - Bumped version of Python Matter Server to `6.2.1`.
  - Updated image versions for `aarch64` and `amd64` architectures in build configuration.

### DIFF
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.2.0
+
+- Bump Python Matter Server to [6.2.0](https://github.com/home-assistant-libs/python-matter-server/releases/tag/6.2.0)
+
 ## 6.1.2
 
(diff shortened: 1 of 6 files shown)


marcelveldt (approved):

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.2.1
```
MartinHjelmare, line 3: What's with the version jump? Are we trying to stay with the server version? I don't think we should do that. We should just follow semantic versioning and remain logical. This add-on is separate from the server.

Issue #3649, opened 2024-06-17
Can't update the dns (gandi): Duplicate keyword name at line 60.


3v1n0:
### Describe the issue you are experiencing

When launching the addon I'm always getting this:

Error parsing credentials configuration '/data/dnsapikey': Duplicate keyword name at line 60.  

### What type of installation are you running?

Home Assistant Supervised

### Which operating system are you running on?

Debian

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.0.18

### Steps to reproduce the issue

1. Launch the addon
2. Don't profit

### System Health information

.

### Anything in the Supervisor logs that might be useful for us?
Basically the `dnsapikey` file contained for me `dns_gandi_api_key` duplicated entry:

```txt
dns_desec_token = null                                  
 dns_digitalocean_token = null                          
 dns_directadmin_url = null                             
 dns_directadmin_username = null                        
 dns_directadmin_password = null                        
 dns_dnsimple_token = null                              
 dns_dnsmadeeasy_api_key = null   
 dns_dnsmadeeasy_secret_key = null
 dns_duckdns_token = null  
 dns_dynu_auth_token = null     
 dns_gandi_api_key = FooBarBaz                        
 dns_gehirn_api_token = null                            
 dns_gehirn_api_secret = null
 dns_hetzner_api_token = null                                                                           
 dns_infomaniak_token = null                            
 dns_linode_key = null                                  
 dns_linode_version = null                              
 dns_luadns_email = null                                                                                         
 dns_luadns_token = null                                                                                         
 dns_namecheap_username = null                                                                                   
 dns_namecheap_api_key = null                                                                                    
 dns_netcup_customer_id = null                                                                                   
 dns_netcup_api_key = null                                                                                       
 dns_netcup_api_password = null                                                                                  
 dns_njalla_token = null                                
 dns_nsone_api_key = null                               
 dns_porkbun_key = null                                 
 dns_porkbun_secret = null                              
 dns_ovh_endpoint = null                                
 dns_ovh_application_key = null                         
 dns_ovh_application_secret = null
 dns_ovh_consumer_key = null      
 dns_rfc2136_server = null 
 dns_rfc2136_port = null                                
 dns_rfc2136_name = null                                
 dns_rfc2136_secret = null                              
 dns_rfc2136_algorithm = null
 aws_access_key_id = null                               
 aws_secret_access_key = null                           
 dns_sakuracloud_api_token = null                       
[...]

Issue #3625, opened 2024-06-03
Let's Encrypt with Gandi DNS challenge and API key not working in version 5.0.18


marcoce7:
### Describe the issue you are experiencing

In version 5.0.18 of the Let's Encrypt add-on, renewing with Gandi DNS challenge and API key does not work any more.
The relevant lines in the add-on logs are:
```
Error parsing credentials configuration '/data/dnsapikey': Duplicate keyword name at line 60.
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.
```
I didn't change the configuration with respect to a previous version that was working correctly.

I noticed that since #3581 the dns_gandi_api_key is added to /data/dnsapikey in
https://github.com/home-assistant/addons/blob/master/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L118-L121
but dns_gandi_api_key is added to /data/dnsapikey also in
https://github.com/home-assistant/addons/blob/master/letsencrypt/rootfs/etc/cont-init.d/file-structure.sh#L49
So the duplicated keyword name error make sense.
However, I do not know how to access the Certbot logs and files in the add-on container while running to check this.

If I use gandi_token instead of gandi_api_key in the add-on YAML configurations, renewing the certificate works.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.0.18

### Steps to reproduce the issue

1. Manually start the add-on
2. Let's Encrypt fail to renew the certificate

## System Information

version | core-2024.5.5
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.12.2
os_name | Linux
os_version | 6.6.28-haos-raspi
arch | aarch64
timezone | Europe/Rome
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
GitHub API Calls Remaining | 5000
Installed Version | 1.34.0
Stage | running
Available Repositories | 1388
Downloaded Repositories | 8

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

[...]

PR #3689, opened 2024-07-15
Update file-structure.sh

marcoce7:
Do not add dns_gandi_api_key to /data/dnsapikey, as this is added by /etc/services.d/lets-encrypt/run too. Adding it twice makes certbot fail with
Error parsing credentials configuration '/data/dnsapikey': Duplicate keyword name at line 60.


marcoce7:
This addresses https://github.com/home-assistant/addons/issues/3625 and https://github.com/home-assistant/addons/issues/3649

agners:
Thanks for the fix! Can you also bump the version in config.yaml and update CHANGELOG.md.

marcoce7:
Done.
Currently on the master branch, CHANGELOG.md reports version 5.0.23 but config.yaml is still on 5.0.22.
So I bumped to 5.0.24, I hope this works.

agners (approved):

agners:
> So I bumped to 5.0.24, I hope this works.

Ah yeah it seems the bump got missed in #3659. It will cause a rebuild still, so this shouldn't be a problem.

PR #3725, opened 2024-08-14
nginx_proxy: Define default for real_ip_from to make it optional

agners:
Unfortunately we can't make the setting truly optional, but since it is a list, we treat an empty list as not set. Set the empty list as default to avoid users requiring to create and delete a dummy real_ip_from value.

Fixes: #3726

## Summary by CodeRabbit

- **New Features**
	- The `real_ip_from` configuration directive is now optional, enhancing setup flexibility for users.
	- Introduced a new configuration option for `real_ip_from`, allowing for future IP address configurations.

- **Documentation**
	- Updated documentation to reflect the new flexible handling of the `real_ip_from` option, which can now accept multiple entries.

- **Version Update**
	- Updated to version 3.10.1, reflecting the latest enhancements and features.

### DIFF
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 3.10.0
+version: 3.10.1
 hassio_api: true
 slug: nginx_proxy
 name: NGINX Home Assistant SSL proxy
@@ -26,6 +26,7 @@ options:
     active: false
     default: nginx_proxy_default*.conf
     servers: nginx_proxy/*.conf
+  real_ip_from: []
 ports:
   443/tcp: 443
   80/tcp: null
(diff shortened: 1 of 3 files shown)


nginx_proxy/config.yaml
```diff
@@ -26,6 +26,7 @@ options:
     active: false
     default: nginx_proxy_default*.conf
     servers: nginx_proxy/*.conf
+  real_ip_from: []
```
ludeeus, line 29: <https://github.com/home-assistant/addons/blob/master/nginx_proxy/DOCS.md#configuration>
The configuration example has no value and should probably be adjusted as well.

ludeeus (approved):

nginx_proxy/config.yaml
agners: Good point, added that!

PR #3734, opened 2024-08-22
Update config.yaml add port 39108

gitaday:
In my understanding exposing this port would enhance the add-on capabilities, for instance:
You could start zigbee2mqtt pointing to the new exposed port on localhost. The core_ssh docker having already a reverse ssh listing on the corresponding internal port will forward data to a remote server with a ZigBee adapter and ser2net installed. As consequence you will have only one installation of home-assistant with a set of local ZigBee devices managed by ZHA and a set of remote ZigBee devices managed with zigbee2mqtt. Cheers 🥂

## Summary by CodeRabbit

- **New Features**
	- Introduced a new TCP port configuration for SSH connections, allowing flexibility with an additional port option (39108/tcp).
	- Added a new metadata configuration for a forked repository, improving organization and access to addons.
- **Bug Fixes**
	- Updated the version number for the SSH configuration to reflect the latest changes (from 9.14.0 to 9.14.2).

### DIFF
--- a/repository.yaml
+++ b/repository.yaml
@@ -1,0 +1,3 @@
+name: gitaday fork of addons
+url: https://github.com/gitaday/addons/
+maintainer: HomeAssistant Team <info@home-assistant.io>
--- a/ssh/config.yaml
+++ b/ssh/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 9.14.2
+version: 9.14.0
 slug: ssh
 name: Terminal & SSH
 description: Allow logging in remotely to Home Assistant using SSH
--- a/ssh/config.yaml
+++ b/ssh/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 9.14.0
+version: 9.14.2
 slug: ssh
 name: Terminal & SSH
 description: Allow logging in remotely to Home Assistant using SSH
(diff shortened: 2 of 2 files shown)


agners:
This PR also contains a change with a new `repository.yaml` which probably wasn't meant to be added here.

In any case, this seems a rather specific use case and I don't think we should add such a change to the main SSH add-on.

gitaday:
> This PR also contains a change with a new `repository.yaml` which probably wasn't meant to be added here.
> 
> In any case, this seems a rather specific use case and I don't think we should add such a change to the main SSH add-on.

gitaday:
Hello agners,
Thank you,
Yes this PR was still in draft and shall not have been committed.
The idea behind this PR was not to enable a specific use case (which was just an example) but to extend the capabilities of the SSH addon. Let me explain:
The add-on SSH as it is now is wonderful, it allows to create a reverse SSH connection to an external server but the usage of this connection is limited to the inner SSH addon. This is because there are no additional ports exposed from the add-on to the host. If you expose and map one (or more) ports from the add-on to the host you would give the possibility to connect any other plugin to one (or more) external server(s).
Please reply to this comment if it is anywhere wrong or incomplete,
BR,
gitaday

PR #3758, opened 2024-09-12
Bump zwave-js to 13.3.0 and zwave-js-server to 1.38.0

MartinHjelmare:


MartinHjelmare:
I've tested building this locally and connecting to the add-on from core-2024.9.1. It works!

frenck (approved):

PR #3782, opened 2024-10-03
Fix otbr for USB cc2652 dongles

tl-sl:
Upstream put the code to deassert DTR/RTS behind a flag, as it was breaking systems that were incorrectly configured using flow control firmware without configuring flow control settings.

This restores the previous behaviour introduced in `2.9.0`

## Summary by CodeRabbit

- **New Features**
	- Updated changelog to reflect version 2.11.1, addressing USB TI CC2652 device issues.
- **Improvements**
	- Enhanced configuration handling for network interfaces and flow control in the OpenThread Border Router script.
	- Improved firewall setup to prevent errors and ensure proper traffic management.
- **Bug Fixes**
	- Resolved issues related to the handling of primary network interfaces and default settings.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.11.1
+
+-  Fix issue with USB TI CC2652 based devices
+
 ## 2.11.0
 
 - Bump to OTBR POSIX version ff7227ea9a2 (2024-09-25 14:54:08 -0700)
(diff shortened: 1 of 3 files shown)


agners (approved):
Hm, I see, that change got added with https://github.com/openthread/openthread/pull/10644, which came in with the latest OTBR POSIX bump shipped with the 2.11.0 add-on version.

I guess the same issue could have happened in our case for users before 2.9.1, if they happened to have turned off the hardware flow control but the device required it. 

Furthermore, there could be hardware which don't use flow control but behave differently when explicitly de-asserting those lines :grimacing:  But to have all three type of sticks working, we'd need another flag for that. 

But since we shipped now since two minor releases with the deassert in place, I agree we should keep consistent.

PR #3794, opened 2024-10-09
Create cs.yaml

LochyCZ:
Czech translate


agners:
We currently do not manage translations for add-ons just yet. We plan to bring them / make them available via Lokalise in the future.

Nevertheless, thank you for being willing to contribute!

PR #3808, opened 2024-10-21
openthread-border-router: Bump firmwares, flasher, and OTBR to latest versions

puddly:
@agners I've also bumped [OTBR itself](https://github.com/openthread/ot-br-posix/compare/ff7227ea9a2dc0e50f92ab93bd5c9a29e82fe793..580cafc150946bc418aa6790a416bd2c951bae24). It looks like `dhcpcd` is now installed and prefix delegation is in use, probably for Thread 1.4. Is this something we want to disable?

## Release Notes for Version 2.12.0

- **New Features**
	- Enhanced OTBR discovery process with additional configuration parameters for improved device and firmware information.
	- Support for auto firmware updates for specific devices.

- **Updates**
	- Universal SiLabs flasher updated to version 0.0.25.
	- OTBR firmware versions updated to the latest releases.
	- Configuration files updated to reflect the new version 2.12.0.
	- Radio firmware version added to discovery information.

- **Bug Fixes**
	- Resolved issues with log output buffering and ipset errors when the firewall is disabled.


puddly:
I've set `DHCPV6_PD_REF=0` in this PR for now to at least get it building. Untested.

agners:
> It looks like `dhcpcd` is now installed and prefix delegation is in use, probably for Thread 1.4. Is this something we want to disable?

What do yo mean by "`dhcpcd` is now installed"? With this change? We don't really use the regular installation script with the add-on, so I don't think that gets installed just like that no?

Afaik, `dhcpcd` was an option since a while.

Hm, Apple BRs use PD addresses already since a while, even with Thread 1.3. Not sure if this is directly related to 1.4.

puddly:
> so I don't think that gets installed just like that no?

It's installed by `bootstrap`: https://github.com/openthread/ot-br-posix/blob/580cafc150946bc418aa6790a416bd2c951bae24/script/bootstrap#L110-L121

agners:
> It's installed by `bootstrap`, which we run: https://github.com/openthread/ot-br-posix/blob/580cafc150946bc418aa6790a416bd2c951bae24/script/bootstrap#L110-L121

Hm, I see that changed just recently with https://github.com/openthread/ot-br-posix/commit/3b145fa40ffec3eb49b535093d6388d5c7331d1b. We could opt-out of PD at this point, which might make sense.

I think there are two things we need to decide here:
a) Do we want to update to Thread 1.4 (e.g. set `OT_THREAD_VERSION=1.4`)
b) Do we want to enable IPv6 PD by default

I tend to say no to b). For a) I am a bit less certain. I'll check what setting Thread version to 1.4 exactly entails.

openthread_border_router/Dockerfile
```diff
@@ -88,7 +89,7 @@ RUN \
         -DOT_COAP=OFF \
         -DOT_COAPS=OFF \
         -DOT_DNS_CLIENT_OVER_TCP=OFF \
-        -DOT_THREAD_VERSION=1.3 \
+        -DOT_THREAD_VERSION=1.4 \
```
agners, line 92: From what I can tell, this mainly enables the ePSKc which is an ephemeral variant of the PSKc, which is used for the Thread specific commissioning. This will be used for credentials sharing, but we don't implement that yet.

Other features behind the flag are Thread domain name, which from what I understand also only becomes relevant if there are other Thread 1.4 BRs.

Setting this to 1.4 will advertise us as an Thread 1.4 BR. I don't think is problematic currently, but maybe it becomes once other systems rely on e.g. credential sharing to work. For now I prefer to stay with 1.3.

```suggestion
        -DOT_THREAD_VERSION=1.3 \
```

agners (changes requested):

puddly:
I've reverted the advertised Thread version back to 1.3 and included the [OTBR coprocessor version API](https://github.com/openthread/ot-br-posix/pull/2602) patch.

openthread_border_router/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.12.0
+- Bump universal-silabs-flasher to 0.0.23
+- Bump OTBR firmwares to latest versions
+- Bump OTBR POSIX version 580cafc150 (2024-10-21 10:47:10 +0000)
```
agners, line 6:
```suggestion
- Bump OTBR POSIX version 580cafc150 (2024-10-21 10:47:10 +0000)
- Add radio firmware version to discovery information
```

agners (approved):
LGTM, thanks!

Issue #3801, opened 2024-10-14
Let's Encrypt: unexpected EOF while looking for matching `''


Ezor:
### Describe the issue you are experiencing

Today, I wanted to renew my HA certificate and I get this error during Let's Encrypt start up :

```bash
s6-rc: info: service legacy-services: starting
services-up: info: copying legacy longrun lets-encrypt (no readiness notification)
s6-rc: info: service legacy-services successfully started
[19:45:16] INFO: Selected http verification
./run: line 317: unexpected EOF while looking for matching `''
s6-rc: info: service legacy-services: stopping
s6-rc: info: service legacy-services successfully stopped
s6-rc: info: service legacy-cont-init: stopping
s6-rc: info: service legacy-cont-init successfully stopped
s6-rc: info: service fix-attrs: stopping
s6-rc: info: service fix-attrs successfully stopped
s6-rc: info: service s6rc-oneshot-runner: stopping
s6-rc: info: service s6rc-oneshot-runner successfully stopped
```

Here is my Let's Encrypt configuration : 
```
domains:
  - mydomain.freeboxos.fr
email: my.email.address@me.com
keyfile: privkey.pem
certfile: fullchain.pem
challenge: http
dns: {}
```

I use the port 8881 for HTTP challenge.

I did not changed anything to the configuration and/or port forwarding since the last update of my certificat 3 month ago.

Do you have any idea of what is happening here ?

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.2.2

### Steps to reproduce the issue

1. Start Let's Encrypt 
2. Obtain the error : **./run: line 317: unexpected EOF while looking for matching `''**

### System Health information

Home Assistant Community Store

GÉRER
GitHub API	ok
GitHub Content	ok
GitHub Web	ok
GitHub API Calls Remaining	4957
Installed Version	1.34.0
Stage	running
Available Repositories	1433
Downloaded Repositories	5
Home Assistant Cloud

GÉRER
Connecté	false
Accéder au serveur de certificats	ok
Accéder au serveur d'authentification	ok
Accéder à Home Assistant Cloud	ok
Home Assistant Supervisor

Système d'exploitation hôte	Home Assistant OS 13.1
Canal de mise à jour	stable
Version du Supervisor	supervisor-2024.10.2
Version de l'agent	1.6.0
[...]

PR #3824, opened 2024-11-13
Lint bash scripts with custom shebangs

mislav:
Many scripts in this repository have this shebang:

However, this doesn't match the shebang regex in `action-shellcheck` and therefore those files wouldn't get included in the linting check.

Supersedes https://github.com/home-assistant/addons/pull/3803 /cc @agners 

This is the only mechanism to include additional scripts in the linting process. I'm not a fan of this approach, since new scripts with unique names added to this repository again won't be linted by default, but it's a start to ensure regressions like https://github.com/home-assistant/addons/issues/3801 are less likely to happen.

## Release Notes

- **New Features**
	- Enhanced DNS challenge handling for SSL certificate requests with additional DNS providers.
	- Improved firmware flashing capabilities for various USB devices, including support for new manufacturers.
	- Added logging for service exit codes and enhanced error handling in multiple service scripts.

- **Bug Fixes**
	- Corrected variable usage in exit code calculations across various service scripts to improve clarity and functionality.

- **Documentation**
	- Added comments and shellcheck directives to improve code quality and maintainability in several scripts.

- **Chores**
	- Streamlined various scripts to enhance robustness and error handling mechanisms.


mislav:
Update: pushed a commit that fixes/allowlists all the shellcheck violations in scripts that are now newly included in the linting process.

mislav:
Note that I'm ignoring most of Coderabbit suggestions because it doesn't understand that I'm merely trying to make CI pass with minimum impact to the codebase. Sure, its suggestions are legitimate, but even such an innocuous change as adding double quotes around an argument passed to a command can have negative consequences if the previous code relied (intentionally or accidentally) on word-splitting, for example. Because I'd rather avoid making code changes across dozens of scripts in this repository, I've opted to pepper `# shellcheck disable` directives throughout.

.github/workflows/lint.yml
```diff
@@ -50,3 +50,19 @@ jobs:
...
+            discovery
+            finish
+            lock-tables-for-backup
+            matter-server-discovery
+            reboot
+            run
+            shutdown
+            universal-silabs-flasher-up
+            unlock-tables-for-backup
+            vlc-discovery
```
agners, line 68: It doesn't matter much, but for clarity, should we maybe use the full path for the add-on specific scripts, e.g. `vlc/rootfs/etc/s6-overlay/scripts/vlc-discovery`? :thinking: 

I'd still use just file name for the generic s6-init scripts (like `run` and friends).

agners:
@coderabbitai resolve

This PR doesn't try to fix shellcheck issues, it tries to enable along with minimal fixes. 

@coderabbitai pause

.github/workflows/lint.yml
mislav: I would have loved to use full paths, but the `additional_files` input to the action only supports file names and not paths: https://github.com/ludeeus/action-shellcheck/blob/00b27aa7cb85167568cb48a3838b75f4265f2bca/action.yaml#L141-L143

mislav:
@agners I appreciate your assistance so far. Let me know if there is anything else to change/add to this PR. I believe this should be good to go, but there was one CI failure that looks intermittent; i.e. I don't think it's likely that it was caused by the changes in this branch.

.github/workflows/lint.yml
agners, line 68: It seems that the Google Assistant SDK add-on doesn't build anymore. Let's exclude it from linting so we get a complete CI run.

```suggestion
            vlc-discovery
          ignore_paths: >-
            google_assistant
```

google_assistant/rootfs/etc/services.d/google-assistant/run
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/with-contenv bashio
+# shellcheck shell=bash
```
agners, line 2: Let's not touch google_assistant to avoid CI run. We'll have to deprecate this add-on.

```suggestion
```

mariadb/rootfs/etc/s6-overlay/s6-rc.d/mariadb-pre/run
```diff
@@ -11,8 +11,8 @@ if ! bashio::fs.directory_exists "${MARIADB_DATA}"; then
 fi
 
 # Save variables
-printf "${MARIADB_DATA}" > /var/run/s6/container_environment/MARIADB_DATA
-printf "${NEW_INSTALL}" > /var/run/s6/container_environment/NEW_INSTALL
+cat <<<"${MARIADB_DATA}" > /var/run/s6/container_environment/MARIADB_DATA
+cat <<<"${NEW_INSTALL}" > /var/run/s6/container_environment/NEW_INSTALL
```
agners, line 15: There is a very slight difference between the two options, printf doesn't add a newline, whereas cat does. I'd go for `echo -n` to make sure this is equivalent.

```suggestion
echo -n "${MARIADB_DATA}" > /var/run/s6/container_environment/MARIADB_DATA
echo -n "${NEW_INSTALL}" > /var/run/s6/container_environment/NEW_INSTALL
```

agners:
Hm, your last commit adding `# vim: ft=bash` actually also uncovered some missing `# shellcheck shell=bash`. Is there a particular reason for that?

 I'd say, if we touch all files to add the `# vim: ft=bash`, we should also add `# shellcheck shell=bash` everywhere (except the google_assistant add-on, since that doesn't seem to build).

agners:
Thanks for the detailed explanation 🤩

> Do you still want me to add shellcheck override directives to `#!/usr/bin/env bashio` scripts even though they're technically not needed?

No, I did not notice that difference. I agree with you, we should not override an additional time.

agners (approved):
LGTM!

Since all changes do not change any runtime behavior (or at least do not change functionality for the mariadb change) we can merge this without bumping add-on versions. This will mean that the new files will only get deployed with the next release of the individual add-ons. Since the changes are transparent for runtime and developer oriented, it won't be necessary to add entries to CHANGELOG.md.

PR #3857, opened 2024-12-09
Move Turbo closer to Large to avoid confusing it with Tiny

googanhiem:
Move Turbo closer to Large to avoid confusing it with Tiny.

## Summary by CodeRabbit

- **Documentation**
	- Updated the Whisper add-on documentation for clarity on the `turbo` model's performance compared to `large-v3`.
  
- **Configuration**
	- Reordered the list of acceptable model values in the configuration file, moving `turbo` to the end of the list.

### DIFF
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.3.1
+
+- Move `turbo` down the list closer to `large` to avoid confusion
+
 ## 2.3.0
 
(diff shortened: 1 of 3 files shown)


frenck (changes requested):
Thanks, @googanhiem 👍 

Can you also update the changelog and version? Otherwise the update will not ship.

../Frenck

googanhiem:
@frenck updated the version number 👍

frenck (approved):
Thanks, @googanhiem 👍 

../Frenck

PR #3877, opened 2024-12-30
samba: Update smb.gtpl to enable files app on iOS / iPads 18+

liambresnahan:
This PR is to enable RW access to config directories via files.app which requires new global attributes per https://www.samba.org/samba/docs/4.9/man-html/vfs_streams_xattr.8.html

## Summary by CodeRabbit

- **Configuration Update**
	- Enhanced Samba configuration by adding support for multiple VFS modules to improve extended file attributes management.
	- Updated version number to 12.3.3.
	- Added a new feature for enabling Apple-related Samba configuration changes.


samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -26,6 +26,8 @@
    dos charset = CP850
    unix charset = UTF-8
 
+   vfs objects = streams_xattr
```
agners, line 29: Reading a bit about Samba and macOS interoperability the two modules `fruit` and `catia` come often up. There is an example config in the fruit man page at https://www.samba.org/samba/docs/current/man-html/vfs_fruit.8.html.

I'd say that the default configs for the fruit module are probably fine for our purpose, but maybe it makes sense to enable those additional modules too? :thinking: 

```suggestion
   vfs objects = catia fruit streams_xattr
```

agners (changes requested):
Thanks, the change generally LGTM.

Any change to an add-on which should get rolled out also requires a bump of the add-on version in `config.yaml` and an entry in CHANGELOG.md.

samba/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Changelog
+## 12.3.3
+
+- Enable Apple related Samba config changes
```
agners, line 4:
```suggestion
- Enable Samba configurations to improve interoperability with Apple devices
```

agners (changes requested):
Other than the nit, LGTM, thanks!

agners (approved):
LGTM, thanks!

PR #3894, opened 2025-01-16
[SVH] Add WINS Server and MasterBrowser options to SAMBA AddOn

stefvanhooijdonk:
Add configuration options to the SAMBA AddOn to enable:
- The NMBD to enable WINS Server support
- The NMBD to become the WORKGROUP browser master
- Add the Configuration Page options and descriptions
- Add the update to the DOCS and ChangeLog

## Release Notes for Samba Add-on v12.4.1

- **New Features**
  - Added option to enable Samba as a local browser master.
  - Added option to enable Samba WINS Server.

- **Configuration Updates**
  - Introduced new configuration options: `winsserver_mode` (default: false) and `masterbrowser_mode` (default: true).
  - Enhanced documentation to guide users on new features.

These updates provide users with increased control over Samba's network interaction capabilities while preserving existing functionality.


stefvanhooijdonk:
@coderabbitai review

samba/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 # Changelog
+## 12.4.1
+
+- Add the option to enable SAMBA to become local browsermaster
+- Add the option to enable SAMBA WINS Server
```
agners, line 5: From https://learn.microsoft.com/en-us/windows-server/networking/technologies/wins/wins-top:

> Windows Internet Name Service (WINS) is a legacy computer name registration and resolution service that maps computer NetBIOS names to IP addresses.
> 
> If you do not already have WINS deployed on your network, do not deploy WINS - instead, deploy Domain Name System (DNS). [...]

Since even Microsoft considers WINS legacy, I don't think we should add support for it on our end.

agners (commented):
Thank you for your PR!

Can you elaborate a bit more what your use case is/what you try to solve adding these options?

samba/CHANGELOG.md
stefvanhooijdonk: MasterBrowser in my network helps with navigating my servers/shares.

wins is old, can live without.

stefvanhooijdonk: Thanks at least for considering

agners, line 5: As discussed in the other review, since WINS is really old and deprecated, let's remove the WINS option from this PR.

samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -3,6 +3,17 @@
...
 
+   {{ if .winsserver_mode }}
+   wins support = yes
+   {{ end }}
+
+   {{ if .masterbrowser_mode }}
+   domain master = yes
+   preferred master = yes
+   local master = yes
+   os level = 255
```
agners, line 14: From what I understand that is really the highest OS level there is. I get you want Home Assistant to be the  Maybe we should pick something slightly lower, e.g. 200? This would allow an other/external Samba instance to pick something higher if they want to be master browser.

```suggestion
   os level = 200
```

agners (changes requested):

PR #3923, opened 2025-02-11
feat: add a Dnsmasq upstream resolver filtering option

alryaz:
This allows for filtering of upstream responses (A and AAAA separately). Quite useful when the ISP does not support IPv6, and the router used does not support AAAA filtering.

Filtering of A responses is added solely for completeness.

P.S. It must be mentioned that the option should be named `filter_AAAA` solely for correctness purposes, not `filter_aaaa`; however YAML lowercase keying imposes styling constraints.

## Summary by CodeRabbit

- **New Features**
  - Introduced new configuration options to control DNS filtering. Administrators can now choose to disable responses for IPv4 (A) and IPv6 (AAAA) records from upstream resolvers, with filtering turned off by default for legacy behavior.
  - Enhanced configuration capabilities with new options `filter_a` and `filter_aaaa` in the settings.
  - Added conditional configurations `filter-A` and `filter-AAAA` to enhance DNS record filtering.


agners:
Why would you filter AAAA responses?

A properly working application should not pick the AAAA address if your host does not have a globally routable IPv6 address assigned. If you have problems with AAAA responses, that likely means your IPv6 setup is actually broken (e.g. you have a global IPv6 address assigned but it is not routed).

alryaz:
> Why would you filter AAAA responses?
> 
> A properly working application should not pick the AAAA address if your host does not have a globally routable IPv6 address assigned. If you have problems with AAAA responses, that likely means your IPv6 setup is actually broken (e.g. you have a global IPv6 address assigned but it is not routed).

There may be lack of IPv6 on WAN egress, and IPv6 may still be used within LAN (for ARP-less communication). While not globally routable, a host may assume IPv6 WAN routing is available due to presence of a default nexthop, and thus attempt to utilize the address.

There are also rare cases of ISPs blocking border-crossing traffic (which is weird, however I've come across four instances of such over the past decade).

There's a case to be made for end users, who would probably be keen on pointing their router's DNS towards HA's Dnsmasq instance and solve the IPv6 issue that way (especially when disabling IPv6 isn't available on some ISP hardware).

agners:
> There may be lack of IPv6 on WAN egress, and IPv6 may still be used within LAN (for ARP-less communication). While not globally routable, a host may assume IPv6 WAN routing is available due to presence of a default nexthop, and thus attempt to utilize the address.

Sure, IPv6 in link-local addresses and ULA addresses can be assigned. But any correctly implemented IP client program should try all addresses returned by the DNS (just like this [tcp_connect.c](https://github.com/ryuever/unix-network-programming-v3/blob/master/lib/tcp_connect.c#L19-L28) example). In fact, it seems that by default, `getaddrinfo` does order the IP address according to what is available locally. In a quick test, `getaddrinfo` on a system which only has link-local addresses and ULA actually returned the IPv4 address first already, so even a "wrongly" implemented client should do "the right thing".

If IPv6 address filtering is required on DNS level, then this hints to a seriously broken IPv6 address configuration (e.g. by manually configure global addresses which in the end are not routed and/or a router which hands out IPv6 and then blocks traffic still). 


> There's a case to be made for end users, who would probably be keen on pointing their router's DNS towards HA's Dnsmasq instance and solve the IPv6 issue that way (especially when disabling IPv6 isn't available on some ISP hardware).

If ISPs provide such fucked up IPv6 implementation, they should complain with the ISP or switch to a sane ISP :shrug: 

> **Edit:** IPv4-only VPNs may also cause problems. If a globally routed IPv6 is available, it could potentially bypass routes set up by an interface, thus causing hosts to potentially leak traffic.

But then fixing the problem on DNS level seems wrong. If something resolves an address somehow differently traffic still might get the wrong way. This seems like it needs a firewall solution.

I am fine merging this, but can we add a hint that this typically should not be required ans hints to a IPv6 misconfiguration of some sort?

frenck:
I think we should not add this; this is solving things in the wrong place.

Additionally, adding A filtering support "solely for completeness", is not a good reason. We should not just add more options for the sake of things.

IMHO, we should not merge this pull request.

../Frenck

PR #3931, opened 2025-02-12
⬆️ upgrade deconz version to 2.29.1-beta as experimental

pokulo:
I would be interested to get the most recent bug fixes of deconz by switching to this experimental 2.29.1-beta release. I am just curious, whether the deconz add-on can be upgraded to a beta release, while leaving auto-updating installs unaffected. My attempt do do that is to set the `stage` back from default `stable` to `experimental`.

If `stage: experimental ` also affects prior versions of the deconz add-on, than this likely means there is no support for auto-updates to stable releases only. In this case, I guess it is just to early to upgrade all users to a beta version and I am totally ok with simply closing this PR.

## Summary by CodeRabbit

- **New Features**
	- Updated to a new beta release (7.1.0-beta) with an underlying update to 2.29.1-beta.
	- Enabled flexible release channels for installation, allowing access to beta releases.
	- Reflected an experimental stage in configuration settings for improved feature rollout.

### DIFF
--- a/deconz/CHANGELOG.md
+++ b/deconz/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.1.0-beta
+
+- Bump deCONZ to 2.29.1-beta
+
 ## 7.0.0
 
 - Bump deCONZ to 2.28.1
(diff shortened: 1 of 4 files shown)


frenck:
Hi there @pokulo 👋 

Thanks for opening a pull request. Unfortunately, we cannot accept unstable versions. Once it has been released as stable, feel free to re-open the pull request and I'm happy to take a look.

Thanks! 👍 

../Frenck

PR #3949, opened 2025-02-21
Add anonymous option back to Mosquitto Addon

0lini:
I am submitting this pull request to inquire about the possibility of reintroducing the anonymous support functionality that was previously removed. While I understand the reasons behind its removal—likely due to security and privacy concerns—there are specific use cases where users could still benefit from this feature.
For instance, in debugging scenarios or when certain devices do not support the use of credentials, anonymous support could offer a valuable way to interact with Home Assistant without requiring user authentication. However, I recognize the importance of maintaining Home Assistant's security standards.
With that in mind, I would like to explore whether there is a way to safely reintroduce this functionality, perhaps with enhanced safeguards. This pull request includes a basic implementation of the functionality and is intended as a starting point for further discussion and refinement. I am open to feedback and suggestions on how to implement this securely and in a way that aligns with Home Assistant's best practices.


agners:
Hi @0lini, thanks for your PR!

This PR as is removes all add-ons except mosquitto, and uses the container images from a private repository. It can't get merged like this.

You have to base your changes on-top of the master branch, and a PR which only makes changes to the mosquitto add-on.

PR #3955, opened 2025-02-25
dnsmasq: Enable host networking

sleiner:
This patch enables host networking mode for the dnsmasq addon.

## Why is this helpful?

As per the addon's description, it allows users to assign custom domain names to IP addresses in their network, e.g. `home-assistant.example.com`.
When starting to use this feature, I have hit a limitation that felt quite frustrating:

I have various other hosts in my network that I would like to assign names, such as `nas.example.com`.
The dnsmasq addon resolves these without issue, from anywhere _outside_ of Home Assistant OS. From inside HAOS, and notably from Home Assistant itself, I cannot use that domain name (e.g. for adding the NAS's SMB share as a backup location).

If my Home Assistant OS has the IP addresses 192.0.2.23 inside my LAN, I can access the DNS resolver at that IP from my laptop just fine, but not so inside HAOS. If I try, I get outputs like this:

    > dig nas.example.com

    ; <<>> DiG 9.18.33 <<>> nas.example.com
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 37159
    ;; flags: qr rd ra ad; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1

    ;; OPT PSEUDOSECTION:
    ; EDNS: version: 0, flags:; udp: 1232
    ; COOKIE: 37b052b040bae689 (echoed)
    ;; QUESTION SECTION:
    ;nas.example.com.		IN	A

    ;; AUTHORITY SECTION:
    example.com.		600	IN	SOA	ns.icann.org. noc.dns.icann.org. 2025011553 7200 3600 1209600 3600

    ;; Query time: 135 msec
    ;; SERVER: 172.30.32.3#53(172.30.32.3) (UDP)
    ;; WHEN: Tue Feb 25 23:15:48 CET 2025
    ;; MSG SIZE  rcvd: 132

Just running dig bypasses the names defined in dnsmasq. On the other hand, if I try to access the server directly via the IP that other devices on my home network use for DNS, I get "connection refused":

    > dig @192.0.2.23 nas.example.com
    ;; communications error to 192.0.2.23#53: connection refused
    ;; communications error to 192.0.2.23#53: connection refused
    ;; communications error to 192.0.2.23#53: connection refused

    ; <<>> DiG 9.18.33 <<>> @192.0.2.23 nas.example.com
    ; (1 server found)
    ;; global options: +cmd
    ;; no servers could be reached

I have found multiple Home Assistant community threads which likely refer to the same issue:

-   https://community.home-assistant.io/t/can-home-assistant-os-use-itself-for-dns/552718
-   https://community.home-assistant.io/t/dnsmasq-configuraton/396360

Activating `host_network` fixes these issues:

    > dig nas.example.com

    ; <<>> DiG 9.18.33 <<>> nas.example.com
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24463
    ;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

    ;; OPT PSEUDOSECTION:
    ; EDNS: version: 0, flags:; udp: 1232
    ; COOKIE: 9df3c10872739d34 (echoed)
    ;; QUESTION SECTION:
    ;nas.example.com.		IN	A

    ;; ANSWER SECTION:
    nas.example.com.	5	IN	A	192.0.2.42

    ;; Query time: 3 msec
    ;; SERVER: 172.30.32.3#53(172.30.32.3) (UDP)
    ;; WHEN: Tue Feb 25 23:39:27 CET 2025
    ;; MSG SIZE  rcvd: 95

## Any other references?

Before testing this change to the `dnsmasq` addon, I tried a similar setup with the [AdGuard Home Addon], which does not have this problem. It also has `host_network` enabled.

[AdGuard Home Addon]: https://github.com/hassio-addons/addon-adguard-home

## Summary by CodeRabbit

- **New Features**
  - Introduced an option that allows the DNS service to leverage the host’s network stack, which may influence network interface interactions and routing.

### DIFF
--- a/dnsmasq/config.yaml
+++ b/dnsmasq/config.yaml
@@ -13,6 +13,7 @@ arch:
   - i386
 image: homeassistant/{arch}-addon-dnsmasq
 init: false
+host_network: true
 options:
   defaults:
     - 8.8.8.8


agners:
So technically, we expose the relevant ports using Docker functionality. I'd expect that it doesn't matter if we connect from network or locally, the requests should end up with dnsmasq.

I've tried to reproduce the problem you are reporting on a native generic-x86-64 HAOS 15.0.dev20250218 installation. And for me it seems to work:

```
homeassistant:/config# dig @192.168.1.9 home.mydomain.io

; <<>> DiG 9.18.33 <<>> @192.168.1.9 home.mydomain.io
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 62986
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;home.mydomain.io.              IN      A

;; ANSWER SECTION:
home.mydomain.io.       0       IN      A       192.168.1.10

;; Query time: 0 msec
;; SERVER: 192.168.1.9#53(192.168.1.9) (UDP)
;; WHEN: Wed Feb 26 15:23:52 CET 2025
;; MSG SIZE  rcvd: 61

```

What system are you running Home Assistant on? :thinking:

sleiner:
Hi @agners,

thanks for the quick reply! 😊 I am running HAOS 14.2 on a HA Yellow with a CM4 inside. I did run `dig` through the [Terminal & SSH](https://github.com/home-assistant/addons/tree/master/ssh) addon, which means I am also running it from a container (I think?), not on the host directly.

agners:
> thanks for the quick reply! 😊 I am running HAOS 14.2 on a HA Yellow with a CM4 inside. I did run `dig` through the [Terminal & SSH](https://github.com/home-assistant/addons/tree/master/ssh) addon, which means I am also running it from a container (I think?), not on the host directly.

I see, yeah that doesn't work for me either.

It is possible to access the DNS server from the Home Assistant Core container though, since that one is running in host network too.

However, it isn't possible to access from the DNS plug-in, hence it can't work if you set the DNS server using `ha dns options --servers dns://<public-ip>` :disappointed_relieved: 

However, what does work is using the hassio bridge's host IP, which is static `172.30.32.1`. 

Simply change dnsmasq add-on to use host network does seem a bit harsh to me, it kinda removes all network isolation we have thanks to Docker.

agners:
That said, I am actually not sure if what you try to do is a good idea in first place: If the DNS plug-in (CoreDNS) uses the add-on as primary DNS, then the system can't resolve names until the add-on has been started. But the system itself does use DNS to check things (like if there are updates).

So this puts the system in a chicken-egg situation, especially during startup.

sleiner:
> However, it isn't possible to access from the DNS plug-in, hence it can't work if you set the DNS server using `ha dns options --servers dns://<public-ip>` 😥
>
> However, what does work is using the hassio bridge's host IP, which is static `172.30.32.1`.

Interesting! I can confirm that it works in my setup 👍🏻 It's still very unexpected: I set my router to announce the HAOS instance as DNS server via DHCP, which works everywhere but for the HAOS machine 😅

> I am actually not sure if what you try to do is a good idea in first place: If the DNS plug-in (CoreDNS) uses the add-on as primary DNS, then the system can't resolve names until the add-on has been started. But the system itself does use DNS to check things (like if there are updates).

I tried restarting HAOS a couple times (both with the Dnsmasq addon as it is and the DNS plugin's upstream server set to the bridge IP and with Dnsmasq in host networking mode): everything came up mostly fine. I assume that the DNS plugin's fallback mechanism is helping here.

HA Core came up without issues, integrations using custom dnsmasq domain names (such as [custom_homematic](https://github.com/sukramj/custom_homematic)) also worked fine, presumably due to retry logic. The only thing that showed up where I had to manually retrigger a retry was the supervisor's  SMB mount for the NAS.

> Simply change dnsmasq add-on to use host network does seem a bit harsh to me, it kinda removes all network isolation we have thanks to Docker.

If I understand the Docker docs correctly here, keeping host networking disabled lets Docker act like a firewall layer, right?
I initially assumed that the isolation is not necessary here, since the AdGuard Home addon does the same thing. On the other hand, another layer of security might be worth users requiring a workaround in this (probably very niche) use case.

PR #3956, opened 2025-02-25
Add anonymous option back to Mosquitto Addon

0lini:
This PR restores the anonymous option to the Mosquitto Addon. 
Reference: https://github.com/home-assistant/addons/pull/3949

### DIFF
--- a/mosquitto/config.yaml
+++ b/mosquitto/config.yaml
@@ -36,6 +36,7 @@ schema:
     - username: str
       password: password
       password_pre_hashed: "bool?"
+  anonymous: bool?
   require_certificate: bool
   certfile: str
   cafile: str?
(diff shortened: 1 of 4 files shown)


mosquitto/config.yaml
```diff
@@ -36,6 +36,7 @@ schema:
     - username: str
       password: password
       password_pre_hashed: "bool?"
+  anonymous: bool?
```
frenck, line 39: I really don't think we should ever allow for this.

I think we should consider saying: If you need this, the solution this add-on provides is not for you.

../Frenck

0lini: Hi Frenck,
I would love to hear the reason for this.
I would be open to add acl support, to restrict access for the anonymous user.
But I think a lot of users would appreciate being able to use anonymous for things like hardcoded IoT.

0lini

frenck: > I would love to hear the reason for this.

Because we should not add options that can get users into security and privacy issues. This is one of those things we removed because really... it shouldn't be used.

../Frenck

PR #3855, opened 2024-12-09
Added transip_global_key to config.yaml

Goopher:
## Summary by CodeRabbit

- **Documentation**
	- Updated installation and usage instructions for the Letsencrypt add-on.
	- Expanded details on obtaining certificates and supported DNS providers.
	- Added examples for various DNS challenges, including Cloudflare, GoDaddy, and Google Cloud DNS.
	- Emphasized security best practices, including the use of API tokens.

- **New Features**
	- Introduced a new configuration field, `transip_global_key`, for the TransIP DNS provider in the configuration options.
	- Updated to version 5.2.10 in the configuration file.

PR #3958, opened 2025-02-26
letsencrypt: fix TransIP global_key handling

GoetzGoerisch:
## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with the handling of the DNS configuration parameter for TransIP, ensuring improved reliability.

- **Chores**
  - Updated the release version from 5.4.0 to 5.4.1 and revised the associated release notes.
  - Enhanced the initial setup process by conditionally managing the output of the DNS global key configuration.
  - Modified logging messages for clarity regarding DNS propagation time.

### DIFF
--- a/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
+++ b/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
@@ -275,11 +275,8 @@ elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-simply" ]; then
 elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-transip" ]; then
     bashio::config.require 'dns.transip_username'
     bashio::config.require 'dns.transip_api_key'
-    if bashio::config.has_value 'dns.transip_global_key'; then
-    "dns_transip_global_key=$(bashio::config 'dns.transip_global_key')"
-    fi
     if (( PROPAGATION_SECONDS < 240 )); then
-        bashio::log.info "Increasing DNS propagation limit for TransIP to at least 240 seconds."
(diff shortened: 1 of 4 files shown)


letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -276,7 +276,7 @@ elif [ "${CHALLENGE}" == "dns" ] && [ "${DNS_PROVIDER}" == "dns-transip" ]; then
     bashio::config.require 'dns.transip_username'
     bashio::config.require 'dns.transip_api_key'
     if bashio::config.has_value 'dns.transip_global_key'; then
-    "dns_transip_global_key=$(bashio::config 'dns.transip_global_key')"
+    echo "dns_transip_global_key = $(bashio::config 'dns.transip_global_key')" >> "/data/dnsapikey"
```
agners, line 279: I think indentation is not correct here.

In any case, those optional settings we usually write in `file-structure.sh`. I think this should be the case with this value too.

GoetzGoerisch: Indentation is wrong,  thanks for the review.

If the global_key is present, it is written in file-structure.sh, but as the comments of the issue 3876, this is not working.
I was also following other cases like gandi, which is writing also in the run script.

Please advise.

agners: The problem is, the config `transip_global_key` got introduced later in #3855, and is an optional config As in, transip can be used also without `transip_global_key`. However, users which had that provider setup do not have that config set.

But `file-structure.sh` assumes that the config is always present.

So what needs to happen here is that we check if `transip_global_key` exists, and only if it does it should be written out to the config file.

PR #3862 was on the right track, but somehow the author disappeared.

Btw, thanks a lot for picking this up!

agners (approved):
LGTM, thanks!

PR #3970, opened 2025-03-16
Samba: Make the vfs objects entry in smb.conf optional

m-paulus:
The 'vfs objects' entry in smb.conf introduced in version 12.3.3 causes problems with file systems that do not support xattr such as exFAT. This change makes the inclusion of this entry optional by exposing it as a configuration option.

## Summary by CodeRabbit

- **New Features**
  - Released version 12.5.1 with a new Apple compatibility mode that allows users to toggle enhanced settings for improved interoperability with Apple devices. This option, enabled by default, helps address potential issues on file systems that do not support extended attributes.

- **Documentation**
  - Updated configuration guides and user documentation to explain the new Apple compatibility option, its default activation, and considerations for systems with limited extended attribute support.

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## 12.5.1
 
-- Add the ability to enable or disable the configurations that improve interoperability with Apple devices.
-  Disabling this setting can fix issues with file systems that do not support xattr such as exFAT.
+- Add configurations option to disable Apple devices interoperability. Disabling this setting might be required for file systems that do not support extended attributes such as exFAT.
 
 ## 12.5.0
 
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## 12.5.1
 
-Add the ability to enable or disable the configurations that improve interoperability with Apple devices.
-Disabling this setting can fix issues with file systems that do not support xattr such as exFAT.
+- Add the ability to enable or disable the configurations that improve interoperability with Apple devices.
+  Disabling this setting can fix issues with file systems that do not support xattr such as exFAT.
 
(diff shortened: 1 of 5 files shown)


agners:
This add-on exposes only folders which are on the data storage normally formatted as ext4, how do you end up using exFAT? It does seem to be a rather specific use case to me.

samba/translations/en.yaml
```diff
@@ -27,6 +27,11 @@ configuration:
     name: Enable Compatibility Mode
     description: >-
       Enable this to use old legacy Samba protocols on the Samba add-on.
+  apple_compatibility_mode:
+    name: Enable Compatibility Settings for Apple Devices
+    description: >-
+      Enable Samba configurations to improve interoperability with Apple devices.
+      May cause issues with file systems that do not support xattr such as exFAT.
```
agners, line 34: This needs line breaks to stay within the 80 character line length limit of YAMLLint, so:

```suggestion
      Enable Samba configurations to improve interoperability with Apple
      devices. May cause issues with file systems that do not support xattr
      such as exFAT.
```

samba/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 12.5.1
+
+- Add the ability to enable or disable the configurations that improve interoperability with Apple devices.
+  Disabling this setting can fix issues with file systems that do not support xattr such as exFAT.
```
agners, line 6:
```suggestion
- Add configurations option to disable Apple devices interoperability. Disabling this setting might be required for file systems that do not support extended attributes such as exFAT.
```

agners (changes requested):
LGTM, small nit on wording of the CHANGELOG.md.

agners (approved):
LGTM, thanks!

Issue #3781, opened 2024-10-02
Let's Encrypt Staging workflow is problematic


giggio:
### Describe the issue you are experiencing

If you test on Let's encrypt staging, after you remove the server the addon will not issue a new certificate, even if you try to remove the certificates fom `/ssl`.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Let's Encrypt

### What is the version of the add-on?

5.0.18

### Steps to reproduce the issue

1. Install the addon
2. Test the configuration using `acme_server: https://acme-staging-v02.api.letsencrypt.org/`
3. Check certs were issued on `/ssl`.
4. Remove the `acme_server` config and certs from `/ssl`.
5. Start the addon again, it will recreate the staging certificates.

### System Health information

N/A.

### Anything in the Supervisor logs that might be useful for us?

```txt
N/A.
```

### Anything in the add-on logs that might be useful for us?

```txt
cont-init: info: running /etc/cont-init.d/file-structure.sh
cont-init: info: /etc/cont-init.d/file-structure.sh exited 0
s6-rc: info: service legacy-cont-init successfully started
s6-rc: info: service legacy-services: starting
services-up: info: copying legacy longrun lets-encrypt (no readiness notification)
s6-rc: info: service legacy-services successfully started
[21:04:37] INFO: Selected DNS Provider: dns-cloudflare
[21:04:37] INFO: Use propagation seconds: 60
[21:04:37] INFO: Use CloudFlare token
[21:04:37] INFO: Detecting existing certificate type for xxx
Saving debug log to /var/log/letsencrypt/letsencrypt.log
[21:04:38] INFO: Existing certificate using 'ecdsa' key type.
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Certificate not yet due for renewal
```

### Additional information

Originally reported at #1705, but incorrectly closed as stale.

giggio:
Not stale.

giggio:
Not stale.

giggio:
Not stale.

giggio:
Not stale.

GoetzGoerisch:
Could you please verify with the update certbot version.

letsencrypt v5.3.3

If this is still the case I will look deeper into this.

giggio:
> Could you please verify with the update certbot version.
> 
> letsencrypt v5.3.3
> 
> If this is still the case I will look deeper into this.

[...]

PR #3976, opened 2025-03-29
letsencrypt: expose `--force-renewal` in configuration

GoetzGoerisch:
## Summary by CodeRabbit

- **New Features**
  - Introduced a configuration option to force certificate renewal for greater control.
  - Added command line options for Certbot, including Dry Run, Force Renewal, Test Certificates, and Verbose Mode.

- **Documentation**
  - Updated the release notes and added a new section detailing various command line options available for Certbot in the latest version (5.4.5).

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.4.5
+
+- Expose the certbot `--force-renewal` option to the configuration
+
 ## 5.4.4
 
 - Update certbot to 3.3.0
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -292,6 +292,37 @@ If your custom ACME server uses a certificate signed by an untrusted certificate
 
 </details>
 
+<details>
+  <summary>Using Certbot CLI options</summary>
+
+  The configuration supports different Certbot command line options, as detailed below.
+
+  Dry Run: Do a dry-run when requesting a certificate.
+  
+  ```yaml
+  dry_run: true
+  ```
+
+  Force Renewal: Force renew pre-existing certificates.
+
+  ```yaml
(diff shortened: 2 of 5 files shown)


giggio (changes requested):
The docs should include the new option being made avaliable.

giggio (approved):

frenck:
I'm kinda disliking this one. Mostly, because if you leave it in your config, you will keep forcing... For that reason, I don't think we should keep this set in config like this.

../Frenck

agners:
> I'm kinda disliking this one. Mostly, because if you leave it in your config, you will keep forcing...

I second that!

> Does the plugin have the possibility to change the config at the end of a run? Then after a run, if would always be set back to false.

That is possible, and bashio even has support for it with `bashio::addon.option`. However, that changes all configs. There is no per-config write implemented from what I can tell.
https://github.com/hassio-addons/bashio/blob/main/lib/addons.sh#L507-L515

> I agree, that this could be consequence. But as of today the only way to force a renewal is to deinstall and purge the config of the addon.

What is the use case here? In what environment/situation (outside of development) would users want such a feature? Given the downsides, I think I'd rather just use the current work around in development. :shrug:

giggio:
> What is the use case here? In what environment/situation (outside of development) would users want such a feature? Given the downsides, I think I'd rather just use the current work around in development. :shrug: 

See the issue I opened for the use case: #3781

agners:
> See the issue I opened for the use case: #3781

Ah right, and I made a suggestion how to solve it instead: https://github.com/home-assistant/addons/issues/3781#issuecomment-2750576728. I think the per-acme server directory is cleaner here.

Alternatively, maybe we could check the output directory (`/ssl`) and behave different if the user deleted it, either force a renewal if its gone or remove the cache directory in `/data` :thinking: .

PR #3980, opened 2025-04-08
Formally deprecate the DHCP Server add-on

agners:
The maintenance of the ISC DHCP Server has been ended in 2022 already. Alpine Linux dropped the package in Alpine 3.21. Hence the add-on is deprecated as well.

### DIFF
--- a/dhcp_server/config.yaml
+++ b/dhcp_server/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.4.0
+version: 1.5.0
 slug: dhcp_server
 name: DHCP server
 description: A simple DHCP server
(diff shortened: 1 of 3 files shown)


dhcp_server/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 1.4.0
+version: 1.5.0
 slug: dhcp_server
 name: DHCP server
```
frenck, line 4: Maybe this adds to the clarity?

```suggestion
name: DHCP server (DEPRECATED)
```

frenck: Sounds good!

frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3985, opened 2025-04-17
Avoid ordering deprecated add-ons first

agners:
With #3980 with added deprecated as a prefix to the add-on name. However, this leads to sorting it first, which isn't ideal.

Move the deprecation notice to the end. Also add it to all the currently deprecated add-ons.

## Summary by CodeRabbit

- **Chores**
  - Updated addon names in configuration files to clearly indicate deprecated status for DHCP Server, Google Assistant SDK, and TellStick. No functional changes were made.

### DIFF
--- a/dhcp_server/config.yaml
+++ b/dhcp_server/config.yaml
@@ -1,7 +1,7 @@
 ---
 version: 1.5.0
 slug: dhcp_server
-name: "[Deprecated] DHCP server"
+name: "DHCP server [deprecated]"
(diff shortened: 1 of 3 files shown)


agners:
The build failure is unfortunately expected, that is part of the reason why google_assistant add-on got deprecated.

Since this only changes not build relevant metadata, merging it is fine even with the build failure.

frenck (approved):
Oh right! Makes sense 👍 

../Frenck

PR #3988, opened 2025-04-18
whisper: Improve support for local custom Whisper models

DT-art1:
## Proposed change
This updates the 'whisper' add-on to clarify how to configure and use a local custom Whisper model.

### Changes:
- Added configuration mapping to access local models.
- Updated documentation about `custom_model` usage.

## Summary by CodeRabbit

- **Documentation**
  - Enhanced instructions for using a local custom Whisper model, specifying directory paths and configuration steps.
  - Added new changelog entries highlighting recent documentation and configuration updates.

- **Chores**
  - Updated configuration to enable access to local models by adding a new mapping section and backup exclusions.

### DIFF
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -12,6 +12,9 @@ discovery:
   - wyoming
 backup_exclude:
   - "models*"
+map:
+  - type: addon_config
+    read_only: false
 options:
   model: auto
   language: en
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.4.0
+version: 2.5.0
 slug: whisper
 name: Whisper
 description: Speech-to-text with Whisper
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -12,7 +12,6 @@ discovery:
   - wyoming
 backup_exclude:
   - "models*"
-  - "/data/addon_configs/*"
 map:
   - addon_config:rw
 options:
--- a/whisper/config.yaml
+++ b/whisper/config.yaml
@@ -12,9 +12,9 @@ discovery:
   - wyoming
 backup_exclude:
   - "models*"
+  - "/data/addon_configs/*"
 map:
-  - type: addon_config
-    read_only: false
+  - addon_config:rw
 options:
   model: auto
   language: en
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+- Added configuration mapping to access local models.
+- Updated documentation about `custom_model` usage.
+
 ## 2.4.0
 
 - Add "auto" for model and beam size (0) to select values based on CPU
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Unreleased]
+## 2.5.0
 
 - Added configuration mapping to access local models.
 - Updated documentation about `custom_model` usage.
--- a/whisper/DOCS.md
+++ b/whisper/DOCS.md
@@ -86,7 +86,8 @@ See [this discussion](https://github.com/openai/whisper/discussions/963) for an 
 
 ## Backups
 
-Whisper model files can be quite large, so they are automatically excluded from backups. The models will be re-downloaded when the backup is restored.
+Whisper model files can be quite large, so they are automatically excluded from backups. The models will be re-downloaded when the backup is restored for a remote model.
+If you're using a local custom Whisper model, you will need to manually copy the model directory again after restoring a backup.
 
 ## Support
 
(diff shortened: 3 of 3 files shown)


agners (changes requested):

synesthesiam:
@DT-art1 The backup exclude rule is `models*`. I think we should advise people to put their custom models in a `models` sub-directory so they aren't included in the backup. What do you think?

DT-art1:
Thanks for pointing out the backup aspect. I agree that organizing custom models in a separate sub-directory could be a good approach, but I opted to add an additional rule **/data/addon_configs/*** to exclude local models from the backup entirely. I believe this approach is simpler and more user-friendly for those who may not have a dedicated sub-directory for their custom models. Let me know if you hare any further thoughts on this.

**Test output**
```
2025-04-25 09:47:26.674 INFO (MainThread) [supervisor.addons.addon] Building backup for add-on local_whisper
2025-04-25 09:47:26.675 DEBUG (MainThread) [supervisor.homeassistant.websocket] Sending: {'type': <WSType.SUPERVISOR_EVENT: 'supervisor/event'>, 'data': {'event': <WSEvent.JOB: 'job'>, 'data': {'name': 'addon_begin_backup', 'reference': 'local_whisper', 'uuid': 'b94740abbeea4622a5804a14df9cfe4e', 'progress': 0, 'stage': None, 'done': True, 'parent_id': 'f02932f76792429f9c2c552ce7df52cd', 'errors': [], 'created': '2025-04-25T07:47:26.665737+00:00'}}, 'id': 2274}
2025-04-25 09:47:26.675 DEBUG (MainThread) [supervisor.homeassistant.websocket] Received: WSMessage(type=<WSMsgType.TEXT: 1>, data='{"id":2274,"type":"result","success":true,"result":null}', extra='')
2025-04-25 09:47:26.683 DEBUG (SyncWorker_8) [supervisor.addons.addon] Ignoring /data/addons/data/local_whisper/.locks/models--rhasspy--faster-whisper-base-int8 because of models*
2025-04-25 09:47:26.683 DEBUG (SyncWorker_8) [supervisor.addons.addon] Ignoring /data/addons/data/local_whisper/models--rhasspy--faster-whisper-base-int8 because of models*
2025-04-25 09:47:26.683 DEBUG (SyncWorker_8) [supervisor.addons.addon] Ignoring /data/addon_configs/local_whisper because of /data/addon_configs/*
2025-04-25 09:47:26.684 INFO (MainThread) [supervisor.addons.addon] Finish backup for addon local_whisper
```

synesthesiam:
Just want to confirm with @agners that the backup exclude rule will work. Otherwise, looks good to me 👍

whisper/config.yaml
```diff
@@ -12,6 +12,9 @@ discovery:
   - wyoming
 backup_exclude:
   - "models*"
+  - "/data/addon_configs/*"
```
agners, line 15: Hm, I see this works since at least today, Supervisor checks against the full paths as seen by Supervisor (https://github.com/home-assistant/supervisor/blob/2025.04.1/supervisor/addons/addon.py#L1237-L1254).

> Thanks for pointing out the backup aspect. I agree that organizing custom models in a separate sub-directory could be a good approach, but I opted to add an additional rule /data/addon_configs/* to exclude local models from the backup entirely. I believe this approach is simpler and more user-friendly for those who may not have a dedicated sub-directory for their custom models. Let me know if you hare any further thoughts on this.

True, excluding all of add-on config makes it simpler today, but it also means that if this add-on were to add something else which needs to be stored in the add-on specific config area (e.g. a complex/separate configuration file) we won't have backups of it.

I think personally I'd prefer if we'd keep the models in a separate directory just in case we want to store other data. But maybe we can reasonably rule out that we'll have something else for this add-on? Modifying the backup exclude rule later will be harder, as we might suddenly include models we didn't intend to.

@synesthesiam thoughts?

synesthesiam: I'd prefer to keep models in a separate directory just in case.

DT-art1: I updated the documentation and restored the backup rules as it was before. The following log shows that no additional rule is required:
```
2025-04-29 20:37:41.673 INFO (MainThread) [supervisor.addons.addon] Building backup for add-on local_whisper
...
2025-04-29 20:37:41.689 DEBUG (SyncWorker_4) [supervisor.addons.addon] Ignoring /data/addons/data/local_whisper/.locks/models--rhasspy--faster-whisper-base-int8 because of models*
2025-04-29 20:37:41.691 DEBUG (SyncWorker_4) [supervisor.addons.addon] Ignoring /data/addons/data/local_whisper/models--rhasspy--faster-whisper-base-int8 because of models*
2025-04-29 20:37:41.693 DEBUG (SyncWorker_4) [supervisor.addons.addon] Ignoring /data/addon_configs/local_whisper/models because of models*
2025-04-29 20:37:41.696 INFO (MainThread) [supervisor.addons.addon] Finish backup for addon local_whisper
```

agners (approved):

agners:
LGTM, thanks!

PR #2362, opened 2022-01-26
Add the ACME custom server args for DNS challenges — fixes #1769

imgrant:
As described in #1769, if using DNS challenge, the custom ACME server parameter(s) are not added to the certbot command invocation, hence the setting is ignored.

Issue #1769, opened 2021-01-22
letsencrypt: Add custom ACME server option does not work if challenge is DNS


jingsno:
## The problem

Add custom ACME server option does not work if the challenge is "dns"
"${ACME_CUSTOM_SERVER_ARGUMENTS[@]}" needs to be added to the "dns" challenge parameters aswell

## Environment

- Add-on with the issue: LetsEncrypt
- Add-on release with the issue: 
- Last working add-on release (if known): 
- Operating environment (OS/Supervised): 

## Problem-relevant configuration

```yaml

```

## Traceback/Error logs

```txt

```

## Additional information
From letsencrypt/rootfs/etc/services.d/lets-encrypt/run

if bashio::config.has_value 'acme_server' ; then
    ACME_CUSTOM_SERVER_ARGUMENTS+=("--server" "${ACME_SERVER}")

SNIP

if [ "$CHALLENGE" == "dns" ]; then
    certbot certonly --non-interactive --keep-until-expiring --expand \
        --email "$EMAIL" --agree-tos \
        --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
        --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}" "${PROVIDER_ARGUMENTS[@]}"
else
    certbot certonly --non-interactive --keep-until-expiring --expand \
        --email "$EMAIL" --agree-tos \
        --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
        --preferred-challenges "$CHALLENGE" "${DOMAIN_ARR[@]}" "${ACME_CUSTOM_SERVER_ARGUMENTS[@]}" --standalone
fi

jingsno:
Still a problem

imgrant:
Still exists and a PR submitted in #2362 with the fix, but closed as stale. 🤷‍♂️

PR #3994, opened 2025-04-21
Fix custom ACME server option with DNS challenge

83noit:
Fixed a long-standing issue (#1769) where the custom ACME server parameter was incorrectly applied only to HTTP challenges and not other challenge types. Although a previous pull request (#2362) attempted to address this, it remained unresolved.

This fix ensures the custom ACME server parameter is consistently included in the certbot command for all challenge mechanisms.

Additionally, this pull request includes a cosmetic improvement by replacing several inconsistent if conditions with a more readable case statement.


83noit:
Just noticed that PR #3976 is also modifying this same section of code.

agners:
> Just noticed that PR #3976 is also modifying this same section of code.

Usually whatever is ready first I merge. The next PR then needs to rebase/fix the merge currently, but that is how it is :shrug: . That PR in question isn't ready to be merged, so don't worry about the conflict.

agners (changes requested):
Otherwise LGTM on first sight.

I've also just merged another PR affecting the Let's Encrypt add-on. Can you increment the version number and adjust the CHANGELOG.md accordingly? Thanks!

agners (approved):
LGTM, nice, thanks!

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -364,27 +399,18 @@ if [ "${VERBOSE}" = "true" ]; then
...
+# Provider specific arguments for a DNS challenge or
+# run a standalone webserver for an HTTP challenge.
+certbot certonly --non-interactive --keep-until-expiring --expand \
+    --email "$EMAIL" --agree-tos \
+    "${KEY_ARGUMENTS[@]}" \
+    "${ADDITIONAL_ARGS[@]}" \
+    --cert-name "${DOMAIN_ARR[1]}" "${DOMAIN_ARR[@]}" \
+    --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
+    --preferred-challenges "$CHALLENGE" \
+    "${PROVIDER_ARGUMENTS[@]:-'--standalone'}" \
```
agners, line 411: Actually, realized in my testing that this leads to `'--standalone'` including apostrophe being passed:

```
[16:06:22] INFO: Selected HTTP verification
usage: 
certificate. 
certbot: error: unrecognized arguments: '--standalone'
```

Without the apostrophe this is not very readable :see_no_evil: 

We could just set `PROVIDER_ARGUMENTS` with `--standalone` further up, or use a default assignment just before calling, or just go with this :shrug: :

```suggestion
    "${PROVIDER_ARGUMENTS[@]:---standalone}" \
```

agners (changes requested):

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
83noit: Good catch!
I think that dealing with `--standalone` further up is a tad more readable than the corrected expansion.

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -364,27 +399,17 @@ if [ "${VERBOSE}" = "true" ]; then
...
-        "${KEY_ARGUMENTS[@]}" \
-        "${ADDITIONAL_ARGS[@]}" \
-        --cert-name "${DOMAIN_ARR[1]}" "${DOMAIN_ARR[@]}" \
-        --config-dir "$CERT_DIR" --work-dir "$WORK_DIR" \
-        --preferred-challenges "$CHALLENGE" "${ACME_CUSTOM_SERVER_ARGUMENTS[@]}" --standalone \
-        --preferred-chain "ISRG Root X1" \
-        "${EAB_ARGUMENTS[@]}"
-fi
+# Provider specific arguments for a DNS challenge or
+# run a standalone webserver for an HTTP challenge.
```
agners, line 403: This comment is a bit out of place right now. I'll just remove it.

```suggestion
```

agners (approved):

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
agners, line 403: Meant to remove both lines.
```suggestion
```

83noit:
Thanks for the review!

agners:
Thanks for the fix and cleanup! :pray:

PR #3995, opened 2025-04-22
ssh: Bump CLI to 4.38.0

agners:
While at it, also bump libwebsockets to 4.3.5.


frenck (approved):
Thanks, @agners 👍 

../Frenck

PR #3996, opened 2025-04-23
letsencrypt: Correct Root CA handling for custom ACME servers.

83noit:
Ensures proper usage of the Root CA chain and ignores Let's Encrypt-specific staging options when a custom ACME server is configured.


letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.4.7
+
+- Fix root certificate format
```
agners, line 5:
```suggestion
- Fix root certificate format when using custom ACME server
```

agners (approved):
Thanks, otherwise LGTM!

letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -328,13 +329,25 @@ if bashio::config.has_value 'acme_server' ; then
     ACME_ARGUMENTS+=("--server" "${ACME_SERVER}")
 
     if bashio::config.has_value 'acme_root_ca_cert'; then
-        echo "${ACME_ROOT_CA}" > /tmp/root-ca-cert.crt
+        # The value comes as a continuous string, but lines breaks around PEM
+        # headers and footers are required for the certificates to be parsed.
+        echo "${ACME_ROOT_CA}" | sed 's/----- /-----\n/g'  | sed 's/ -----/\n-----/g'  > /tmp/root-ca-cert.crt
```
agners, line 334: Hm, I see this is caused by the text input field in the UI.

When using YAML with appropriate folding like so:

```yaml
acme_root_ca_cert: |-
  -----BEGIN CERTIFICATE-----
  MIIFazCCA1OgAwIBAgIRAIIQz7DSQONZRGPgu2OCiwAwDQYJKoZIhvcNAQELBQAw
   ...
```

Which makes me wonder, does this maybe break a properly supplied string? :thinking: 

The config rendering is not ideal, unfortunately. But we might miss some metadata here, e.g. str could be a single line or multiline string...

agners: Hm, I see thanks to the spaces your `sed` substitute regex a properly passed multiline string doesn't get mangled. I don't really like that post processing all that much, but  since we don't have a better configuration frontend I guess it's the best we can do. We can remove and update the config automatically once we have a proper multiline text input field.

83noit: Thanks for the review.

I took a quick peek at options for multiline input in the config frontend, but since I'm not really familiar with it, I didn't stumble upon anything. Good to hear it's potentially in the cards for the future!

And yeah, as long as the headers and footers are on their own lines, PEM seems pretty forgiving from what I've seen.

83noit: 🙈
It turns out there's a better approach: preserving line returns for acme_root_ca_cert within the YAML file with `acme_root_ca_cert:  |` .

The current UI default of `>-` can still cause parsing issues with longer certificate chains.

I'm unsure how to modify the UI's default behavior at the moment, but I'll investigate. I'll push a branch that removes the sed workarounds and either corrects the UI behavior or updates the add-on documentation accordingly in the next few days.

PR #4006, opened 2025-04-29
Fix trust store when using a custom ACME server

83noit:
Prior to #3999, the code replaced Requests' entire trust store with only the custom ACME server's root certificate. This broke Certbot's ability to establish trusted SSL connections with other servers, such as DNS providers for challenges.

#3999 added the custom certificate to the local system trust store, but this isn't sufficient because Requests uses its own internal trust store.

The final solution involves overriding Requests' trust store to use a combined store containing both the default trusted certificates and the custom ACME server's root certificate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with the trust store when using a custom ACME server, ensuring proper handling of custom root certificates.

- **Chores**
  - Updated documentation with a new changelog entry for version 5.4.9.
  - Bumped version number to 5.4.9 in the configuration file.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.4.9
+
+- Fix pending issue in 5.4.8 with trust store when using a custom ACME server
+
 ## 5.4.8
 
 - Further improve root certificate handling when using a custom ACME server
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 5.4.8
+version: 5.4.9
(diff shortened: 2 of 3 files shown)


83noit:
My apologies for the rapid succession of PRs. The custom server functionality presented two distinct layers of issues. To avoid generating numerous certificates that would require revocation, I tested these layers individually rather than end-to-end. In #3999, I used openssl to inspect the certificate and verify SSL connections but didn't realize that the Requests library operates independently of the system's trust store.

donicrosby:
If we could get this merged that would be ideal.

I'm currently unable to issue new certificates due to the fact that the REQUESTS_CA_BUNDLE was removed in the prior PR

agners (approved):
LGTM, thanks!

PR #4007, opened 2025-05-01
Added detail on formatting of ACME custom root certificate in LetsEncrypt UI config flow.

linkedupbits:
The example config is fine for editing the yaml, but for people using HAOS etc. there needs to be explanation on how to input the ACME custom root certificate via the UI.

## Summary by CodeRabbit

- **Documentation**
  - Updated instructions to clarify the required single-line format for the "ACME Root CA Certificate" field in the UI configuration flow.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -251,6 +251,9 @@ If your custom ACME server uses a certificate signed by an untrusted certificate
     -----END CERTIFICATE-----
   ```
 
+When you use the UI config flow, the "ACME Root CA Certificate" must be formatted as:   
+-----BEGIN CERTIFICATE-----(SPACE)(KEY value all on one line)(space)-----END CERTIFICATE-----
+
 When you specify a custom ACME server, the *Dry Run* and *Issue test certificates* options, which are intended for use with the [Let's Encrypt staging server](https://letsencrypt.org/docs/staging-environment/), are automatically disregarded.
 
 </details>


linkedupbits (commented):
Have signed the CLA.

agners:
> The example config is fine for editing the yaml, but for people using HAOS etc. there needs to be explanation on how to input the ACME custom root certificate via the UI.

From what I tested, just pasting a properly formatted certificates (with the newlines) leads to exactly that formatting. So is documentation really required if pasting into the text field just formats it automatically correct? :thinking: 

/cc @83noit

83noit:
Properly formatted certificates (with PEM headers and footers on separate lines) can be pasted, which is the expected behavior. The UI-to-YAML conversion might introduce spaces instead of line breaks, but the current logic handles this.

However, pasting improperly formatted PEM data, such as:
- `-----BEGIN CERTIFICATE-----FOOO -----END CERTIFICATE-----`
- `-----BEGIN CERTIFICATE -----FOOO-----END CERTIFICATE-----`
will fail.

To enhance the robustness of the [formatting logic](https://github.com/home-assistant/addons/blob/a657d9b1a04681b2bef177df159e1bfb1e86f3c9/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L337C1-L338C1), consider replacing:
-  ` | sed 's/----- /-----\n/g' | sed 's/ -----/\n-----/g’ \`
-  ` | sed 's/-----BEGIN CERTIFICATE-----/\n-----BEGIN CERTIFICATE-----\n/' | sed 's/-----END CERTIFICATE-----/\n-----END CERTIFICATE-----\n/' \`

This approach could offer a more intuitive experience compared to relying on strict formatting or specific documentation.

agners:
> Properly formatted certificates (with PEM header and footer on separate lines) can be pasted, which is the expected behavior. [..]

With that in mind, I don't really see a point in adding this explicit documentation here. We have the example how-to use multiline YAML already, in case folks run into problems.

PR #4009, opened 2025-05-05
Add Z-Wave region option

MartinHjelmare:
- If the user selects `Automatic`, which is the default, we get the country from Core and use that to look up an appropriate region.
- The country to radio frequency region map was made by referencing this page:
  https://www.silabs.com/wireless/z-wave/global-regions
- If the country hasn't approved Z-Wave yet, the Default (EU) region will be used when Automatic setting is chosen.

## Todo

- [x] Bump minimum home assistant version to 2025.5.0, to allow Core to update country in Supervisor.
- [x] Wait until core-2025.5.0 has been released.

## Summary by CodeRabbit

- **New Features**
  - Introduced a configurable radio frequency region option for the Z-Wave JS add-on, enabling users to select specific regional settings or rely on automatic detection.
- **Documentation**
  - Added details about the new `rf_region` setting and its valid options in the add-on documentation and translation files.
- **Chores**
  - Updated compatibility requirements to support Home Assistant Core version 2025.5.0 and above.


zwave_js/DOCS.md
```diff
@@ -163,6 +163,28 @@ When `log_to_file` is true, Z-Wave JS will create a log file for each
...
 
+### Option `rf_region` (optional)
+
+This setting tells the add-on what radio frequency region the controller should use.
+Valid options are:
+
+- Automatic
+- Australia/New Zealand
+- China
+- Default (EU)
```
AlCalzone, line 174: See comment below. `Default` is not a valid setting.

AlCalzone (changes requested):
Do we want to set the correct TX power options for ZWA-2 in this PR too, or should that be separate?

MartinHjelmare:
@AlCalzone will the driver handle if a region is configured that isn't supported by the controller?

AlCalzone:
If the controller does not support the desired region, no reconfiguration will happen. There's no error or anything in that case, just a log message.

MartinHjelmare:
> Do we want to set the correct TX power options for ZWA-2 in this PR too, or should that be separate?

We can do that separately.

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -24,6 +27,275 @@ readonly DOCS_EXAMPLE_KEY_4="CF338FE0CB99549F7C0EA96308E5A403"
...
+    ["Hong Kong"]=3
+    ["India"]=5
+    ["Israel"]=6
+    ["Russia"]=7
+    ["China"]=8
+    ["USA (Long Range)"]=9
+    ["Europe (Long Range)"]=11
+    ["Japan"]=32
+    ["Korea"]=33
+    ["Unknown"]=254
```
AlCalzone, line 295:
```suggestion
```
I don't think it makes sense to include this one.

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -107,6 +379,22 @@ else
...
+    rf_region=$(bashio::config 'rf_region')
+fi
+rf_region_integer=${rf_region_integer_map["${rf_region}"]}
+
+if [[ "${rf_region_integer}" -eq 255 ]]; then
+    rf_region_json="{}"
+    bashio::log.info "Using default rf region settings"
+else
+    rf_region_json=$(jq -n --argjson region "${rf_region_integer}" '{ region: $region }')
+    bashio::log.info "Setting rf region to (${rf_region})"
```
AlCalzone, line 395:
```suggestion
    bashio::log.info "Setting RF region to (${rf_region})"
```

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -107,6 +379,22 @@ else
...
+    rf_region=${country_rf_region_map[$(bashio::supervisor.country)]}
+    bashio::log.info "rf region set to Automatic"
+else
+    rf_region=$(bashio::config 'rf_region')
+fi
+rf_region_integer=${rf_region_integer_map["${rf_region}"]}
+
+if [[ "${rf_region_integer}" -eq 255 ]]; then
+    rf_region_json="{}"
+    bashio::log.info "Using default rf region settings"
```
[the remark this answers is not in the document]
AlCalzone: AFAIK the common abbreviation is uppercase: https://en.wikipedia.org/wiki/Radio_frequency

zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.14.0
```
AlCalzone, line 3:
```suggestion
## 0.14.0

### Breaking changes
- Version 0.14.0 of the add-on requires version 2025.5.0 or higher of Home Assistant Core.
```

AlCalzone (changes requested):

zwave_js/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.14.0
+
+### Breaking changes
```
MartinHjelmare, line 5:
```suggestion
### Breaking changes

```

balloob:
The default region (Europe), does that support long range ?

AlCalzone:
The name "Default (EU)" is a bit misleading. Since adding region support to Z-Wave JS, I learned that it means _look up the region from the manufacturer token storage and default to EU_ if that is empty. We don't know which region it actually refers to for a given controller. Also it cannot actually be set: https://github.com/home-assistant/addons/pull/4009#discussion_r2074005690

The region "Europe" will automatically be upgraded to "Europe (Long Range)" by Z-Wave JS if supported by the controller, just like for USA.

AlCalzone (changes requested):
Maybe I'm misunderstanding the code, but the region can not be set to `Default (EU)` and this is not a valid option for the driver option.
In these cases we should just not set the region at all.

zwave_js/rootfs/etc/cont-init.d/config.sh
```diff
@@ -107,6 +377,23 @@ else
...
+    rf_region="${country_rf_region_map[$country]:-"Default (EU)"}"
+    bashio::log.info "RF region set to Automatic: ${rf_region}"
+else
+    rf_region=$(bashio::config 'rf_region')
+fi
+rf_region_integer=${rf_region_integer_map["${rf_region}"]}
+
+if [[ "${rf_region_integer}" -eq 255 ]]; then
+    rf_region_json="{}"
+    bashio::log.info "Using default RF region settings"
```
MartinHjelmare, line 391: That's handled here. :point_up:

edenhaus (approved):

PR #4015, opened 2025-05-08
Add apple_model option

LenaWil:
[Relevant feature request](https://community.home-assistant.io/t/add-support-for-user-configurable-fruit-model-in-the-samba-addon/887170).

I do not have a testing environment set up, so I have no idea if this works or not, but it seemed like a really easy change, hence the pull request.

`fruit:aapl = yes` is probably optional. And not sure about the default value, I put in the homepod value but might better to leave it out.

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 12.5.2
+- Add an option to choose which icon appears for server on macOS Finder.
+
 ## 12.5.1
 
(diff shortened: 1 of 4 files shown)


agners:
Hi @LenaWil, thanks for the PR.

While the PR generally looks good, I am a bit hesitant to merge. It feels a bit like unnecessary feature creep :sweat_smile: :see_no_evil: . You can also copy that particular add-on into your [own repository](https://developers.home-assistant.io/docs/add-ons/repository) if you want.

agners:
Closing with the above in mind.

PR #4016, opened 2025-05-08
Remove incorrect `--find-links` arguments that were being passed to pip

ItsHarper:
While the specific URLs were different, both of them assume a directory structure at https://wheels.home-assistant.io that no longer exists (I assume they probably did at one point, I know that `musllinux` wasn't always the name used).

I imagine the arguments I've removed were intended to accomplish roughly the same thing as this, which is now applied to all pip installations in all Home Assistant docker containers:

https://github.com/home-assistant/docker-base/blob/597ee655f913037c4bc9c8d3c2c6b887a7c1efc8/alpine/rootfs/etc/pip.conf#L3

## Summary by CodeRabbit

- **Chores**
  - Updated Python package installation process in Docker images to use default package sources, removing reliance on a custom wheel repository.

### DIFF
--- a/letsencrypt/Dockerfile
+++ b/letsencrypt/Dockerfile
@@ -51,8 +51,7 @@ RUN \
         musl-dev \
         openssl-dev \
         cargo \
-    && pip3 install --no-cache-dir --find-links \
-        "https://wheels.home-assistant.io/alpine-$(cut -d '.' -f 1-2 < /etc/alpine-release)/${BUILD_ARCH}/" \
+    && pip3 install--no-cache-dir \
         acme==${ACME_VERSION} \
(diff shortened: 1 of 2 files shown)


ItsHarper:
I'd of course prefer to double-check that I'm still able to build the modified, but my current system is a bit on the exotic side, and I'm truly not set up for any of the documented ways of building addon containers right now. It hardly seems worth getting it working for the level of change this is.

agners (approved):
Nice find! Yeah this seems really rely on a outdated directory structure on our wheels server, and since the base images have the repository pre-configured this is no longer required. I also checked that this is indeed the case.

As the URLs didn't exist really, this change has no real world impact actually. So its safe to skip the usual CHANGELOG.md entry. This just code cleanup.

Thanks for your PR! :pray:

PR #4017, opened 2025-05-09
silabs-flasher: Drop firmwares and update to latest universal flasher

agners:
The firmware files bundled with this add-on are outdated at this point. Since firmware updates for Home Assistant Yellow and ZBT-1 are handled by Home Assistant Core anways, let's simply drop the bundled firmwares.

While at it, update the add-on to use the latest universal flasher and base images.

## Summary by CodeRabbit

- **New Features**
  - Updated to require users to provide a firmware URL for flashing Silicon Labs radios.
- **Bug Fixes**
  - Simplified firmware flashing process by removing automatic device detection and bundled firmware files.
- **Documentation**
  - Updated README and changelog to clarify that official firmware updates for Home Assistant Yellow and ZBT-1 are now managed by Home Assistant Core.
  - Improved formatting and usage instructions in documentation.
- **Chores**
  - Upgraded base Alpine Linux and tool versions for improved compatibility and security.


puddly:
Let me double check if we rely on the flasher addon anywhere in Core to install Zigbee firmware. If we do, this will have to be restricted to 2025.6.0 and newer.

agners:
> Unfortunately, we do: https://github.com/home-assistant/core/blob/763f2bcfcc697a59802c873078cd3750141013e1/homeassistant/components/homeassistant_hardware/firmware_config_flow.py#L261-L271
> 
> This will have to be replaced to use the firmware flasher in Core directly.

Hm I see. I guess I could also retain the firmware this time around :thinking: . But probably cleaner to remove that reference and then restrict the add-on to the Core release where it is fixed or newer.

PR #4061, opened 2025-06-30
[openthread_border_router] Remove firmware flashing from the addon

puddly:
Home Assistant Core 2025.7.0 will install the correct firmware before installing this addon, removing the need for `autoflash_firmware`. This will allow us to release OpenThread RCP firmware updates independently from the addon.

## Summary by CodeRabbit

* **Documentation**
  * Clarified firmware requirements for 802.15.4 radios and removed the configurable "autoflash_firmware" option.
  * Added changelog entry for version 2.14.0 noting removal of onboard firmware flashing.

* **Chores**
  * Removed built-in firmware flashing tooling and related service setup.
  * Updated configuration to require Home Assistant Core 2025.7.0.
  * Added startup handling for any legacy autoflash setting.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.13.1
+## 2.14.0
 - Remove firmware flashing from the addon, this is now handled by Core 2025.7.0.
 
 ## 2.13.0
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.13.1
+version: 2.14.0
 slug: openthread_border_router
 name: OpenThread Border Router
 description: OpenThread Border Router add-on


frenck:
Marking this PR as a draft, as it requires Home Assistant 2025.7, which is not available yet. Releasing this now, means nobody can install it.

../Frenck

puddly:
Oh no, that means that we can't release this PR at all for at least a few months 🤔.

It seems like addons don't actually have any concept of releases, right? If the latest/only version is not compatible, the addon can't be installed?

agners:
> Oh no, that means that we can't release this PR at all for at least a few months 🤔.

We've required add-on to use a just released Core version already in the past. Usually when folks install new add-ons they most likely work on their HA installation anyways, and it is not all that big of a deal to also update Home Assistant Core. So no, I wouldn't hold off with this PR. Just wait until 2025.7.0 is officially available.

> It seems like addons don't actually have any concept of releases, right? If the latest/only version is not compatible, the addon can't be installed?

Yes, we only ever have the latest version available. Unless you restore a backup, then you'll get whatever version that backup has.

puddly:
This would prevent anyone using 2025.6.x or below from setting up a ZBT-1 or Yellow as a Thread border router from Core, no? They'd click on the "Thread" option in the config flow and the addon would fail to install?

agners:
> This would prevent anyone using 2025.6.x or below from setting up a ZBT-1 or Yellow as a Thread border router from Core, no? They'd click on the "Thread" option in the config flow and the addon would fail to install?

Yeah. But at that point you are trying to setup a new communication protocol on your Home Assistant installation, which is a major operation on its own. IMHO, it is fine to require the user to update to the latest Core version. You also get the best experience with that...

agners:
I just realized that with HAOS 16.0 the serial port paths changed (again), rendering the Yellow detection non-functional with the current OpenThread Border Router :see_no_evil: 

```
-----------------------------------------------------------
 Add-on: OpenThread Border Router
 OpenThread Border Router add-on
-----------------------------------------------------------
 Add-on version: 2.13.0
 You are running the latest version of this add-on.
 System: Home Assistant OS 16.0  (aarch64 / yellow)
 Home Assistant Core: 2025.1.0.dev202412160236
 Home Assistant Supervisor: 2025.07.2.dev0901
-----------------------------------------------------------
 Please, share the above information when looking for help
 or support in, e.g., GitHub, forums or the Discord chat.
-----------------------------------------------------------
s6-rc: info: service banner successfully started
s6-rc: info: service universal-silabs-flasher: starting
[10:32:09] INFO: The selected serial port is not a USB device.
[10:32:09] WARNING: No firmware found for the selected device, assuming firmware is installed.
```


The correct path now would be:

```
/sys/devices/platform/soc/fe201800.serial/fe201800.serial:0/fe201800.serial:0.0/tty/ttyAMA1
```

However, I don't think we should bother at this point. Let's get this PR merged, and advice people to upgrade to latest Core 2025.7.x to get the new built-in firmware management experience :sunglasses: 

/cc @sairon

agners:
Marking this as ready for review as Core 2025.7 is out since a while, 2025.7 and 2025.8 make up 60% of the user base, and also most new users of the OTBR add-on will have most likely a new version. So very few users get forced to upgrade if they want to *start* using the OTBR at this point.

openthread_border_router/config.yaml
```diff
@@ -28,7 +28,6 @@ options:
   device: null
   baudrate: "460800"
   flow_control: true
-  autoflash_firmware: true
```
agners, line 31: To avoid warnings in the Supervisor logs this release should also remove the option from existing installations via bashio. The documentation has examples on how to do this: https://developers.home-assistant.io/docs/add-ons/configuration#options--schema

agners (approved):

openthread_border_router/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.13.1
```
sairon: I think it's worth bumping at least the minor version here given the character of the changes.

Issue #4028, opened 2025-05-21
Matter Server: --primary-interface not recognizing VLAN


mwolter805:
### Describe the issue you are experiencing

Hello,
I am having an issue where the matter server is not honoring the "--primary-interface enp0s31f6.21" extra Matter Server argument. .21 is the vlan I would like the matter server to use as it's primary interface.

The hub I wish to connect to via Matter is not trusted and therefore placed on a vlan isolated from the internet and also isolated other vlans. I have added a vlan to HA for the same vlan as the untrusted hub. I can ping the untrusted hub on its IPv6 link-local address, so I know HA has connectivity to it through this vlan. However when attempting to add the hub to the matter server it cannot connect. Reviewing the matter server logs I see the following

[17:13:25] INFO: Using 'enp0s31f6' as primary network interface.
and not enp0s31f6.21

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

Matter Server

### What is the version of the add-on?

8.0

### Steps to reproduce the issue

1. Add --primary-interface enp0s31f6.21 to the Extra Matter Server arguments 
2. Restart and review the logs.
...

## System Information

version | core-2025.5.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.13.3
os_name | Linux
os_version | 6.12.23-haos
arch | x86_64
timezone | America/Los_Angeles
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 1789
Downloaded Repositories | 33

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | February 17, 2026 at 4:00 PM
relayer_connected | true
relayer_region | us-east-1
remote_enabled | true
remote_connected | true
alexa_enabled | false
google_enabled | false
cloud_ice_servers_enabled | true
remote_server | us-east-1-7.ui.nabu.casa
certificate_status | ready
instance_id | 0fd36e6e6c524f25bd11489ed0bd2ce0
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

[...]

PR #4070, opened 2025-07-06
Matter server: primary_interface config option

elestril:
Allow to configure primary_interface in addon json

That was probably the intent of https://github.com/home-assistant/addons/pull/3468, but the run script overrides `--primary_interface` even when provided in `matter_server_args`

Closes: #4028

See also: https://community.home-assistant.io/t/home-assistant-cant-add-matter-devices-that-are-in-a-different-vlan/750324/54

```
[23:15:19] INFO: Using 'end0.2' as primary network interface.
[...]
2025-07-05 23:29:34.356 (MainThread) INFO [matter_server.server.device_controller] Starting Matter commissioning with code using Node ID 19.
2025-07-05 23:29:35.656 (Dummy-2) INFO [chip.ChipDeviceCtrl] Established secure session with Device
2025-07-05 23:29:40.972 (Dummy-2) INFO [chip.ChipDeviceCtrl] Commissioning complete
2025-07-05 23:29:40.973 (MainThread) INFO [matter_server.server.device_controller] Commissioned Node ID: 19 vs 19
2025-07-05 23:29:40.974 (MainThread) INFO [matter_server.server.device_controller] Matter commissioning of Node ID 19 successful.
2025-07-05 23:29:40.974 (MainThread) INFO [matter_server.server.device_controller] Interviewing node: 19
2025-07-05 23:29:41.898 (MainThread) INFO [matter_server.server.device_controller] <Node:19> Setting-up node...
2025-07-05 23:29:41.902 (MainThread) INFO [matter_server.server.device_controller] <Node:19> Setting up attributes and events subscription.
2025-07-05 23:29:42.164 (MainThread) INFO [matter_server.server.device_controller] <Node:19> Subscription succeeded with report interval [1, 60]
2025-07-05 23:29:42.166 (MainThread) INFO [matter_server.server.device_controller] Commissioning of Node ID 19 completed.
```

### DIFF
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.0.1
+
+- Allow to configure primary_device
+
 ## 8.0.0
 
 - Bump Python Matter Server to [8.0.0](https://github.com/home-assistant-libs/python-matter-server/releases/tag/8.0.0)
--- a/matter_server/config.yaml
+++ b/matter_server/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 8.0.0
+version: 8.0.1
 slug: matter_server
 name: Matter Server
 description: Matter WebSocket Server for Home Assistant Matter support.
@@ -28,6 +28,7 @@ options:
   log_level_sdk: error
   beta: false
   enable_test_net_dcl: false
+  primary_interface:: null
 schema:
   log_level: list(verbose|debug|info|warning|error|critical)
   log_level_sdk: list(automation|detail|progress|error|none)?
@@ -38,6 +39,7 @@ schema:
     - str?
   matter_server_version: str?
   matter_sdk_wheels_version: str?
+  primary_interface: str?
 ports:
   5580/tcp: null
 stage: stable
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -53,7 +53,11 @@ if bashio::config.true "enable_test_net_dcl"; then
     extra_args+=('--enable-test-net-dcl')
 fi
 
-primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+if bashio::config.has_value "primary_interface"; then
+  primary_interface=$(bashio::config 'primary_interface')
+else
+  primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+fi
 
 # Try fallback method (e.g. in case NetworkManager is not available)
 # shellcheck disable=SC2086
--- a/matter_server/translations/en.yaml
+++ b/matter_server/translations/en.yaml
@@ -42,5 +42,10 @@ configuration:
     description: >-
       Install custom Matter SDK wheels version. NOTE: The API might not be
       compatible with the Python Matter server.
+  primary_interface:
+    name: Primary network interface
+    description: >-
+      Override the default system primary interface. This should be the interface
+      used for provisioning new Matter devices.
 network:
   5580/tcp: Matter Server WebSocket server port.


agners:
Hm, if you specify `--primary-interface` in matter_server_args, it really should specify the argument twice, and by default in Python argument parsing (as the [Python Matter Server is using](https://github.com/home-assistant-libs/python-matter-server/blob/main/matter_server/server/__main__.py#L90)), the last one wins. And since `matter_server_args` get appended at the end of the args list, using matter should actually work. :thinking: 

Did you use separate fields for the argument (`--primary-interface`) and the value?

Gunni:
Regardless of whether that works or not, having the option spelled out like this makes it much easier to discover rather than a wormhole of googling and trying to figure out how to isolate matter/thread devices from the internet.

agners:
> Regardless of whether that works or not, having the option spelled out like this makes it much easier to discover rather than a wormhole of googling and trying to figure out how to isolate matter/thread devices from the internet.

The problem is that the primary interface argument is really somewhat misleading: It has very limited significance currently, as it only scopes link-local addresses which we receive from the Android App (see https://github.com/home-assistant/addons/pull/3468#issuecomment-1948257851, or more specifically https://github.com/home-assistant-libs/python-matter-server/pull/501).

Ultimately this argument likely will be removed again from the Matter Server, once we have a way to rediscover the Matter device commissioned through Android.

If discovering the configuration is a problem, we can promote the  `--primary-interface` argument in the Matter Server add-on `DOCS.md`, e.g. by using the [GitHub tip markdown syntax](https://github.com/orgs/community/discussions/16925) to highlight the feature, e.g.:

> [!TIP]
> If you are using Android to commission Matter devices, the Matter device needs to be on reachable through the network interface considered as primary interface by Home Assistant (check `ha network info` to see which device is the primary interface). If you are trying to commission a device on a different network than what is considered primary, specify `--primary-interface` and the interface name in the "Extra Matter Server arguments" configuration option.

That said, I just double checked again, specifying the argument in the extra argument list seems to work. I am closing this PR as adding a dedicated option is not the right approach here.

csuich2:
>That said, I just double checked again, specifying the argument in the extra argument list seems to work. I am closing this PR as adding a dedicated option is not the right approach here.

Can you elaborate on what you did to make it work? We have multiple reports from people (including myself) that adding the argument has no impact and that is verified by the log messages which are using the default interface rather than the specified interface.

agners:
> Can you elaborate on what you did to make it work? We have multiple reports from people (including myself) that adding the argument has no impact and that is verified by the log messages which are using the default interface rather than the specified interface.

I've simply added it to the Extra Matter Server arguments.

The follow log message is a red herring, this is a SDK detected primary interface which has no significance:

```
[1754569841.127987][9946:9946] CHIP:DL: Found the primary Ethernet interface:eno2
```

See: https://github.com/home-assistant/addons/pull/4092#issuecomment-3164029245

csuich2:
I see, you're saying that the log message is coming from the script which is just indicating what it autodetected, but the service itself is properly using the *second* primary interface argument which is the one specified by the user?

agners:
> I see, you're saying that the log message is coming from the script which is just indicating what it autodetected, but the service itself is properly using the _second_ primary interface argument which is the one specified by the user?

Exactly, the Python Matter Server code itself is using the last `--primary-interface` argument.

I've opened a PR to print the primary interface name as known by the Matter Server (see https://github.com/home-assistant-libs/python-matter-server/pull/1212). With that it should be easier to verify that the argument indeed is processed correctly.

Gunni:
Does changing primary interface do anything to prevent Thread devices from accessing the internet?

agners:
> Does changing primary interface do anything to prevent Thread devices from accessing the internet?

No. Currently, the primary interface only has influence during commissioning with the Android Companion app.

csuich2:
>No. Currently, the primary interface only has influence during commissioning with the Android Companion app.

I think that's actually what I'm after. When I try to add a Matter device on another vlan it connects to wifi just fine but then fails to be added to Home Assistant. I tried adding another network interface to Home Assistant on that vlan but that didn't work which is how I ended up on this issue.

agners:
> When I try to add a Matter device on another vlan it connects to wifi just fine but then fails to be added to Home Assistant. 

... using Android? Yes, then that is what primary interface is helping to solve. You should be able to fix it by adding `--primary-interface` and `<iface>` to the extra argument list (as separate things, IIRC).

elestril:
Sorry for letting this go stale. 

To wrap this up:
```
matter_server_args:
  - "--primary-interface=end0.2"
```
does achieve the same thing, and using the `=` syntax avoids awkwardly splitting in two separate args. 

Log from mainline version, using this config (relevant lines): 
```
[11:08:30] INFO: Starting Matter Server...
[..]
[11:08:33] INFO: Using 'end0' as primary network interface.
[..]
2025-09-01 11:08:43.170 (MainThread) INFO [matter_server.server.server] Matter Server initialized
2025-09-01 11:08:43.170 (MainThread) INFO [matter_server.server.server] Using 'end0.2' as primary interface (for link-local addresses)
2025-09-01 11:08:43.173 (MainThread) INFO [matter_server.server.server] Starting the Matter Server...
[..]
2025-09-01 11:08:43.886 (MainThread) INFO [matter_server.server.server] Matter Server successfully initialized.
2025-09-01 11:10:04.004 (MainThread) INFO [matter_server.server.device_controller] Starting Matter commissioning using Node ID 1 and IP fe80::42ae:30ff:fea5:db64%end0.2.
2025-09-01 11:10:05.873 (Dummy-2) INFO [chip.ChipDeviceCtrl] Established secure session with Device
2025-09-01 11:10:11.373 (Dummy-2) INFO [chip.ChipDeviceCtrl] Commissioning complete
2025-09-01 11:10:11.374 (MainThread) INFO [matter_server.server.device_controller] Matter commissioning of Node ID 1 successful.
```
So the `11:08:33` log is indeed a red herring. 

I would still argue that `primary_interface` should be a first class documented config option, not hidden in extra args. But it does work without it.

snovotill:
This problem continues to manifest in Incus OCI host-networking deployments.
https://github.com/home-assistant/core/issues/152848

PR #4096, opened 2025-07-30
ssh: Disable keyboard interactive authentication method if key used

HWXLR8:
There is existing logic in place to disable password authentication for SSH if a key-pair is being used, but keyboard interactive mode is left enabled:

```
debug1: Authentications that can continue: publickey,keyboard-interactive
```
this change disables keyboard interactive mode only allowing public keys to be used if provided.

## Summary by CodeRabbit

* **Security Improvements**
  * Keyboard-interactive authentication is disabled when authorized keys are used to strengthen SSH login security.

* **Documentation**
  * Added a changelog entry documenting this change under version 9.19.0.

* **Chores**
  * Bumped the SSH addon version to 9.19.0 to reflect the update.

### DIFF
--- a/ssh/CHANGELOG.md
+++ b/ssh/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 9.19.0
 
-Disable keyboard interactive authentication method if keys are used
+- Disable keyboard interactive authentication method if keys are used
 
 ## 9.18.0
 


agners:
> this change disables keyboard interactive mode only allowing public keys to be used if provided.

Is keyboard interactive mode left enabled causing a problem in your case? From this [ServerFault thread](https://serverfault.com/questions/189643/sshd-blocking-password-authentication) it seems that there are other authentication method (besides password authentication) which potentially use interactive mode. But then again, it is not possible to use this currently, from what I can tell :shrug: .

HWXLR8:
I wouldn't say it is causing a problem, and the thread is correct that password authentication is just one type of keyboard-interactive authentication. Based on the fact that there is logic to disable password auth if keys are present, combined with the fact that there is no implementation of non-password keyboard interactive modes present leads me to believe that this should be disabled to keep the security policy consistent. Open to thoughts and suggestions of course.

agners:
> Based on the fact that there is logic to disable password auth if keys are present, combined with the fact that there is no implementation of non-password keyboard interactive modes present leads me to believe that this should be disabled to keep the security policy consistent. Open to thoughts and suggestions of course.

Ok, yeah fine by me :+1:

ssh/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.19.0
+
+Disable keyboard interactive authentication method if keys are used
```
agners:
```suggestion
- Disable keyboard interactive authentication method if keys are used
```

PR #4101, opened 2025-08-09
Allow to specify ezsp-baudrate in silabs_flasher

maxim-smirnov:
There are some [firmwares](https://github.com/darkxst/silabs-firmware-builder/blob/main/firmware_builds/zbdonglee/ncp-uart-hw-v7.4.5.0-zbdonglee-230400.gbl) with not standard (115200) baudrate for ezsp application. 
Added config variable to specify that.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a custom EZSP baudrate for the universal-silabs-flasher.

* **Documentation**
  * Docs and changelog updated to show separate baudrate options for bootloader and EZSP and how to use them.

* **Chores**
  * Configuration schema and translations updated to include the new EZSP baudrate option; package version bumped to 0.3.3.

### DIFF
--- a/silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
+++ b/silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
@@ -11,6 +11,7 @@ set -e
 declare autoflash_firmware
 declare device
 declare bootloader_baudrate
+declare ezsp_baudrate
 declare firmware
 declare verbose
 declare usb_device_path
@@ -55,6 +56,7 @@ function is_home_assistant_yellow {
 
 device=$(bashio::config 'device')
 bootloader_baudrate=$(bashio::config 'bootloader_baudrate')
+ezsp_baudrate=$(bashio::config 'ezsp_baudrate')
 
 if is_home_assistant_yellow; then
     bashio::log.info "Detected Home Assistant Yellow"
@@ -109,12 +111,13 @@ if bashio::config.true 'verbose'; then
     verbose="-v"
 fi
 
-bashio::log.info "Starting universal-silabs-flasher with ${device} (bootloader baudrate ${bootloader_baudrate})"
+bashio::log.info "Starting universal-silabs-flasher with ${device} (bootloader baudrate ${bootloader_baudrate}) (ezsp baudrate ${ezsp_baudrate})"
 # shellcheck disable=SC2086
 universal-silabs-flasher \
     ${verbose} \
     --device ${device} \
     --bootloader-baudrate "${bootloader_baudrate}" \
+    --ezsp-baudrate "${ezsp_baudrate}" \
     ${gpio_reset_flag} \
     flash --force --firmware "/root/${firmware}"
 
--- a/silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
+++ b/silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
@@ -11,7 +11,6 @@ set -e
 declare autoflash_firmware
 declare device
 declare bootloader_baudrate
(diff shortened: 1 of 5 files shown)


agners:
Hm, I wonder if it maybe would be a better idea to add more baudrates to the list of default baudrates of the universal-silabs-flasher :thinking: 

https://github.com/NabuCasa/universal-silabs-flasher/blob/v0.0.31/universal_silabs_flasher/const.py#L48

@puddly thoughts?

puddly:
@agners I don't think I've seen 230400 commonly used so I think an explicit override would be best. Each new combination slows things down by 5+ seconds.

An alternative to explore would be for the flasher package to auto-detect the bootloader trigger for common devices.

silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
```diff
@@ -109,12 +111,13 @@ if bashio::config.true 'verbose'; then
...
 fi
 
-bashio::log.info "Starting universal-silabs-flasher with ${device} (bootloader baudrate ${bootloader_baudrate})"
+bashio::log.info "Starting universal-silabs-flasher with ${device} (bootloader baudrate ${bootloader_baudrate}) (ezsp baudrate ${ezsp_baudrate})"
 # shellcheck disable=SC2086
 universal-silabs-flasher \
     ${verbose} \
     --device ${device} \
     --bootloader-baudrate "${bootloader_baudrate}" \
+    --ezsp-baudrate "${ezsp_baudrate}" \
```
puddly: We should omit this option if it isn't provided, as the EZSP baudrate is chosen by the flasher right now. This would hard-code it to the default of 115200.

maxim-smirnov: Ok, should I make `ezsp_baudrate` without the default value then? And specify it as optional `int` in config?

agners: Yes, make it optional without default value and use `bashio::config.exists` to check if the config has been set by the user.

maxim-smirnov:
Ok, made it `null` by default and checking later that it `has_value`. According to this compiling `ezsp_baudrate_arg`.

silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
```diff
@@ -109,12 +111,18 @@ if bashio::config.true 'verbose'; then
     verbose="-v"
 fi
 
+ezsp_baudrate_arg=""
+if bashio::config.has_value 'ezsp_baudrate'; then
+    ezsp_baudrate_arg="--ezsp-baudrate \"${ezsp_baudrate}\""
```
agners: I'd recommend using arrays here, like we have it in other places (e.g. the Matter Server add-on):

```sh
ezsp_baudrate_arg=()
if bashio::config.has_value 'ezsp_baudrate'; then
    ezsp_baudrate_arg("--ezsp-baudrate", ${ezsp_baudrate})
```


```sh
universal-silabs-flasher \
...
    "${ezsp_baudrate_arg[@]}"
...
```

silabs_flasher/rootfs/etc/s6-overlay/scripts/universal-silabs-flasher-up
```diff
@@ -55,6 +56,7 @@ function is_home_assistant_yellow {
 
 device=$(bashio::config 'device')
 bootloader_baudrate=$(bashio::config 'bootloader_baudrate')
+ezsp_baudrate=$(bashio::config 'ezsp_baudrate')
```
agners: This should no longer be here.

PR #4104, opened 2025-08-11
Fix DuckDNS Single TXT record limitations

Xebozone:
**run:**
- Call Dehydrator per-domain or alias due to limitation in DuckDNS which can only handle a single TXT record at a time

**hooks:**
- Output text to show that domain or alias is being processed
- Remove filtering of the domain = alias, otherwise, aliases are not getting renewed
- DuckDNS NameServer TTL is observed at 60s, so, we need to wait longer than that

### DIFF
--- a/duckdns/config.yaml
+++ b/duckdns/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.20.0
+version: 1.21.0
 slug: duckdns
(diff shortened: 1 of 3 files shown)


Xebozone:
Made a mistake, closing. Will open a new one

PR #4108, opened 2025-08-11
duckDns: Fix Bash Logging Error

Xebozone:
- Fix Bash IO logging issue
- Version Bump
- Update Changelog


Xebozone:
Thanks! My domains are all renewing now in prod 🥳

PR #4122, opened 2025-08-15
duckDns: Revert renew each domain first

Xebozone:
- Revert behavior for renewing each domain separately
- Dehydrated needs some fixes to support this first

## Summary by CodeRabbit

* Bug Fixes
  * Streamlined certificate renewal into a single bulk run, reducing redundant per-domain executions and improving stability.
  * Adjusted renewal behavior to align more closely with earlier versions while retaining wildcard support.
  * Reduced unnecessary log output during renewal.

* Documentation
  * Added 1.26.0 changelog entry summarizing renewal updates and improvements.

* Chores
  * Bumped version to 1.26.0.

### DIFF
--- a/duckdns/rootfs/etc/s6-overlay/s6-rc.d/duckdns/run
+++ b/duckdns/rootfs/etc/s6-overlay/s6-rc.d/duckdns/run
@@ -51,12 +51,16 @@ function le_renew() {
     for domain in "${domainsAndAliases[@]}"; do
 		domain_args+=("--domain" "${domain}")
 		
+		# Note, the below is commented out for now as dehydrated force-renews each domain when domain input is different from the previous run
+		# This is causing rate-limiting from Let's Encrypt
+		# Need to resolve issue with dehydrated first to properly support single TXT records
+		
 		# DuckDNS does not support more than a single TXT record, so, process each domain or alias separately and combine later
-		dehydrated --cron --algo "${ALGO}" --hook /root/hooks.sh --challenge dns-01 --domain "${domain}" --out "${CERT_DIR}" --config "${WORK_DIR}/config" || true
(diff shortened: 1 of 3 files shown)


Xebozone:
I'm reverting the one change I made earlier that renews each domain first due to the DuckDNS limitation of single TXT record.
Dehydrated is force-renewing the domains each run which can result in Let's Encrypt rate limits for people with multiple domains.
Dehydrated will need to provide better support for single TXT records. This is something to be tackled before enabling this again.

duckdns/rootfs/etc/s6-overlay/s6-rc.d/duckdns/run
```diff
@@ -51,12 +51,16 @@ function le_renew() {
...
+		# Need to resolve issue with dehydrated first to properly support single TXT records
+		
 		# DuckDNS does not support more than a single TXT record, so, process each domain or alias separately and combine later
-		dehydrated --cron --algo "${ALGO}" --hook /root/hooks.sh --challenge dns-01 --domain "${domain}" --out "${CERT_DIR}" --config "${WORK_DIR}/config" || true
+		#dehydrated --cron --algo "${ALGO}" --hook /root/hooks.sh --challenge dns-01 --domain "${domain}" --out "${CERT_DIR}" --config "${WORK_DIR}/config" || true
     done
 	
 	# This final stage will combine everything into a single cert with SANs for the aliases
-	bashio::log.info "Final pass to combine individual certificates"
+	#bashio::log.info "Final pass to combine individual certificates"
```
agners, line 63: Why disabling the massage? Did you mean to disable the command below too?

Xebozone: There's only one pass now, so the message is redundant. The line below is still needed.
I'm working on a fork of dehydrated but for now this needs to happen

Xebozone:
Closing. Replaced with a proper fix: https://github.com/home-assistant/addons/pull/4126

PR #4126, opened 2025-08-16
Proper Support for Single-TXT Domains Fix Let's Encrypt Rate-Limit Resolution

Xebozone:
- Use Xebozone fork of Dehydrated with support for Single-TXT domains (such as DuckDNS)
- Revise run to use the new parameters for dehydrated
- Version bump
- Updated Changelogs

### DIFF
--- a/duckdns/Dockerfile
+++ b/duckdns/Dockerfile
@@ -5,7 +5,7 @@ FROM $BUILD_FROM
 ARG DEHYDRATED_VERSION
 RUN apk add --no-cache openssl \
   && curl -s -o /usr/bin/dehydrated \
-    "https://raw.githubusercontent.com/lukas2511/dehydrated/v${DEHYDRATED_VERSION}/dehydrated" \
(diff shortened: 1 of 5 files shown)


duckdns/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM $BUILD_FROM
 ARG DEHYDRATED_VERSION
 RUN apk add --no-cache openssl \
   && curl -s -o /usr/bin/dehydrated \
-    "https://raw.githubusercontent.com/lukas2511/dehydrated/v${DEHYDRATED_VERSION}/dehydrated" \
+    "https://raw.githubusercontent.com/Xebozone/dehydrated/${DEHYDRATED_VERSION}/dehydrated" \
```
agners, line 8: Hm, I see this now relies on the forked verison of dehydrated, but you already made a PR upstream (https://github.com/dehydrated-io/dehydrated/pull/977) :+1: .

Xebozone: That's right!
When it gets approved, I'll update this addon with the official release 😊

PR #4164, opened 2025-09-14
Start nginx_proxy as a service

zeehio:
Before this PR, home assistant starts before the nginx proxy. On my system the add-on takes a while to start.

I am not sure when all addons with "startup: application" are started, is it maybe after all integrations have started to load or maybe later?

Anyway, after this PR, the nginx_proxy addon starts first, and waits for the frontend to be ready (We wait to prevent 504 errors from nginx).

This makes the frontend available earlier to the user, at least on my system and my experience.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.13.1
+
+- Start the addon earlier so HA is accessible earlier.
+
 ## 3.13.0
 
 - Update Alpine Linux to 3.22 (nginx 1.28.x)
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 3.13.0
+version: 3.13.1
 hassio_api: true
 slug: nginx_proxy
 name: NGINX Home Assistant SSL proxy
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -13,6 +13,7 @@ arch:
   - i386
 image: homeassistant/{arch}-addon-nginx_proxy
 init: false
+startup: services
 map:
   - ssl
   - share
--- a/nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/nginx/run
+++ b/nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/nginx/run
@@ -51,6 +51,19 @@ if bashio::config.true 'cloudflare'; then
     fi
 fi
 
+bashio::log.info "Waiting for core to be ready..."
+port=$(bashio::core.port)
+until curl \
+   --output /dev/null \
+   --silent \
+   --fail \
+   "http://homeassistant.local.hass.io:$port";
+do
+    printf '.'
+    sleep 0.1
+done
+
+
 # start server
 bashio::log.info "Running nginx..."
 stat "/ssl/$(bashio::config 'certfile')" -c %y > /tmp/certificate_timestamp


zeehio:
The changelog suggestion from @coderabbit was good, I updated it.

agners:
> Before this PR, home assistant starts before the nginx proxy. On my system the add-on takes a while to start.

With "the add-on" you mean the NGINX proxy add-on? It takes 3s on my x86-64 system, a bit surprising that it takes long.

> I am not sure when all addons with "startup: application" are started, is it maybe after all integrations have started to load or maybe later?

Application add-ons are started once:
  - The Home Assistant Core container is running
  - The Home Assistant API is responding
  - Home Assistant's core state is "RUNNING"

So depending on what Home Assistant is exactly doing (e.g. migrating database on upgrade) it may take a bit longer indeed.

> Anyway, after this PR, the nginx_proxy addon starts first, and waits for the frontend to be ready (We wait to prevent 504 errors from nginx).

In a way this adds another layer of polling Home Assistant for it to be ready. The Supervisor is already doing that. It probably doesn't help if we have more services which poll HA for it to be ready.

> This makes the frontend available earlier to the user, at least on my system and my experience.

Ok, I see what you are after here. I guess the question becomes why exactly NGINX is started late on your system... :thinking:  Is it maybe just the amount and/or order of `application` type add-ons? Supervisor starts all add-ons at once, and then waits for them to have state started.

zeehio:
> > Before this PR, home assistant starts before the nginx proxy. On my system the add-on takes a while to start.
> 
> With "the add-on" you mean the NGINX proxy add-on? It takes 3s on my x86-64 system, a bit surprising that it takes long.
> 

My explanation wasn't clear, your guess was good. The add-on itself started fast, this PR intends to make it start earlier. 


> 
> > This makes the frontend available earlier to the user, at least on my system and my experience.
> 
> Ok, I see what you are after here. I guess the question becomes why exactly NGINX is started late on your system... 🤔 Is it maybe just the amount and/or order of `application` type add-ons? Supervisor starts all add-ons at once, and then waits for them to have state started.

My guess is that the late start of this add-on might have been caused by a buggy home assistant custom integration that was taking a long time to start, but I am not even sure if this is a possible cause for the delay I was seeing or not.

Anyway, I switched back to the current nginx proxy stable add-on (without this PR) and I am not noticing a significantly larger delay as I was noticing it before, so it's hard for me to investigate this further.

If you believe that (a) a buggy custom integration can cause delays in the startup of this add-on, given how the startup process works and (b) that this PR is worth to merge because it helps mitigate that problem, then I will be happy to keep working to get this merged.

If you do not agree with both points, I think it is best for us to close this PR unmerged. If it eventually becomes a problem for someone else and it gets reported we can always come back here and rescue this.

Sorry for wasting your time if that's the case. The startup process of home assistant is scattered among the supervisor, the OS and the core and it's hard for me to understand the whole picture. This PR did help me in the past enough to submit it here, but I do not find it effective anymore, so I can't tell its value right now, unfortunately.

Thank you for your time anyway!

agners:
> If you believe that (a) a buggy custom integration can cause delays in the startup of this add-on, given how the startup process works and (b) that this PR is worth to merge because it helps mitigate that problem, then I will be happy to keep working to get this merged.
 
Buggy integration can delay startup of Home Assistant, so this was likely a culprit.

With that in mind, I'd rather prefer not merge this PR, at least not as-is. We've previously discussed to make NGINX a native feature, essentially promote it from add-on state to a plug-in, and ship it by default. With that, we could also include special handling to make sure it gets started "at the right time". 

> If you do not agree with both points, I think it is best for us to close this PR unmerged. If it eventually becomes a problem for someone else and it gets reported we can always come back here and rescue this.
> 
> Sorry for wasting your time if that's the case. The startup process of home assistant is scattered among the supervisor, the OS and the core and it's hard for me to understand the whole picture. This PR did help me in the past enough to submit it here, but I do not find it effective anymore, so I can't tell its value right now, unfortunately.

No worries, thanks for your effort!

PR #4170, opened 2025-09-19
Update to Z-Wave JS 15.14.0

balloob:
Update Z-Wave JS to 15.14.0

https://github.com/zwave-js/zwave-js/releases/tag/v15.14.0

## Summary by CodeRabbit

* **New Features**
  * Support configuring a socket as an alternative to a physical device; added UI text for the socket option.

* **Bug Fixes**
  * More robust NVM conversion when encountering unknown sections.

* **Chores**
  * Bumped Z‑Wave JS dependency to 15.14.0 and released add-on v0.24.0 with top-level changelog.
  * Configuration schema updated: added socket option, made device optional, and startup now requires either device or socket.

### DIFF
--- a/zwave_js/config.yaml
+++ b/zwave_js/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 0.23.0
+version: 0.24.0
 slug: zwave_js
 name: Z-Wave JS
 description: Control a Z-Wave network with Home Assistant Z-Wave JS
--- a/zwave_js/config.yaml
+++ b/zwave_js/config.yaml
@@ -36,7 +36,7 @@ options:
 ports:
   3000/tcp: null
 schema:
-  device: device(subsystem=tty)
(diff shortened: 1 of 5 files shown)


zwave_js/config.yaml
```diff
@@ -35,7 +35,7 @@ options:
 ports:
   3000/tcp: null
 schema:
-  device: device(subsystem=tty)
+  device: str?
```
frenck: This change will remove the serial port selection from the UI and will thus rely on the end-user to come up with the right paths.

I understand why we do this, just want to make sure this is the wished end result and that you are aware of that repercussion.

agners: There is also the translation/display of by-id paths instead of `/dev/ttyUSB0` for tty devices. The by-id paths are quite annoying to type :cry:

frenck (approved):

PR #4188, opened 2025-10-06
Drop Codenotary configuration

agners:
The builder no longer creates Codenotary signature and Supervisor no longer checks Codenotary signatures. This information has become obsolete, so remove it from all add-ons.


mattreim:
@agners Shouldn't they remove Google Assistant?

agners:
> @agners Shouldn't they remove Google Assistant?

Right, something when wrong after removing the add-on :see_no_evil:  Thanks for bringing this up!

mattreim:
@agners Can these entries also be removed from the `config.yaml`?

`codenotary: notary@home-assistant.io`

agners:
> @agners Can these entries also be removed from the `config.yaml`?
> 
> `codenotary: notary@home-assistant.io`

Yes, they can be removed in all add-ons. Good catch, I'll create a PR for this.

PR #4189, opened 2025-10-06
Remove deprecated Google Assistant add-on

agners:
The Google Assistant add-on has been deprecated since #3870. It no longer builds which is a problem if PRs touch build files.

Even though it has many installations, it is unlikely to be used since the Authentication mechanism is no longer working.

Remove it from the repository.

## Summary by CodeRabbit

* **Chores**
  * Removed the deprecated Google Assistant add-on and all user-facing runtime components, web UI, assets, styling, translations, and deployment/build metadata — the add-on is no longer available or buildable.
* **Documentation**
  * Deleted all Google Assistant add-on documentation and changelog entries; README entry for the add-on was removed.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -66,5 +66,3 @@ jobs:
             universal-silabs-flasher-up
             unlock-tables-for-backup
             vlc-discovery
-          ignore_paths: >-
-            google_assistant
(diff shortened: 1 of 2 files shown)


agners:
FWIW, today there are 3825 installations still. Affected systems will have a repair issue saying "Add-on Google Assistant SDK has been removed from the repository it was installed from." etc. this is expected and removal should not cause any issue for users since the add-on is non-functional today.

image

PR #4196, opened 2025-10-16
[openthread_border_router] Migrate last active datasets when adapter changes

puddly:
This PR adds a new Python script to automatically migrate the last used dataset on OTBR startup by reading the adapter hardware address and manually creating a partial OTBR `.data` file containing just the active and pending Thread datasets.

This lets you just change the adapter serial port and have OTBR magically start working again, as you'd expect.

OTBR keeps track of all data by storing them in files unique the adapter. We compute the last-used dataset by looking at when these files were last modified and then deal with mis-matches appropriately:

1. If the last-used dataset is for an adapter different from the one currently configured, we extract its datasets and create a partial dataset file for OTBR to start with. This is basically the auto migration.
2. If you migrate _back_, the last-used `.data` file is not the file specific for your adapter. In this case, we back up the old `.data` file for your adapter and proceed with the above migration path.
3. All other cases just exit immediately.

## Summary by CodeRabbit

* **New Features**
  * Adds support for universal-silabs-flasher in the build to enable device flashing.
  * Automatically migrate the OTBR active/pending dataset and child info to a newly selected serial adapter on startup.

* **Documentation**
  * Bumped release version to 2.15.0 and added changelog entry describing the automatic migration behavior.

### DIFF
--- a/openthread_border_router/rootfs/usr/local/bin/migrate_otbr_settings.py
+++ b/openthread_border_router/rootfs/usr/local/bin/migrate_otbr_settings.py
@@ -45,7 +45,7 @@ def parse_otbr_settings(data: bytes) -> list[tuple[OtbrSettingsKey, bytes]]:
         value = data[4 : 4 + length]
         assert len(value) == length
 
-        settings.append(OtbrSettingsKey(key), value)
+        settings.append((OtbrSettingsKey(key), value))
         data = data[4 + length :]
 
     return settings


puddly:
@agners let me know what you think about this approach.

For background, we'd like to have automatic OTBR integration migration in Core. The two approaches I can think of are:
1. Have OTBR integration push the preferred dataset if it detects an OTBR addon with no dataset. This has the downside of silently losing your current dataset upon migration if it isn't marked as your preferred dataset.
2. Have the OTBR addon itself internally migrate. This is the approach I'm implementing here but it has a few corner cases. My original intent was to just patch OpenThread to write to a single `openthread.data` file (instead of a `0_{eui64}.data` file per-adapter but this is blocked by OpenThread's data format not actually including the adapter EUI64 in the data file itself, just in the filename.

PR #4208, opened 2025-10-28
Remove Codenotary signing information

agners:
Remove the signer username from all add-on configurations.

## Summary by CodeRabbit

* **Chores**
  * Removed deprecated notary metadata from several system addons (Configurator, deCONZ, DHCP Server, Mosquitto, Samba, SSH, Z-Wave JS) to streamline addon metadata and improve consistency across configurations.

### DIFF
--- a/configurator/config.yaml
+++ b/configurator/config.yaml
@@ -4,7 +4,6 @@ slug: configurator
 name: File editor
 description: Simple browser-based file editor for Home Assistant
 url: https://github.com/home-assistant/addons/tree/master/configurator
-codenotary: notary@home-assistant.io
 arch:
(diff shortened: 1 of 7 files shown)


agners:
There is no need to bump the add-on config since this is simply removing an already today unused option. With https://github.com/home-assistant/supervisor/pull/6236 add-ons with the config will receive a warning, so ideally this should be merged before that time. Supervisor does keep a copy of the config for installed add-ons, but there is no warnings for already installed add-ons (hence no need to bump the config which would make sure the copy of the config gets updated).

sairon (approved):

PR #4211, opened 2025-10-29
deCONZ: Add more debug options [1]

mattreim:
This pull request adds four(+one fix) more debug options. It has been tested and works as expected.

## Summary by CodeRabbit

* **Documentation**
  * Restructured and expanded debug documentation: clarified debug options are integer log levels, updated defaults and example configuration, and added new debug options.

* **Chores**
  * Bumped version to 8.5.1 and added a changelog entry for the release.
  * Introduced new debug flags and removed one legacy debug flag; reordered debug entries for consistency.
  * Minor startup and discovery log message formatting tweaks.

### DIFF
--- a/deconz/rootfs/etc/services.d/deconz/run
+++ b/deconz/rootfs/etc/services.d/deconz/run
@@ -51,8 +51,10 @@ bashio::log.info "deCONZ waiting for VNC to start"
 bashio::net.wait_for 5900
 
 # Delete unused key
-if bashio::config.exists 'dbg_otau'; then
-    bashio::addon.option 'dbg_otau'
+options=$(bashio::addon.options)
+if bashio::jq.exists "${options}" ".dbg_otau"; then
+    bashio::log.info "Removing dbg_otau"
+    bashio::addon.option "dbg_otau"
 fi
 
 # Load debug values


deconz/config.yaml
```diff
@@ -34,11 +34,15 @@ privileged:
   - SYS_RAWIO
 schema:
   device: device(subsystem=tty)
-  dbg_aps: int?
   dbg_info: int?
-  dbg_otau: int?
```
agners, line 39: If you remove options we should also delete them, see the Note in the [Options/schema documentation](https://developers.home-assistant.io/docs/add-ons/configuration#options--schema).

deconz/config.yaml
```diff
@@ -34,11 +34,15 @@ privileged:
...
+  dbg_aps: int?
   dbg_zcl: int?
   dbg_zdp: int?
+  dbg_ddf: int?
+  dbg_dev: int?
+  dbg_ota: int?
+  dbg_error: int?
+  dbg_http: int?
```
agners: Is there a level? Or should these be boolean maybe?

If so, ideally we'd migrate the existing ones.

deconz/DOCS.md
```diff
@@ -124,9 +124,13 @@ the add-on configuration:
 
 - `dbg_info`
 - `dbg_aps`
```
agners, line 126: I guess `dbg_aps` should to go here too.

deconz/config.yaml
mattreim: I only removed the "U" and adjusted the order to deCONZ.

deconz/config.yaml
mattreim: The values ​​can be 0 or 1, so "int" was already used and I haven't tested it with "bool". But I think it's predefined by deconz\run.

deconz/DOCS.md
mattreim: The DOCS also includes all changes.

deconz/config.yaml
agners: Well we could translate the bool and set a 0 or 1 for the environment variable. The bool options get rendered as a toggle in the user interface, which makes these settings more user friendly...

deconz/DOCS.md
agners: Ah right, we only reordered `dbg_aps`. Sorry for the noise.

deconz/config.yaml
agners: The "size" of the change does not matter: The option has a new name, so this will raise warnings in Supervisor logs. Simply add this to the start script:


```
if bashio::config.exists 'dbg_otau'; then
    # Delete unused key
    bashio::addon.option 'dbg_otau'
fi
```

deconz/config.yaml
mattreim: I like the idea, but I don't know how to do it right now.

agners: The current implementation allows to set any number, so the environment variable potentially end up with 2, 3...

mattreim: It would work with booleans, but once selected, they remain permanently visible, which wasn't the case before. Therefore, I would stick with the old behavior; users have been used to it for years.

Debug

deconz/config.yaml
mattreim: I've added it. Thanks.

deconz/config.yaml
mattreim: @agners I was just thinking about introducing a value limit, but it's not limited; it's just an error message. Would you prefer this?
```yaml
  dbg_info: int(0,2)?
  dbg_aps: int(0,2)?
  dbg_zcl: int(0,1)?
  dbg_zdp: int(0,1)?
  dbg_ddf: int(0,1)?
  dbg_dev: int(0,1)?
  dbg_ota: int(0,1)?
  dbg_error: int(0,2)?
  dbg_http: int(0,1)?
```
If the field is empty, the option is reset to zero and it is hidden again. It has always worked this way.

agners: You mean the value is (currently) only limited by the frontend, not actually enforced by Supervisor?

Is 2 a valid value for dbg_info, dbg_aps and dbg_error? I guess that makes the boolean actually not well suited.

mattreim: Currently, the values ​​are not limited at all, only by the user.

The number 2 refers to the second level (L2) in deCONZ. This screenshot is taken directly from the deCONZ GUI.

deCONZ-debug

mattreim: @agners If the values ​​need to be limited, I can extend the DOCS.md file.

- `dbg_info`  -> empty = 1 and option is off, but function is on [0,2]
- `dbg_aps`   -> empty = 0 and option is off [0,2]
- `dbg_zcl`   -> empty = 0 and option is off [0,1]
- `dbg_zdp`   -> empty = 0 and option is off [0,1]
- `dbg_ddf`   -> empty = 0 and option is off [0,1]
- `dbg_dev`   -> empty = 0 and option is off [0,1]
- `dbg_ota`   -> empty = 0 and option is off [0,1]
- `dbg_error` -> empty = 0 and option is off [0,2]
- `dbg_http`  -> empty = 0 and option is off [0,1]

These options require a number that represents the log level.

Example add-on config with `dbg_aps` enabled on log level L2:

```yaml
device: /dev/ttyUSB0
dbg_aps: 2

```

agners: I'd simply document that `dbg_info`'s default value is 1, and all other debug options is 0.

Even though it is not enforced by Supervisor, to me it seems sensible to use the int min/max schema for clarity.

mattreim: @agners Done, I think the PR is finished now.

deconz/rootfs/etc/services.d/deconz/run
```diff
@@ -50,17 +50,30 @@ tigervncserver \
 bashio::log.info "deCONZ waiting for VNC to start"
 bashio::net.wait_for 5900
 
+# Delete unused key
+if bashio::config.exists 'dbg_otau'; then
+    bashio::addon.option 'dbg_otau'
+fi
```
agners: Unfortunately, I realized that `bashio::config.exists` doesn't work after the `config.yaml` got updated. However, when reading all options the config is present. So the correct way to delete the unused key is this:

```suggestion
options=$(bashio::addon.options)
if bashio::jq.exists "${options}" ".dbg_otau"; then
    bashio::log.info "Removing dbg_otau"
    bashio::addon.option "dbg_otau"
fi
```

(I've just recently updated the documentation at https://developers.home-assistant.io/docs/add-ons/configuration#options--schema).

agners (changes requested):
Other than that, LGTM!

agners (approved):

PR #4216, opened 2025-11-05
Bump to wyoming-faster-whisper 3.0.1

synesthesiam:
- Bump wyoming-faster-whisper: https://github.com/rhasspy/wyoming-faster-whisper/compare/v2.5.0...v3.0.1
- Add support for `sherpa-onnx` and Nvidia's parakeet model
- Prefer parakeet model in `auto`
- Add support for GigaAM for Russian
- Add `stt_library` option to choose backend

## Summary by CodeRabbit

* **New Features**
  * Multiple speech-to-text backends supported (faster-whisper, sherpa-onnx/parakeet, transformers, GigaAM)
  * New stt_library option to choose backend; auto mode prefers parakeet where applicable
  * GigaAM adds Russian ASR support

* **Documentation**
  * Enhanced backup/restore guidance
  * Added UI text and translations for selecting speech-to-text backend

* **Chores**
  * Bumped package to version 3.0.1
  * Removed armhf, armv7, i386 badges/support

### DIFF
--- a/whisper/translations/en.yaml
+++ b/whisper/translations/en.yaml
@@ -38,6 +38,8 @@ configuration:
       The default is 'faster-whisper' which requires a CTranslate2-converted
       Whisper model. If set to 'transformers', a HuggingFace transformers-based
       Whisper model may be used.
+
+      Overrides `stt_library` when `custom_model` is set.
   initial_prompt:
     name: Initial prompt
     description: >-
@@ -50,6 +52,8 @@ configuration:
 
       The default `auto` will select the best backend/model based on
       language/hardware.
+
+      Overridden by `custom_model_type` when `custom_model` is set.
   debug_logging:
     name: Debug logging
     description: >-
--- a/whisper/DOCS.md
+++ b/whisper/DOCS.md
@@ -105,6 +105,8 @@ Speech-to-text backend library to use:
 - `sherpa` - force sherpa onnx backend (parakeet model only)
 - `transformers` - force HuggingFace transformers backend
 
+**Note**: When `custom_model` is set, then `custom_model_type` will override `stt_library`.
+
 ## Backups
 
 Whisper model files can be large, so they are automatically excluded from backups and re-downloaded on restore for remote models.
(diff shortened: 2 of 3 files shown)


whisper/translations/en.yaml
```diff
@@ -43,6 +43,13 @@ configuration:
     description: >-
       Description of audio that can help Whisper transcribe unusual words
       better.
+  stt_library:
+    name: Speech-to-text library
+    description: >-
+      Speech-to-text backend library to use.
+
+      The default `auto` will select the best backend/model based on
+      language/hardware.
```
agners, line 54: How does this interact with the `model` and `custom_model` configurations?

synesthesiam: The `custom_model_type` will determine the STT library (either `faster-whisper` or `transformers`). Maybe I should clarify here that the user should set `stt_library` to `auto` when using custom models?

agners: What happens if something other than `auto` is selected as `stt_library`? Does `custom_model`/`custom_model_type` becomes useless?

II guess what I am wondering is how does `custom_model_type`/`custom_model` and `stt_library` interact, as in do I need a certain `stt_library` to use a certain `custom_model_type`/`custom_model`? The interaction between `custom_model_type` and `custom_model` is well documented in `custom_model`.

synesthesiam: `custom_model_type` is used to select the correct `stt_library` when it's value is `auto`. Under the hood, only `stt_library` is actually passed as a command-line argument. I'll clarify more in the docs, and I think I'll just use `custom_model_type` to always override `stt_library` when `custom_model` is set.

whisper/README.md
```diff
@@ -2,12 +2,14 @@
 
 ![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield]
 
-Home Assistant add-on that uses [faster-whisper](https://github.com/guillaumekln/faster-whisper/) for speech-to-text.
+Home Assistant add-on that uses multiple speech-to-text backends:
```
agners, line 5: Given this, should we maybe replace the logo png's with something other then OpenAI? 🤔

synesthesiam: Whisper is still at the core of this add-on, though. I know it would be better if it was generically named "speech-to-text" or something, but we're kind of stuck with the whisper name for now :/

PR #4217, opened 2025-11-05
Bump to wyoming-piper 2.1.1

synesthesiam:
- Upgrade to `wyoming-piper` 2.1.1 - https://github.com/rhasspy/wyoming-piper/compare/v1.6.3...v2.1.1
- Streaming is enabled by default (remove `streaming` option)
- Remove `max_piper_procs` option (no longer applicable)
- Drop support for `armv7`
- Fix zeroconf discovery

## Summary by CodeRabbit

* **New Features**
  * Streaming is now enabled by default.
  * Zeroconf discovery improved.
  * Updated to version 2.1.1.

* **Bug Fixes**
  * Fixed zeroconf discovery issues.

* **Breaking Changes**
  * Dropped armv7 support.
  * Removed `max_piper_procs` and `streaming` configuration options.

### DIFF
--- a/piper/rootfs/etc/s6-overlay/s6-rc.d/piper/run
+++ b/piper/rootfs/etc/s6-overlay/s6-rc.d/piper/run
@@ -18,7 +18,7 @@ options=$(bashio::addon.options)
 for old_key in 'max_piper_procs' 'streaming'; do
     if bashio::jq.exists "${options}" ".${old_key}"; then
         bashio::log.info "Removing ${old_key}"
-        bashio::addon.option '${old_key}'
+        bashio::addon.option "${old_key}"
     fi
 done
 


piper/config.yaml
```diff
@@ -20,21 +20,17 @@ options:
   length_scale: 1.0
   noise_scale: 0.667
   noise_w: 0.333
-  max_piper_procs: 1
   debug_logging: false
   update_voices: true
-  streaming: true
```
agners, line 26: The Supervisor still has these removed options stored. To avoid warnings currently the recommended way is to delete the options explicitly in one of the add-on start scripts, see [the note in the options docs](https://developers.home-assistant.io/docs/add-ons/configuration#options--schema).

synesthesiam: I'm removing the old options now in the `run` script. The note in the docs said to guard with `bashio::config.exists` but that didn't work for me. The old keys didn't exist, but still caused a warning so I always call `bashio::addon.option` to delete them.

agners: Strange, definitly works here, tested with:
```
if bashio::config.exists 'autoflash_firmware'; then
    echo "Removing"                                                                                                        
    bashio::addon.option 'autoflash_firmware'
fi
```

Do you get an error? Or what happens?

It obviously only works once. Deleting every time leads a config save call to Supervisor every startup, which is certainly not ideal 😰

agners: Oh right, I see Supervisor filters the config according to schema, which essentially renders `bashio::config.exists` non working 🙈 

What does work is this:

```
key="test"
if bashio::jq.exists "${options}" ".${key}"; then
    bashio::log.info "Removing ${key}"
    bashio::addon.option '${key}'
fi
```

(note the dot before the key for the `jq.exists` call)

That said, I wonder if we should maybe make a breaking change in Supervisor and simply drop such orphan configs. The only downside would be that if an add-on accidentally removes a option, it can't simply readd it as system will have it removed 🤔

synesthesiam: I've updated it with the code you provided. I don't think it's a problem having to manually delete the old options, but we should at least fix the docs to use this new code.

synesthesiam: Make sure to fix your last line with double quotes instead of single: `bashio::addon.option "${key}"`

agners: Right! Thanks for the hint.

agners: PR to fix documentation: https://github.com/home-assistant/developers.home-assistant/pull/2860.

agners (approved):
LGTM, thanks!

PR #4234, opened 2025-11-19
nginx_proxy: Support HTTP/3 (QUIC)

kmhallen:
Add option to nginx_proxy for HTTP/3 (QUIC) support over UDP.

Try it out here with this custom addon repository:
https://github.com/kmhallen/ha-addon-nginx_proxy_quic

HTTP/3 should enable more seamless roaming from WiFi to cellular because the connection ID feature enables a connection to continue when the client IP address changes.

I couldn't find any reliable online HTTP/3 test websites, but browser developer consoles can be used to verify if HTTP/3 is being used. Chromium based browsers seem to be more likely to use HTTP/3 than Firefox. I'm not sure about the Android and iOS companion apps.

The port is specified both in options and ports because I couldn't find a way to read the port value directly. The port is needed to advertise HTTP/3 support in the `Alt-Svc` header.

## Summary by CodeRabbit

* **New Features**
  * Optional HTTP/3 (QUIC) support: UDP port mapping (default 443/udp), conditional enablement, and Alt‑Svc header exposure for HTTP/3.

* **Documentation**
  * Changelog bumped to 3.15.0; docs and translations updated with HTTP/3 configuration guidance, port/UDP details, and troubleshooting tips for firewall/Alt‑Svc issues.


agners:
Thank you for your PR, nice addition :star_struck: 

> The port is specified both in options and ports because I couldn't find a way to read the port value directly. The port is needed to advertise HTTP/3 support in the `Alt-Svc` header.

It should be possible to read the port using `quic_port="$(bashio::addon.port '443/udp')"`.

nginx_proxy/rootfs/etc/s6-overlay/s6-rc.d/nginx/run
```diff
@@ -7,10 +7,15 @@
...
+if [[ -z "$QUIC_PORT" ]]; then
+  QUIC_PORT=null
+fi
+
 bashio::log.info "Merging options & variables for template"
 # shellcheck disable=SC2046
-JSON_CONF=$(jq --arg port $(bashio::core.port) \
-    '({options: .}) + ({variables: {port: $port}})' \
+JSON_CONF=$(jq --argjson port $(bashio::core.port) --argjson quic_port $QUIC_PORT \
+    '({options: .}) + ({variables: {port: $port, quic_port: $quic_port}})' \
```
agners: With simple quotes the values are still consumed as numbers (since bash removes them). This eliminates the linter error:

```suggestion
JSON_CONF=$(jq --argjson port "$(bashio::core.port)" --argjson quic_port "$QUIC_PORT" \
    '({options: .}) + ({variables: {port: $port, quic_port: $quic_port}})' \
```

nginx_proxy/rootfs/etc/nginx/nginx.conf.gtpl
```diff
@@ -69,16 +74,30 @@ http {
         listen 443 ssl;
         listen [::]:443 ssl;
         http2 on;
+        {{- if .variables.quic_port }}
+        listen 443 quic reuseport;
```
agners: Is `reuseport` required here? We don't use it for the other instances. I'd expect a problem if that port is already used by some other process. 🤔

kmhallen: I don't know, but every example of NGINX and QUIC I found used the `reuseport` keyword.

DrJume:
This would be great! 
When using h3, it's recommended to set a SVCB (Service Binding) DNS record for h3 advertisement & port remapping. So when connecting to the server it directly uses h3 (without trying http/1.1 or h2).
When exposing HA over a different port than 443, SVCB records allow a subdomain to use the other port without needing to specify it in the URL 👍

kmhallen:
@agners Friendly ping :)

nginx_proxy/rootfs/etc/nginx/nginx.conf.gtpl
sairon: I wonder if this doesn't cause issues, from nginx docs:
> The listen directive can have several additional parameters specific to socket-related system calls. These parameters can be specified in any listen directive, but only once for a given address:port pair. 

I remember setting `ipv6only=off` in multiple server->listen directives caused errors for me. Generally I think a the practice is to set common parameters where the `default_server` is set.

kmhallen: The `reuseport` option is required. QUIC over UDP is handled differently than TCP. Should I move it to the `default_server` section?

https://www.f5.com/company/blog/nginx/binary-packages-for-preview-nginx-quic-http3-implementation
> The reuseport parameter is required for correct operation when there are multiple NGINX worker processes. It enables the kernel to distribute incoming HTTP/3 connections among them.

https://blog.nginx.org/blog/our-roadmap-quic-http-3-support-nginx
> We developed an eBPF program to take advantage of NGINX’s multi‑process architecture. NGINX uses multiple worker processes which enables highly scalable performance across multiple CPU cores. For maximum efficiency, we like to handle each connection with the same worker process. QUIC makes this challenging, as a connection is not tied to the client’s IP address, and the Linux kernel doesn’t provide UDP port-to-process affinity. To solve this we implemented an eBPF extension to integrate with SO_REUSEPORT so that we can map the QUIC connection ID to the worker that first handled it. This is neatly integrated into the NGINX core so that NGINX loads the eBPF bytecode into the kernel’s socket selection code.

kmhallen:
I tested that connection migration works by maintaining the same connection when IP/port changes using this:
https://github.com/stackia/quic-migration-tester

Normally `quic_bpf on;` would be necessary in the global scope to assign connections to the correct worker process for connection migration, but `worker_processes` is not set and defaults to 1.

sairon (approved):
Thanks for the quick changes, looks good now!

agners (approved):
Hm, I guess this:

> https://www.f5.com/company/blog/nginx/binary-packages-for-preview-nginx-quic-http3-implementation

> >   The reuseport parameter is required for correct operation when there are multiple NGINX worker processes. It enables the kernel to distribute incoming HTTP/3 connections among them.

Together with this:

> Normally `quic_bpf on;` would be necessary in the global scope to assign connections to the correct worker process for connection migration, but `worker_processes` is not set and defaults to 1.

Would make the `reuseport` not necessary in this particular case? 🤔

That said, probably safer to keep it, in case we adjust workers.

LGTM, thanks!

PR #4251, opened 2025-12-02
Drop build for deprecated archs, use latest native runners

sairon:
In #4248 breaking bump of the builder action was merged. This means that future attempts at builds of deprecated architectures will fail. This change removes the deprecated architectures from the build matrix and changes the build job to use native runner for aarch64 build, while updating to latest runner versions as well.

## Summary by CodeRabbit

* **Chores**
  * Internal infrastructure updates to build and deployment workflows.

---

**Note:** These changes are internal infrastructure updates with no direct impact on end-user functionality.

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   init:
-    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


agners (approved):
LGTM!

PR #4284, opened 2026-01-06
nginx_proxy: add use_ssl_backend option to support cases where backend is using SSL

krikk:
add use_ssl_backend option to support cases where the http section is using the ssl_certificate, ssl_key options

default to false, so will not affect existing installs, and will so work with default install

## Summary by CodeRabbit

* **New Features**
  * Added use_ssl_backend option (default: disabled) to allow the proxy to connect to backends over HTTPS when required.
  * Bumped release version to 4.1.0.

* **Documentation**
  * Updated setup instructions to reflect enabling use_ssl_backend for SSL backends.
  * Added user-facing label and description explaining the new option and its effect.


agners:
@krikk this needs rebasing meanwhile, sorry for the late review.

Do you have a particular use case in mind? I guess it could be seen as an alternative way of implementing the use case described in #4403, where Home Assistant is reached locally using regular TLS (no-client authentication) and connecting using TLS client authentication via NGINX Proxy from the Internet.

krikk:
the use case for this change is quite simple, home assistant supports using TLS/SSL for its web interface (yeah, we should use TLS/SSL even on local LAN), and this should also be supported by the reverse-proxy (nginx) we put in front of it, ran into this issue myself, so tried to fix it :)

0rax:
I just stumbled upon the issue this is trying to resolve when securing a new HomeAssistant deployment with all "local ports" (8123 for HA and 6052 for ESPHome) configured with TLS active.

The goal being to be able to have those ports secure while also be able to use domain based routing using the NGINX addons to each service. As krikk said, even on local LAN we should try to use TLS/SSL as much as possible when possible.

Thanks a lot for your work here !

PR #4289, opened 2026-01-11
Add smb encrypt option

7CqKcKvfAf:
Adds option to set "smb encrypt" to off/desired/required. If set to required, all traffic between client and server will be encrypted instead of plain text message traffic.
Encryption requires to use SMB3 and cannot be used in combination with compatibility mode option.

## Summary by CodeRabbit

* **New Features**
  * Added SMB encryption option with three settings — off, desired (default), required — to control encryption of client-server traffic; "required" enforces SMB3 and restricts guest access.

* **Documentation**
  * Added docs and translation strings for the encryption option and expanded guidance on server signing and compatibility mode behavior.

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.5.5
+
+- Add option for setting smb encrypted parameter
+
 ## 12.5.4
 
 - Fix invalid inverted commas in server signing parameter
--- a/samba/config.yaml
+++ b/samba/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 12.5.4
+version: 12.5.5
 slug: samba
 name: Samba share
 description: Expose Home Assistant folders with SMB/CIFS
@@ -37,6 +37,7 @@ options:
     - ssl
   compatibility_mode: false
   apple_compatibility_mode: true
+  encryption: "desired"
   server_signing: "default"
   veto_files:
     - ._*
@@ -60,6 +61,7 @@ schema:
     - "match(^(?i:(addons|addon_configs|backup|config|media|share|ssl))$)"
   compatibility_mode: bool
   apple_compatibility_mode: bool
+  encryption: list(off|desired|required)
   server_signing: list(default|auto|mandatory|disabled)
   veto_files:
     - str
--- a/samba/rootfs/usr/share/tempio/smb.gtpl
+++ b/samba/rootfs/usr/share/tempio/smb.gtpl
@@ -18,9 +18,19 @@
    interfaces = lo {{ .interfaces | join " " }}
    hosts allow = 127.0.0.1 {{ .allow_hosts | join " " }}
 
+   smb encrypt = {{ .encryption }}
+   {{ if eq .encryption "required" }}
+   client min protocol = SMB3
+   client max protocol = SMB3
+   server min protocol = SMB3
+   server max protocol = SMB3
+   server signing = auto
+   {{ else }}
+   server signing = {{ .server_signing }}
    {{ if .compatibility_mode }}
    client min protocol = NT1
    server min protocol = NT1
+   {{ end }}
    {{ end }}
 
    mangled names = no
@@ -30,8 +40,6 @@
    {{ if .apple_compatibility_mode }}
    vfs objects = catia fruit streams_xattr
    {{ end }}
-
-   server signing = {{ .server_signing }}
 
 {{ if (has "config" .enabled_shares) }}
 [config]
--- a/samba/rootfs/usr/share/tempio/smb.gtpl
+++ b/samba/rootfs/usr/share/tempio/smb.gtpl
@@ -24,7 +24,7 @@
    client max protocol = SMB3
    server min protocol = SMB3
    server max protocol = SMB3
-   server signing = auto
+   server signing = mandatory
    {{ else }}
    server signing = {{ .server_signing }}
    {{ if .compatibility_mode }}
--- a/samba/translations/en.yaml
+++ b/samba/translations/en.yaml
@@ -33,6 +33,12 @@ configuration:
       Enable Samba configurations to improve interoperability with Apple
       devices. May cause issues with file systems that do not support xattr
       such as exFAT.
+  encryption:
+    name: Encryption
+    description: >-
+      Configure SMB encryption.
+      SMB3 protocol is used if set to required.
+      Compatibility mode option will be ignored in this case.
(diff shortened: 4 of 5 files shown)


samba/config.yaml
```diff
@@ -60,6 +61,7 @@ schema:
     - "match(^(?i:(addons|addon_configs|backup|config|media|share|ssl))$)"
   compatibility_mode: bool
   apple_compatibility_mode: bool
+  encryption: list(off|desired|required)
```
agners, line 64: The documentation of Samba 4.18 (as shipped with the current app version) lists `if_required` as well (see [smb.conf man](https://www.samba.org/samba/docs/4.18/man-html/smb.conf.5.html)). Maybe it also makes sense to add `default` (see above).

samba/config.yaml
```diff
@@ -37,6 +37,7 @@ options:
     - ssl
   compatibility_mode: false
   apple_compatibility_mode: true
+  encryption: "desired"
```
agners, line 40: According to Samba 4.18 [smb.conf man](https://www.samba.org/samba/docs/4.18/man-html/smb.conf.5.html) the default is `if_required`. Is there a reason you choose `desired` here? I fear that this might cause existing setups to break.

Also, `smb.conf` also knows the `default` string, maybe this is the safest choice, as it relies on the Samba maintainers judgment what is the best default.

krikk: default == if_required: will only enable encrytion if the client requires it
desired: will always enable encryption, except in the case the client does not support it

so desired is better because it encrypts all possible connections, while still supporting clients which can't encrypt

...and this is a server service, which should be shipped with "secure" defaults, SMB 3 was introduced with Windows 10 / Server 2016, so every supported windows Client today will support SMB Encryption, if one need access with older clients, he can still use compability mode

agners: Do you know why the Samba developers did not choose this option by default?

krikk: the developer of this addon is setting up a smb-server, as such he is indirectly the admin of a samba server, and as admin he is responsible do setup the server with secure defaults, because the end-user who is installing this addon/app does not care/does not know how to do a secure setup...

running unencrypted services, even on LAN is NOT secure and should not be embraced by unsecure defaults, especially as there are practically no more clients which do not support smb encryption, even Windows 8 and Windows Server 2012 do support it, also all smbclients on linux support it

.../me is voting for secure defaults ...but this are just my 2 cents :)

agners: Yeah I get that you want to pick a good, secure default. I generally agree with that principle.

It just strikes me as odd that this is not the default in Samba :thinking:Makes me wonder if there are reasons  :shrug: .

I way more towards least surprise: This PR as it stands today *changes* the default for existing users. At the very least, this should be made clear in the changlog.

Also, no matter what default we are going with, at the very least let's add the old default `if_required`, and I'd also add `default` as an option as well.

7CqKcKvfAf: I believe we should provide the end user options which make it easy to understand what will be the achieved effect. This is why I decided to only expose `off`/`desired`/`required` as global options.

Concerning the `default` value: you can imagine I was a little bit surprised when I observed local network traffic between my linux client and HA server with wireshark and found out that I was able to ready everything in plaintext.

If we provide `default` as option, many people might just use it without reading smb.conf documentation, not knowing that encryption will not be enabled on many or most connections. Personally I often trust (or hope) that developers/publishers pick safe config defaults.

Furthermore I think it would be good if the user could supply his own smb.conf to the app without being forced to build a custom HA app. But that is a little bit offtopic.

samba/translations/en.yaml
```diff
@@ -33,6 +33,12 @@ configuration:
       Enable Samba configurations to improve interoperability with Apple
       devices. May cause issues with file systems that do not support xattr
       such as exFAT.
+  encryption:
+    name: Encryption
+    description: >-
+      Configure SMB encryption.
+      SMB3 protocol is used if set to required.
```
agners, line 40: This can be interpreted as if SMB3 is not used if this is not set to `required`. But that is not the case: "SMB protocol takes care of choosing the appropriate protocol." so SMB3 might be used even without this.

samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -18,10 +18,20 @@
    interfaces = lo {{ .interfaces | join " " }}
    hosts allow = 127.0.0.1 {{ .allow_hosts | join " " }}
 
+   smb encrypt = {{ .encryption }}
+   {{ if eq .encryption "required" }}
+   client min protocol = SMB3
+   client max protocol = SMB3
```
agners, line 24: Client features are not used by the add-on so this doesn't really do anything.

samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -18,10 +18,20 @@
    interfaces = lo {{ .interfaces | join " " }}
    hosts allow = 127.0.0.1 {{ .allow_hosts | join " " }}
 
+   smb encrypt = {{ .encryption }}
```
agners, line 21: To align with the other config uses:
```suggestion
   server smb encrypt = {{ .encryption }}
```

samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -18,10 +18,20 @@
...
+   {{ if eq .encryption "required" }}
+   client min protocol = SMB3
+   client max protocol = SMB3
+   server min protocol = SMB3
```
agners, line 25: Why is this needed? With `smb encrypt` From smb.conf man it sounds like this is anyways required when `smb encrypt` is set to `required`

samba/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.5.5
+
+- Add option for setting smb encrypted parameter
```
agners, line 5: This needs a note that the default changes.

samba/config.yaml
agners: > If we provide `default` as option, many people might just use it without reading smb.conf documentation, not knowing that encryption will not be enabled on many or most connections.

If that is the concern, then let's add a quick note in the option description: "Note: default will only encrypt if clients require encryption."

Not having today's default as a possible option is a no-go from my point of view. At the very least it is useful to regression test in case something doesn't work the way users expect after upgrading the Samba app to this new version.

agners (changes requested):

PR #4292, opened 2026-01-15
matter_server: use matter.js server when beta is enabled

Apollon77:
This PR proposes the addopn change to use and install the new matter.js based Matter server when the Beta flag is set.

Node.js is always installed to the container, but the matterjs-server only when beta is used and then also updated to latest on each start.

## Summary by CodeRabbit

* **New Features**
  * Matter.js-based server option available when Beta flag is enabled (v8.2.0+).
  * Automatic data migration when switching to the Beta server; users can switch back to the Python server at any time.
  * Node.js runtime added to the container to support the new server option.

* **Documentation**
  * Docs updated to describe server variants, Beta behavior, and clarified configuration field descriptions.

### DIFF
--- a/deconz/CHANGELOG.md
+++ b/deconz/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.5.3
+
+- Align ingress colors with Homo Assistant
+
 ## 8.5.2
 
 - Add configurable OTA update switches for Bosch, IKEA and OSRAM/LEDVANCE (default is off)
--- a/deconz/config.yaml
+++ b/deconz/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 8.5.2
+version: 8.5.3
 slug: deconz
 name: deCONZ
 description: >-
--- a/deconz/rootfs/usr/share/www/ingress.html
+++ b/deconz/rootfs/usr/share/www/ingress.html
@@ -8,7 +8,7 @@
     <style>
       :root {
         /* Home Assistant Light Theme */
-        --primary-color: #03a9f4;
+        --primary-color: #009AC7;
         --primary-text-color: rgba(0, 0, 0, 0.87);
         --secondary-text-color: rgba(0, 0, 0, 0.6);
         --disabled-text-color: rgba(0, 0, 0, 0.38);
@@ -17,15 +17,15 @@
         --card-background-color: #ffffff;
         --divider-color: rgba(0, 0, 0, 0.12);
 
-        --toolbar-background: #009AC7;
+        --toolbar-background: var(--primary-color);
         --toolbar-text-color: #fff;
 
-        --button-bg: #009AC7;
+        --button-bg: var(--primary-color);
         --button-text-color: #fff;
         --button-hover-color: #006787;
 
         --link-color: var(--primary-color);
-        --link-hover-color: #007aa3;
+        --link-hover-color: #006787;
 
         --focus-color: #989898;
 
@@ -35,7 +35,7 @@
       @media (prefers-color-scheme: dark) {
         :root {
           /* Home Assistant Dark Theme */
-          --primary-color: #03a9f4;
+          --primary-color: #009AC7;
           --primary-text-color: #e1e1e1;
           --secondary-text-color: #9c9c9c;
           --disabled-text-color: rgba(255, 255, 255, 0.3);
@@ -47,12 +47,12 @@
           --toolbar-background: #101E24;
           --toolbar-text-color: var(--primary-text-color);
 
-          --button-bg: #009AC7;
+          --button-bg: var(--primary-color);
           --button-text-color: #fff;
           --button-hover-color: #006787;
 
           --link-color: var(--primary-color);
-          --link-hover-color: #007aa3;
+          --link-hover-color: #006787;
 
           --focus-color: #989898;
 


matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -23,23 +24,39 @@ log_level=$(bashio::string.lower "$(bashio::config log_level info)")
...
+# Determine server type based on beta flag
+use_js_server=$(bashio::config.true "beta" && echo "true" || echo "false")
+
+if [[ "${use_js_server}" == "true" ]]; then
+  # Beta mode: Use JavaScript Matter Server
+  bashio::log.info "Using JavaScript Matter Server (beta mode)"
+  bashio::log.info "Installing latest matter-server from npm..."
+  # shellcheck disable=SC2164
+  cd /opt/matterjs
+  npm install matter-server@latest
```
marcelveldt: you don't want to pin to a specific version here ?

Apollon77: No, I guess for beta we always go latest ... When we switch later completely we can take over the version parameter we already have

Apollon77: Especially also because I expect to have fixes when people use it ... better to spread them to all possible to get better test coverage on current version.

matter_server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BUILD_FROM
+ARG BUILD_FROM=ghcr.io/matter-js/python-matter-server:8.1.2
```
MartinHjelmare: Why are we no longer using the argument in the build.yaml file?

matter_server/Dockerfile
```diff
@@ -69,6 +69,26 @@ RUN \
...
+RUN \
+    set -x \
+    && DPKG_ARCH=$(dpkg --print-architecture) \
+    && case "${DPKG_ARCH}" in \
+         amd64) NODE_ARCH="x64" ;; \
+         arm64) NODE_ARCH="arm64" ;; \
+         armhf) NODE_ARCH="armv7l" ;; \
+         *) NODE_ARCH="${DPKG_ARCH}" ;; \
+       esac \
+    && curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz" -o /tmp/node.tar.xz \
```
MartinHjelmare, line 83: Let's add a comment why we need to download this from nodejs.org instead of installing the OS package.

matter_server/Dockerfile
Apollon77: I can revert, but there was an error before ... let me try

matter_server/Dockerfile
Apollon77: added, reason is mainly that we want to have up to date versions of node.js and no repo has that

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
MartinHjelmare: I think we should make sure that users can report what version they are testing if we don't install a specific version, eg via diagnostics.

matter_server/translations/en.yaml
```diff
@@ -5,42 +5,46 @@ configuration:
...
+      Matter Server. Use `--help` to get a list of possible arguments. Note
+      that arguments are also added by the startup script controlled by other
+      add-on options.
   matter_server_version:
     name: Matter Server version
     description: >-
-      Install custom Matter Server version. WARNING: An older version might have
-      an incompatible storage format! Use this feature with caution! Make sure
-      you have a recent backup of the add-on!
+      Install custom Python Matter Server version (does not affect when you use
```
MartinHjelmare:
```suggestion
      Install custom Python Matter Server version (not applicable when using
```

matter_server/translations/en.yaml
```diff
@@ -5,42 +5,46 @@ configuration:
...
+      Beta flag). WARNING: An older version might have an incompatible storage
+      format! Use this feature with caution! Make sure you have a recent backup
+      of the add-on!
   matter_sdk_wheels_version:
     name: Matter SDK wheels version
     description: >-
-      Install custom Matter SDK wheels version. NOTE: The API might not be
-      compatible with the Python Matter server.
+      Install custom Python Matter SDK wheels version (does not affect when you
+      use Beta flag). NOTE: The API might not be compatible with the Python
```
MartinHjelmare:
```suggestion
      Install custom Python Matter SDK wheels version (not applicable when
      using Beta flag). NOTE: The API might not be compatible with the Python
```

MartinHjelmare:
Did you build and test the add-on with these changes locally?

matter_server/Dockerfile
```diff
@@ -69,6 +69,28 @@ RUN \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /usr/src/*
 
+# Install Node.js 22.x for JavaScript Matter Server (beta mode)
+# We download from nodejs.org instead of using OS packages to ensure
+# we have up-to-date Node.js versions with latest features and fixes
+ARG NODE_VERSION=22.21.1
```
sairon: Let's move this also to `build.yaml`. Also, `v22.22.0` is now latest in v22 LTS.

matter_server/Dockerfile
```diff
@@ -69,6 +69,28 @@ RUN \
...
+RUN \
+    set -x \
+    && DPKG_ARCH=$(dpkg --print-architecture) \
+    && case "${DPKG_ARCH}" in \
+         amd64) NODE_ARCH="x64" ;; \
+         arm64) NODE_ARCH="arm64" ;; \
+         armhf) NODE_ARCH="armv7l" ;; \
```
sairon: We no longer support armhf, so this is essentially a dead code:

```suggestion
```

matter_server/Dockerfile
```diff
@@ -69,6 +69,28 @@ RUN \
...
+# Install Node.js 22.x for JavaScript Matter Server (beta mode)
+# We download from nodejs.org instead of using OS packages to ensure
+# we have up-to-date Node.js versions with latest features and fixes
+ARG NODE_VERSION=22.21.1
+RUN \
+    set -x \
+    && DPKG_ARCH=$(dpkg --print-architecture) \
+    && case "${DPKG_ARCH}" in \
+         amd64) NODE_ARCH="x64" ;; \
+         arm64) NODE_ARCH="arm64" ;; \
```
sairon: These tags are equal, so we can leave it to the wildcard.
```suggestion
```

matter_server/translations/en.yaml
```diff
@@ -5,42 +5,46 @@ configuration:
...
+      better tested and allows to commission directly in proximity of the
+      device itself.
   matter_server_args:
     name: Extra Matter Server arguments
     description: >-
-      This allows to pass additional command line arguments to the Python Matter
-      Server. Use `--help` to get a list of possible arguments. Note that
-      arguments are also added by the startup script controlled by other add-on
-      options.
+      This allows to pass additional command line arguments to the Python
```
sairon: This applies to both implementations, so at least we shouldn't mention _Python_ here.

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -99,9 +116,10 @@ matter_server_args+=(
   ${extra_args[@]}
 )
 
-if bashio::config.true "beta"; then
-    exec /usr/bin/gdb --quiet -ex="set confirm off" -ex run -ex backtrace -ex "quit \$_exitcode" --args /usr/local/bin/python \
-         /usr/local/bin/matter-server "${matter_server_args[@]}"
+if [[ "${use_js_server}" == "true" ]]; then
+    # Start JavaScript Matter Server
+    exec node --enable-source-maps /opt/matterjs/node_modules/matter-server/dist/esm/MatterServer.js "${matter_server_args[@]}"
```
sairon, line 125: Is it okay that e.g. `--log-level-sdk` is passed also to the JS server? Even if so, at least to avoid clutter, we could create a common set and then append to the array based on what implementation is used.

matter_server/Dockerfile
```diff
@@ -69,6 +69,28 @@ RUN \
...
+         *) NODE_ARCH="${DPKG_ARCH}" ;; \
+       esac \
+    && curl -fsSL "https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-${NODE_ARCH}.tar.xz" -o /tmp/node.tar.xz \
+    && tar -xJf /tmp/node.tar.xz -C /usr/local --strip-components=1 --no-same-owner \
+    && rm -f /tmp/node.tar.xz \
+    && node --version \
+    && npm --version
+
+# Create directory for JavaScript Matter Server
+RUN mkdir -p /opt/matterjs
```
sairon: Let's avoid the unnecessary layer - either shove it into the previous `RUN` or (IMHO better for now) create it in runtime.

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -23,23 +24,39 @@ log_level=$(bashio::string.lower "$(bashio::config log_level info)")
...
-fi
-
-if bashio::config.has_value "matter_sdk_wheels_version"; then
-  chip_version=$(bashio::config 'matter_sdk_wheels_version')
-  bashio::log.info "Installing Matter SDK ${chip_version}"
-  pip3 install --pre --no-dependencies \
-    home-assistant-chip-clusters=="${chip_version}" \
-    home-assistant-chip-core=="${chip_version}"
+# Determine server type based on beta flag
+use_js_server=$(bashio::config.true "beta" && echo "true" || echo "false")
```
sairon: bashio will coerce it to string for you:
```suggestion
use_js_server=$(bashio::config "beta")
```

sairon (changes requested):
Couple of comments below, with one extra question - is the config supposed to be migrated seamlessly? Because if not, or if there are other concerns, we might want to put this version to `breaking_versions` in config (the main effect is that it prevents the add-on from auto-updating).

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
Apollon77: The version they test is at least seen in the Matter server web UI ... Or what do we need to do how to report that?

matter_server/Dockerfile
Apollon77: Yes but 22.22 is 2 days old ... and contains huge changes internally, I want to ensure to test that before using it. So 22.21.1 is the version from before I feel better for now

matter_server/translations/en.yaml
Apollon77: Right, I remove

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
Apollon77: Yes the JS server reads but ignores them. It might log a line about the existence, but will work

Apollon77:
@sairon @MartinHjelmare I will update the mentioned things tomorrow. For the diagnostic idea I need help because I do not know what I would need to do and how

Apollon77:
> is the config supposed to be migrated seamlessly? Because if not, or if there are other concerns, we might want to put this version to breaking_versions in config (the main effect is that it prevents the add-on from auto-updating).

We are currently testing this in the alpha. Yes the currently design is a "drip in replacement". We will load and parse the config on first start (thats why also increased time for the keepalive check). We also update the XXXXXX.json when nodes are added and deleted, so that users can switch back to the python variant any time.
All what was tested looks good ... lets still see the reality but ideally it is not breaking, it is migrating automatically

matter_server/Dockerfile
Apollon77: Done

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
TheJulianJES: If you download diagnostics in Home Assistant, both the Matter.js and Matter-server versions are already included, so I think this is good to go. For my install with this addon PR, it looks like this:
```json
"sdk_version": "matter-server/0.2.4 (matter.js/0.16.5)",
```

Apollon77: exxactly. The versions included are dynamically determined from the installed files. Ok then thats also tackled. Thanks @TheJulianJES

Apollon77:
@MartinHjelmare can you please also "kick" the home-assistant bot to recheck the cla, thanks

Apollon77:
@MartinHjelmare 
> Did you build and test the add-on with these changes locally?

No, not yet.
@TheJulianJES found https://developers.home-assistant.io/docs/add-ons/testing/ or maybe other tweaks ... need to understand that still

TheJulianJES:
Doing some final testing now. If anyone else wants to test it already, you can run this in your `amd64` HAOS instance:
```cli
docker pull thejulianjes/amd64-addon-matter-server:8.2.0
docker tag thejulianjes/amd64-addon-matter-server:8.2.0 homeassistant/amd64-addon-matter-server:8.1.2
```
Then restart the addon.

(8.1.2 so the addon will use that image. There won't be any indication that you're using the new addon and settings will still reference the Python variant, but if you enable the beta toggle, you'll see that NPM is installed and Matter.js starts up.)

Apollon77:
Setting back to draft to ensure a "planned release" :-)

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 8.2.0
+
+- Added Node.js to the container
+- Use the new [matter.js-based Matter-Server](https://github.com/matter-js/matterjs-server) when using "Beta"
+  - Data are automatically migrated without any necessary user action and allows switching back to Python anytime
```
TheJulianJES: "Data are automatically migrated" can be improved a bit. Example with other tweaks as well:
```suggestion
- Switch to the new [matter.js-based Matter Server](https://github.com/matter-js/matterjs-server) when enabling the **Beta** option
  - Existing data is migrated automatically with no user action required
  - Users can switch back to the Python-based Matter Server at any time
- Add Node.js to the add-on container
```

Would look like this then:

- Switch to the new [matter.js-based Matter Server](https://github.com/matter-js/matterjs-server) when enabling the **Beta** option
  - Existing data is migrated automatically with no user action required
  - Users can switch back to the Python-based Matter Server at any time
- Add Node.js to the add-on container

Apollon77:
@sairon @MartinHjelmare I just released a new test version and if feedback is good I would like to merge this tomorrow. Does that work for you?

agners:
I'll give this a try in the afternoon as well.

One question, instead of using the beta flag, maybe using a separate flag "Use matter.js server"? I think it would be a bit clearer and allows a more descriptive setting.

Apollon77:
Oi, we re-decided ... lets publish it now

TheJulianJES:
> instead of using the beta flag, maybe using a separate flag "Use matter.js server"?

We talked about that but ultimately landed on the decision to keep the existing Beta flag for this. I think the reasons were something along these lines:
- Users might not know the difference between the "Python Matter Server" and the "new Matter.js-based Matter server". For the user, "Beta" just means "New". I guess that's also why the Discord channel for this is called "new-matter-server" and not "matter-js-server".
  - The description for the beta flag already contains "Installs the latest beta of the Matter Server (matter.js based)", so I think for technical users, the change should be clear.
- It was also a smaller change and it'll give us a few more users who beta-test the new Matter.js Server after upgrading if they still have that Beta flag enabled. This is fine IMO and they can just disable the flag if they have any issues.

Apollon77:
And yes when we go to "full matter,js release" we could think having a new "revet back to python" flag for some time as a "pure worst case option", but lets discuss that then when we are happy with the Beta

PR #4298, opened 2026-01-19
feat(otbr): Add option to override backbone interface

t3hk0d3:
This PR adds option to override interface used for backbone network.

Currently it always uses primary interface, and that's really inconvenient with multi-VLAN setup.

For example i prefer to have my IoT devices to be isolated in a separate network with strict firewall policies enforced on router.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Backbone Network Interface option.

* **Documentation**
  * Added user-facing description and translation for the new backbone interface setting and its fallback behavior.

* **Bug Fixes / Reliability**
  * Improved interface resolution: use configured value if present, otherwise detect primary interface; removed implicit eth0 fallback and now show a clear error if no interface can be determined.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,38 +1,52 @@
 # Changelog
 
+## 2.16.5
+
+- Add `backbone_interface` option to override the network interface used for IPv6 routing.
+
 ## 2.16.4
 
 - Fix race condition during startup if web frontend is enabled
 
 ## 2.16.3
+
 - Ignore ephemeral temporary settings files in migration
 
 ## 2.16.2
+
 - Fix TREL being disabled by default in beta mode
 
 ## 2.16.1
+
 - Fix listen address of OTBR Web UI
 
 ## 2.16.0
+
 - Add beta toggle to switch between Thread 1.3 (stable) and Thread 1.4 (beta)
 - Beta mode uses OpenThread's built-in mDNS instead of mDNSResponder
 
 ## 2.15.3
+
 - Fix inconsistent startup for adapters that remap hardware flow control pins for firmware flashing.
 
 ## 2.15.2
+
 - Add baudrate list option 1000000 (Nordic Semiconductor nRF Connect SDK firmware)
 
 ## 2.15.1
+
 - Make radio spinel recovery more reliable by clearing source match tables before restoring
 
 ## 2.15.0
+
 - Automatically migrate the active dataset to a new adapter when changing the addon serial port path.
 
 ## 2.14.0
+
 - Remove firmware flashing from the addon, this is now handled by Core 2025.7.0.
 
 ## 2.13.0
+
 - Bump to OTBR POSIX version b067e5ac (2025-01-13 22:32:22 -0500)
 - Bump universal-silabs-flasher to 0.0.28
 - Remove dataset deletion REST API backwards compatibility patch. The minimum Core version for this add-on is now 2023.9.0
@@ -55,10 +69,12 @@
 - Update flasher script to work with Home Assistant Yellow with CM5
 
 ## 2.12.1
+
 - Fix possible race condition between otbr-agent-configure and otbr-agent-rest-discovery
   services causing failed startup ([#3826](https://github.com/home-assistant/addons/issues/3826))
 
 ## 2.12.0
+
 - Bump universal-silabs-flasher to 0.0.23
 - Bump OTBR firmwares to latest versions
 - Bump to OTBR POSIX version b041fa52daa (2024-11-14 08:18:28 -0800)
@@ -66,7 +82,7 @@
 
 ## 2.11.1
 
--  Fix issue with USB TI CC2652 based devices
+- Fix issue with USB TI CC2652 based devices
 
 ## 2.11.0
 
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.16.4
+version: 2.16.5
 slug: openthread_border_router
 name: OpenThread Border Router
 description: OpenThread Border Router add-on


GollyJer:
would love to have this ability.

t3hk0d3:
@agners Do you think a more sophisticated fallback for default network interface would be necessary? Previously it was just hardcoded "eth0" (that no longer in use with modern systemd/Linux kernels).

Current behaviour is to pick first interface if primary interface is not available from Supervisor API. 
Maybe more correct behaviour would to give an fatal error instead?

t3hk0d3:
please check PR

Adpocalyptic:
> @agners Do you think a more sophisticated fallback for default network interface would be necessary? Previously it was just hardcoded "eth0" (that no longer in use with modern systemd/Linux kernels).
> 
> 
> 
> Current behaviour is to pick first interface if primary interface is not available from Supervisor API. 
> 
> Maybe more correct behaviour would to give an fatal error instead?

I mean if it's a more sophisticated fallback that's more inline with the current behaviour (_as opposed to an outright fatal error_) then it does seem nicer for end-user experience. Like an automatic fix that can simply be logged if ever needed.

Unless an opt-out toggle to force it to stop trying different fallbacks is otherwise worth adding, maybe for theoretical troubleshooting/debug reasons? (Legacy Fallback Behaviour Off/On) In which case it can halt on fatal error with whatever additional output could be relevant there.

But that being said, I don't know if I'm misreading what you meant there 🤔

agners:
> @agners Do you think a more sophisticated fallback for default network interface would be necessary? Previously it was just hardcoded "eth0" (that no longer in use with modern systemd/Linux kernels).

Yeah I agree hard coded `eth0` does not make much sense today.

> 
> Current behaviour is to pick first interface if primary interface is not available from Supervisor API. Maybe more correct behaviour would to give an fatal error instead?

I think I'd lean towards fatal error indeed. Are there any known situations where people run into this?


I previously declined similar PRs, mostly on the grounds of feature creep. With the Matter Server it is possible to pass additional arguments to define which should be seen as primary interface. But the Matter Server still uses all network interfaces by default.

I think a much more elegant solution would be to have macvlan support, and then one could define what network an add-on should be isolated on.

t3hk0d3:
> > Current behaviour is to pick first interface if primary interface is not available from Supervisor API. Maybe more correct behaviour would to give an fatal error instead?
> 
> I think I'd lean towards fatal error indeed. Are there any known situations where people run into this?

I am just afraid of situation where some random Docker interface is picked up instead, and people end up wasting extra time troubleshooting Matter connectivity cased by such implicit logic.

> I previously declined similar PRs, mostly on the grounds of feature creep. With the Matter Server it is possible to pass additional arguments to define which should be seen as primary interface. But the Matter Server still uses all network interfaces by default.

It's pretty much must-have for VLAN-ed networking setups. It's almost impossible to properly join Thread device if commissioner (eg iPhone) is on the different sub-network.

> With the Matter Server it is possible to pass additional arguments to define which should be seen as primary interface.

With OTBR that's not a case - backbone interface is passed in the mandatory params (`trel://${backbone_if}`). I don't see a way how that can be worked around.

> I think a much more elegant solution would be to have macvlan support, and then one could define what network an add-on should be isolated on.

MACVLAN network that would be ideal. I personally run all must containers (beside HAOS addons) in a separate VLAN with each container having separate IP address via macvlan. That solves so much problems for me - no need for port forwarding, host networking mode, etc.

However AFAIK to achieve that some major changes to Supervisor would be necessary.

Another problem - IPv6 networking in Docker is kinda crap. It doesn't support SLAAC/DHCPv6 without some extraordinary workarounds. You need to explicitly specify ip ranges. Router Announcements and Prefix Delegation are not supported either.

t3hk0d3:
> > Current behaviour is to pick first interface if primary interface is not available from Supervisor API. Maybe more correct behaviour would to give an fatal error instead?
> 
> I think I'd lean towards fatal error indeed. Are there any known situations where people run into this?

@agners I've updated PR to throw an error if no primary network interface is available.

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -72,10 +76,17 @@ case "${otbr_log_level}" in
...
 esac
 
-# shellcheck disable=SC2086
-if [ -z ${backbone_if} ]; then
-    bashio::log.warning "No primary network interface found! Using static eth0."
-    backbone_if="eth0"
+# Fallback to primary supervisor network interface
+if [ -z "${backbone_if}" ]; then
+    backbone_if="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+    if [ -n "${backbone_if}" ]; then
```
agners: This logic seems reversed? 🤔

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -72,10 +76,17 @@ case "${otbr_log_level}" in
...
+        bashio::log.warning "No network interface configured! Using primary interface instead (${backbone_if})."
+    fi
```
agners: Can't we just make this the else branch of `if bashio::config.has_value 'backbone_interface'; then`? I think it would be more readable. Then we can also simply bail out using `bashio::exit.nok` in case `backbone_if` returned by the Supervisor API is zero length.

agners:
> > With the Matter Server it is possible to pass additional arguments to define which should be seen as primary interface.
> 
> With OTBR that's not a case - backbone interface is passed in the mandatory params (`trel://${backbone_if}`). I don't see a way how that can be worked around.

Yeah this makes the OTBR case a bit different to Matter, I agree.

> 
> > I think a much more elegant solution would be to have macvlan support, and then one could define what network an add-on should be isolated on.
> 
> MACVLAN network that would be ideal. I personally run all must containers (beside HAOS addons) in a separate VLAN with each container having separate IP address via macvlan. That solves so much problems for me - no need for port forwarding, host networking mode, etc.
> 
> However AFAIK to achieve that some major changes to Supervisor would be necessary.

I don't think it would be all that hard, but yeah it needs changes. I'd suggest an extra configuration block in the apps configuration tab if an app declares `host_network: true`. It would be a simple drop down to select the interface. Then the add-on gets hooked up to such a macvlan. Passing the options to Docker should be fairly straight forward.

The IP configuration is the bigger challenge: Should this simply be delegated to the add-on?

> Another problem - IPv6 networking in Docker is kinda crap. It doesn't support SLAAC/DHCPv6 without some extraordinary workarounds. You need to explicitly specify ip ranges. Router Announcements and Prefix Delegation are not supported either.

But isn't SLAAC handled by the kernel if a macvlan interface is used? :thinking:  DHCPv6 would need a per app specific solution I guess :cry:

t3hk0d3:
@agners 
> The IP configuration is the bigger challenge: Should this simply be delegated to the add-on?

Thats where biggest problem arises. Docker doesn't support DHCP for container out-of-box, so people would have to manually assign IP-addresses to container. That really sucks.

> > Another problem - IPv6 networking in Docker is kinda crap. It doesn't support SLAAC/DHCPv6 without some extraordinary workarounds. You need to explicitly specify ip ranges. Router Announcements and Prefix Delegation are not supported either.
> 
> But isn't SLAAC handled by the kernel if a macvlan interface is used? 🤔 

According to this [doc](https://docs.docker.com/engine/daemon/ipv6/#use-ipv6-for-the-default-bridge-network) - everything needs to be configured explicitly in `/etc/docker/daemon.json` 😢 

> DHCPv6 would need a per app specific solution I guess 😢

That's actually something Supervisor can do for the app. Supervisor does DHCPv6 request for App, grabs IPv6 and use it to initialize Docker container.

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
t3hk0d3: Makes sense. Good catch.

I've updated PR to fix it.

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
t3hk0d3: That's logic for showing that primary interface was used instead of configured value.

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -23,7 +23,20 @@ if bashio::config.exists 'autoflash_firmware'; then
...
+
+# Backbone network interface can be configured by user or auto-detected by looking for the primary network interface in the system.
+# If no primary network interface is found, the add-on will exit with an error and ask the user to configure a backbone network interface in the add-on options.
+if bashio::config.has_value 'backbone_interface'; then
+    backbone_if="$(bashio::config 'backbone_interface')"
+else
+    backbone_if="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface // empty')"
+    if [ -z "${backbone_if}" ]; then
+        # If no backbone interface is found, exit with error
+        bashio::exit.nok "No primary network interface found. Please configure a backbone network interface in the add-on options."
```
agners:
```suggestion
        bashio::exit.nok "No primary network interface found. Please configure a backbone network interface in the apps configuration."
```

openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
```diff
@@ -23,7 +23,20 @@ if bashio::config.exists 'autoflash_firmware'; then
     bashio::addon.option 'autoflash_firmware'
 fi
 
-backbone_if="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
+backbone_if=""
+
+# Backbone network interface can be configured by user or auto-detected by looking for the primary network interface in the system.
+# If no primary network interface is found, the add-on will exit with an error and ask the user to configure a backbone network interface in the add-on options.
```
agners:
```suggestion
# If no primary network interface is found, the app will exit with an error and ask the user to configure a backbone network interface in the app configuration.
```

PR #4306, opened 2026-01-22
matter_server: Add necessary dependencies for BLE

Apollon77:
as by https://github.com/stoprocent/noble?tab=readme-ov-file#linux

## Summary by CodeRabbit

* **New Features**
  * Added support for Bluetooth Low Energy (BLE) functionality with required system-level dependencies.

* **Chores**
  * Bumped release version to 8.2.1 and updated changelog.
  * Installer/startup adjusted to run package lifecycle scripts in the foreground to improve install-time handling.

fixes https://github.com/matter-js/matterjs-server/issues/106

### DIFF
--- a/matter_server/Dockerfile
+++ b/matter_server/Dockerfile
@@ -33,7 +33,6 @@ RUN \
         curl \
         ca-certificates \
         xz-utils \
-        bluetooth \
         bluez \
         libbluetooth-dev \
         libudev-dev \


lboue:
Good idea. I'd like to validate that in my HA test instance.

matter_server/Dockerfile
```diff
@@ -33,6 +33,10 @@ RUN \
         curl \
         ca-certificates \
         xz-utils \
+        bluetooth \
```
agners: Nit: This is a metapackage which only has `bluez` as dependency. I guess we need bluez explicitly, so I'd drop this.


```suggestion
```

PR #4313, opened 2026-01-26
Builder workflow: rename add-on to app

c0ffeeca7:
as per Architecture discussion: Rename “Add-ons” to “Apps” https://github.com/home-assistant/architecture/discussions/1287

related to task: https://github.com/home-assistant/home-assistant.io/issues/43180


c0ffeeca7:
> I think that in this case we could also change the variables but it can be done in a follow-up PR.

sure, will do! Let me create a new PR once these ones are merged. In case there are more variables on other pages.

Issue #4301, opened 2026-01-20
Bug report: OTBR add-on crashes mDNSResponder (FD_SETSIZE=1024) causing Thread BR to disappear


benek984:
### Describe the issue you are experiencing

The OpenThread Border Router add-on repeatedly crashes its internal mDNSResponder process due to hitting the hard FD_SETSIZE=1024 limit used by select().

When this happens, the add-on stops advertising _meshcop._udp / _trel._udp, and the Thread Border Router disappears from Home Assistant’s Thread integration until the add-on is restarted.

This happens consistently on my system after some runtime (minutes to hours), even though the Thread radio itself is functioning correctly.

Observed behavior

After some uptime, OTBR logs show:

Default: requestIOEvents called with fd 1024 > FD_SETSIZE 1024.
mdnsd: mDNSPosix.c:2527: requestIOEvents: Assertion `0' failed.
mDNS ended with exit code 256 (signal 6)

Immediately after this:

mDNSResponder restarts

OTBR fails to publish _meshcop._udp / _trel._udp

Home Assistant no longer sees the Thread Border Router

Restarting the add-on temporarily restores functionality

Root cause analysis (confirmed)

After extensive debugging:

The crash is caused by mDNSResponder using select(), which has a compile-time FD_SETSIZE=1024 limit

The OTBR add-on runs with:

host_network: true

many visible interfaces

multiple add-ons active (≈24)

TREL (_trel._udp) significantly increases mDNS socket usage

Once any FD ≥1024 is registered, mDNSResponder asserts and aborts

This is not hardware-related and not USB-related.

Why this is a bug in the add-on

The add-on enables TREL by default

There is no configuration option to disable it

TREL + mDNSResponder + host networking causes a deterministic crash

The crash takes down discovery, breaking Thread integration

This affects otherwise valid, supported HA setups

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

OpenThread Border Router

### What is the version of the add-on?

2.15.3

### Steps to reproduce the issue

1.
2.
3.
...

## System Information

version | core-2026.1.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.13.11
os_name | Linux
os_version | 6.12.63-haos
arch | x86_64
timezone | Europe/London
config_dir | /config

[...]

Issue #4141, opened 2025-08-21
Open Thread Border Router responses to `_meshcop._udp.local` requests are missing TXT records


miawgogo:
### Describe the issue you are experiencing

Android is unable to find the border router which is leading to weird bugs relating to setting up thread credentials on android as it requires that the border router send certian values in a mDNS response.

When OTBR does a unsolicited advertisement it does contain the txt record, which i think is why iOS/iPadOS does not show the same issues as it uses a cached response, but android seems to not want to use its mdns cache and sends new requests for `_meshcop._udp.local`

OTBR running on raspbian(the native setup in the OTBR guide) does have the missing TXT record, indicating that its related to the home assistant mDNS setup

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

OpenThread Border Router

### What is the version of the add-on?

2.13.0 

### Steps to reproduce the issue

1. Setup Home assistant with the Open Thread Border Addon (optionally have a second border router running to compare)
2. Run Wireshark on annother device on the network with a display filter `dns.resp.name contains "_meshcop._udp.local" || dns.qry.name contains "_meshcop._udp.local" `
3. Open settings and search for "Thread Networks"
4. When its loading you should notice a request from your android device for the service and a response from home assistant and your border router(s)
5. Note that the TXT record from home assistant is missing.

## System Information

version | core-2025.8.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.13.3
os_name | Linux
os_version | 6.12.34-haos-raspi
arch | aarch64
timezone | Europe/London
config_dir | /config

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 16.1
-- | --
update_channel | stable
supervisor_version | supervisor-2025.08.1
agent_version | 1.7.2
docker_version | 28.3.3
disk_total | 57.8 GB
disk_used | 4.1 GB
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | rpi3-64
supervisor_api | ok
version_api | ok
[...]

PR #4281, opened 2026-01-04
 OpenThread Border Router - Bump Thread Reference to 2025-06-12

lboue:
## Summary

This PR updates the OpenThread Border Router add-on to Thread 1.4 with a newer OTBR reference, while adopting the native OpenThread mDNS implementation for improved DNS-SD handling.

**Key Changes:**

- **Bump OTBR version** to `thread-reference-20250612` (June 12, 2025 reference)
- **Enable Thread 1.4** by setting `THREAD_VERSION` to `1.4`
- **Switch to native OpenThread mDNS module** with these build options:
  - Remove `OTBR_MDNS=mDNSResponder` - no longer needed
  - Remove `OTBR_DNSSD_DISCOVERY_PROXY=ON` - functionality handled by native mDNS
  - Remove `OTBR_SRP_ADVERTISING_PROXY=ON` - functionality handled by native mDNS
  - Set `DOTBR_DNS_UPSTREAM_QUERY=OFF` - not compatible with native mDNS module
- **Replace channel monitor patch** with configuration flag:
  - Remove `0001-channel-monitor-disable-by-default.patch`
  - Add `OPENTHREAD_CONFIG_CHANNEL_MONITOR_AUTO_START_ENABLE=0` to config header
  - Cleaner approach than patching source code

**Benefits:**

- Updated to Thread 1.4 reference implementation
- Native mDNS eliminates dependency on external mDNSResponder package
- Better Docker image size - mDNSResolver not installed during bootstrap
- Enhanced coexistence with host avahi-daemon
- Improved debugging with additional CLI commands:
  - `ot-ctl srp server service` / `ot-ctl srp server host`
  - `ot-ctl mdns services` / `ot-ctl mdns hosts`
- Cleaner configuration using configuration flags instead of source patches

**Build Status:** ✅ All builds successful (amd64, aarch64)

PR #4323, opened 2026-01-27
openthread_border_router: Beta toggle for testing Thread 1.4

puddly:
This PR adds a beta toggle to the OTBR addon by bundling both the current OTBR release and [the latest OTBR commit](https://github.com/openthread/ot-br-posix/commit/624a7d98e0dac5984c2982068f71fc81d2dbceb5) (we probably should pick a more deliberate ot-br-posix commit and look through the new build config options more closely).

For testing, I've switched the beta build from mDNSResponder to the new embedded OpenThread mDNS.

Everything works in my rudimentary tests but I'm keeping this PR drafted until I can test a bit more. You can find updated (beta) OpenThread firmware for the ZBT-2 here: https://github.com/NabuCasa/silabs-firmware-builder/pull/182 ([direct link](https://github.com/NabuCasa/silabs-firmware-builder/actions/runs/21343419780/artifacts/5251707624)).

The Dockerfile has been rewritten to use isolated build stages for each variant and install only the minimum runtime dependencies. This shrinks the final image from 400MB to about 275MB.

---

Possibly related issues: #4301, #4141, #4141

## Summary by CodeRabbit

* **New Features**
  * Added beta mode toggle to switch between Thread 1.3 (stable) and Thread 1.4 (beta), with native OpenThread mDNS support in beta mode

* **Documentation**
  * Added beta configuration option documentation

* **Chores**
  * Version bumped to 2.16.0
  * Updated base system images to latest stable version

### DIFF
--- a/openthread_border_router/DOCS.md
+++ b/openthread_border_router/DOCS.md
@@ -70,6 +70,7 @@ Add-on configuration:
 | firewall           | Enable OpenThread Border Router firewall to block unnecessary traffic |
 | nat64              | Enable NAT64 to allow Thread devices accessing IPv4 addresses |
 | network_device     | IP address and port to connect to a network-based RCP (see below) |
+| beta               | Enable beta mode with Thread 1.4 and native OpenThread mDNS |
 
 > [!WARNING]
 > The OTBR expects the RCP connected radio to be on a reliable link such as
--- a/openthread_border_router/translations/en.yaml
+++ b/openthread_border_router/translations/en.yaml
@@ -29,6 +29,10 @@ configuration:
     description: >-
       Enable IPv6 to IPv4 network address translation. This allows Thread
       devices to communicate with devices on the Internet.
+  beta:
+    name: Beta
+    description: >-
+      Enable beta mode with Thread 1.4 and native OpenThread mDNS.
 network:
   8080/tcp: OpenThread Web port
   8081/tcp: OpenThread REST API port


t3hk0d3:
FYI: https://github.com/home-assistant/addons/pull/4281

puddly:
Ah. I didn't see @lboue made a similar PR last month!

sairon:
@t3hk0d3 FTR, I nudged @agners today about the other PR, because I'm not that much involved in Matter/Thread development and I didn't feel to be the right one to call such a big change in. He told me that he's a bit hesitant to go with changes proposed in that PR by default, and an opt-in might be a better way. Great that @puddly tackled it. So #4281 should be closed in favor of this one, maybe the changes from there consolidated in here.

openthread_border_router/Dockerfile
```diff
@@ -1,72 +1,115 @@
...
+        -DOTBR_PRODUCT_NAME="OpenThread Border Router" \
+        # The Meshcop service name is specified in `openthread-core-ha-config-posix.h`
+        -DOTBR_WEB=ON \
+        -DOTBR_BORDER_ROUTING=ON \
+        -DOTBR_REST=ON \
+        -DOTBR_BACKBONE_ROUTER=ON \
+        -DOTBR_TREL=ON \
+        -DOTBR_NAT64=ON \
+        -DOT_POSIX_NAT64_CIDR="192.168.255.0/24" \
+        -DOTBR_DNS_UPSTREAM_QUERY=ON \
```
puddly, line 59: @agners do you happen to know why we have this enabled? We have NAT64 disabled by default but I think with this option enabled, Thread devices can still make DNS queries.

phaze75:
> Everything works in my rudimentary tests but I'm keeping this PR drafted until I can test a bit more. You can find updated (beta) OpenThread firmware for the ZBT-2 here: [NabuCasa/silabs-firmware-builder#182](https://github.com/NabuCasa/silabs-firmware-builder/pull/182) ([direct link](https://github.com/NabuCasa/silabs-firmware-builder/actions/runs/21343419780/artifacts/5251707624)).

Is there also an updated (beta) OpenThread firmware for ZBT-1?

lboue:
> Is there also an updated (beta) OpenThread firmware for ZBT-1?

ZBT-1 firmware: [firmware-build-skyconnect_openthread_rcp_3.0.0.0_GitHub-61e43cffb_gsdk_2025.12.0](https://github.com/NabuCasa/silabs-firmware-builder/actions/runs/21343419780/artifacts/5251709039)

puddly:
My only real question with this PR is `OTBR_DNS_UPSTREAM_QUERY` but since we're keeping the value the same and it's all behind a beta flag, I think we can just release an updated version if we decide to change it in the future.

t3hk0d3:
> My only real question with this PR is `OTBR_DNS_UPSTREAM_QUERY` but since we're keeping the value the same and it's all behind a beta flag, I think we can just release an updated version if we decide to change it in the future.

Donyou think it would make sense to have it as config option?

Alternatively we can make freetext input for env variables, for maximum flexibility (maybe as separate MR)

puddly:
I don't think it's configurable after compilation unless you use the DBus API to change it. Or block DNS entirely within the container and have that be a configurable option.

openthread_border_router/Dockerfile
agners: Yeah not sure, I probably copied that from the `.github/workflows/border_router.yml` workflow upstream. It got introduced with https://github.com/openthread/ot-br-posix/pull/1777 upstream.

I'd guess that its useful for Thread devices which need Internet connectivity... It seems this can be controlled at runtime as well, maybe we can control it together with the NAT64 flag? :thinking:

puddly: Now everything makes sense. We already do:

```bash
if bashio::config.true 'nat64'; then
    bashio::log.info "Enabling NAT64."
    ot-ctl nat64 enable
    ot-ctl dns server upstream enable
fi
```

puddly:
Ah, I take that back. When you enable `OTBR_FEATURE_FLAGS=ON`, these switch to being disabled on startup:
- NAT64
- DNS upstream query
- DHCPv6-PD

We explicitly re-enable them with `ot-cli` when the addon starts based on your config. Right now the NAT64 option controls both NAT64 and DNS upstream query, which I think is correct behavior.

puddly:
The computed name isn't configurable but I think this may actually be an improvement, since we can (theoretically) run multiple BRs from the same Home Assistant instance?

openthread_border_router/Dockerfile
```diff
@@ -1,72 +1,119 @@
...
+ENV PLATFORM=debian
+ENV RELEASE=1
+ENV WEB_GUI=1
+ENV REST_API=1
+ENV DHCPV6_PD_REF=0
+ENV DOCKER=1
+
+
+COPY openthread-core-ha-config-posix.h /usr/src/
+COPY 0001-channel-monitor-disable-by-default.patch /usr/src/
```
agners, line 86: Ins't this patch obsolete with `OPENTHREAD_CONFIG_CHANNEL_MONITOR_AUTO_START_ENABLE` being available upstream? 🤔

agners: Ah nevermind, required for non-beta of course :sweat_smile:

puddly: It's only for the "stable" variant, we've upstreamed all of our patches so the "beta" variant builds without them. That config was not included in `b067e5ac5f8b3e92750df24922017eee2bc0fa04` (Jan 13, 2025), [the PR](https://github.com/openthread/openthread/pull/11164) was merged on Feb 21, 2025.

PR #4344, opened 2026-01-28
[openthread_border_router] Migrate OpenThread settings when switching between beta/stable

puddly:
I noticed that the binary format has not actually changed for OpenThread itself: no new fields have been added, just the internal version has bumped from `4` to `5`. We theoretically can just tweak this value to keep OpenThread from regenerating Thread network information.

Note that the active dataset isn't changed in the current addon state, just the Thread network information. This will result in devices rerouting a little since your border router effectively disconnects and another fresh one re-connects.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.16.1
+- Preserve Thread network identity when switching between stable and beta mode
+
 ## 2.16.0
 - Add beta toggle to switch between Thread 1.3 (stable) and Thread 1.4 (beta)
 - Beta mode uses OpenThread's built-in mDNS instead of mDNSResponder
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.16.0
+version: 2.16.1
 slug: openthread_border_router
 name: OpenThread Border Router
 description: OpenThread Border Router add-on
--- a/openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
+++ b/openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
@@ -148,10 +148,18 @@ echo "${otbr_rest_listen_port}" >> /tmp/otbr-agent-rest-api
 
 # Migrate OTBR settings to new adapter if needed
 bashio::log.info "Migrating OTBR settings if needed..."
+
+if bashio::config.true 'beta'; then
+    thread_version=5
+else
+    thread_version=4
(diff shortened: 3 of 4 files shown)


puddly:
I've tested this a few times without any ill effects. Compared to Thread 1.3, Thread 1.4 proactively finds a child device within 15s of startup. It still takes about 60s for the Matter server to actually start communicating with devices though.

agners:
> I've tested this a few times without any ill effects. Compared to Thread 1.3, Thread 1.4 proactively finds a child device within 15s of startup. 

Was there any difference in speed with/without this change (with new extaddr?)

> It still takes about 60s for the Matter server to actually start communicating with devices though.

This is a bit the beauty and the ugly of Matter/Thread: The network layer is completely separately. Largely, Matter doesn't know or care about how the Thread network underneath it reconnects/reestablish. So as long as the network addresses are still the same, a device can loose its Thread network intermittently, it will still be able to inform the Matter server about new events (motion or whatnot) instantly (given the network is available at the time of the event).

The Matter server only starts to look for the device if it does not get the subscription device check-in within the specified max interval (afaik, we use 60s for wall powered devices and 600s for batter operated devices). Similar on a device reboot: The controller (Matter client) needs to connect to the device (Matter server) and reestablish the subscription. However, in theory, the device has still all relevant information.

If the Thread border router was the designated SRP, devices should also reestablish the service registration. This causes a mDNS service announcement on the network, to which a Matter server might react. 

That said, some devices persist the subscription, and pro-actively ping the controller after reboot to inform it of its new state. In this case the Matter subscription is immediately reestablished (I know that at least Nanoleaf bulbs do that).

> I noticed that the binary format has not actually changed for OpenThread itself: no new fields have been added, just the internal version has bumped from `4` to `5`.

Maybe this is a bit far fetched, but maybe the version has been bumped to force a extaddr change, to force some regeneration/recalculations in the Thread mesh :thinking: .

Anyhow, I guess if we don't see any ill effects, why not.

agners:
With discussed this in todays protocol meeting and nobody really complained about the change during beta testing. After a restart (without backup BRs) the user has to wait for a couple minutes anyways, so it doesn't really matter that the OTBR app appears as new BR.

agners:
I did dive into when the logic of this internal version change got introduced to find out if there is a good reason the OTBR/OpenThread forces this. It seems this was a quite conscious decision back in Thread 1.2:

https://github.com/openthread/openthread/pull/4325/changes#diff-47bc8c061f3aab7ae9b35d5bf7ab39d53cf6def0647d24fdf24b6e6b99ed415cR392

At least in Thread 1.2 a new feature got introduced (enhanced keep-alive) which required a reattach to be used. So this force this MLE level reattach.

I think it was the right choice to leave this OpenThread internal alone and accept the churn.

PR #4367, opened 2026-02-01
Rename devcontainer from 'Add-ons' to 'Apps'

AJediIAm:
Not sure if this is safe to change.


sairon (approved):
This should be purely cosmetic, I don't see any major side-effect besides this one:

Image

PR #4370, opened 2026-02-01
ssh: Update description for additional software packages

AJediIAm:
## Summary by CodeRabbit

* **Documentation**
  * Updated description to clarify that additional Alpine packages are installed in the app container.

### DIFF
--- a/ssh/translations/en.yaml
+++ b/ssh/translations/en.yaml
@@ -9,7 +9,7 @@ configuration:
   apks:
     name: Packages
     description: >-
-      Additional software packages to install in the Home Assistant app.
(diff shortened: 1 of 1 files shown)


ssh/translations/en.yaml
```diff
@@ -9,7 +9,7 @@ configuration:
   apks:
     name: Packages
     description: >-
-      Additional software packages to install in the add-on container.
+      Additional software packages to install in the Home Assistant app.
```
sairon: Your suggestion makes it even more confusing, let's go with something like this:
```suggestion
      Additional Alpine packages to install in the app container.
```

PR #4387, opened 2026-02-05
[openthread_border_router] Fix TREL being disabled by default in beta mode

puddly:
OpenThread refactored internally and TREL is now a runtime-controllable flag, same as all the others. We should explicitly enable TREL on startup.

## Summary by CodeRabbit

* **Bug Fixes**
  * TREL (Thread Relay) is no longer disabled by default in beta mode.


Apollon77:
@agners Please merge

PR #4403, opened 2026-02-08
Added additional ports that can be used for custom servers.

loewexy:
The addon currently allows only exposing ports 80 and 443. But it allows the user to add custom servers. If one of those should use another port this is not possible using the addon. Nginx will boot and listen to the port but there is no way to expose that port.

To fix this I suggest adding 3 custom ports, that can be used by the use for custom servers. From my point of view it is okay to just offer some ports since it is easy for the user to select the exact port that should be opened on the host outside the container via the addon config.

I know it is currently possible to have multiple server {} directives differentiating via the name but for some use cases this is not enough. My actual use case for which i think this is the perfect solution is requiring client cert authentication when connecting from the internet. For that i can configure a second server that enforces client cert auth and make the port forwarding in the router to this port. This is only hindered by the missing possibility to expose the second port.

Please let me know if the pull request can be improved i am happy to fix your remarks.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that ports 8443, 8444 and 8445 are available for custom server use and should be enabled in network settings.

* **Configuration**
  * Added three TCP port entries (8443, 8444, 8445) for custom server configuration.

### DIFF
--- a/nginx_proxy/DOCS.md
+++ b/nginx_proxy/DOCS.md
@@ -74,7 +74,7 @@ The filename of the NGINX configuration for the default server, found in the `/s
 
 The filename(s) of the NGINX configuration for the additional servers, found in the `/share` directory.
 
-You can use the ports 8443, 8444 and 8445 for custom servers.
+Ports 8443, 8444, and 8445 are available for custom servers. Enable them in the add-on network configuration before use.
 
 ### Option `cloudflare` (optional)
 
--- a/nginx_proxy/translations/en.yaml
+++ b/nginx_proxy/translations/en.yaml
@@ -33,6 +33,6 @@ network:
   443/tcp: HTTPS (SSL) Port
   443/udp: HTTP/3 (QUIC) Port, set to same as HTTPS (SSL) port to enable HTTP/3
   80/tcp: HTTP (non-SSL) Port
-  8443/tcp: Port that can be used for custom servers
-  8444/tcp: Port that can be used for custom servers
-  8445/tcp: Port that can be used for custom servers
+  8443/tcp: Port tcp/8443 that can be used for custom servers
+  8444/tcp: Port tcp/8444 that can be used for custom servers
+  8445/tcp: Port tcp/8445 that can be used for custom servers


loewexy:
I fixed the two nitpick comments from CodeRabbit review, i think the new version makes it better to understand for users.

agners:
I see your point, but it feels quite hacky to have this "spare ports" there :cry: 

Did you consider just enforce client authentication via NGINX proxy and connect directly to Home Assistant for the non-client authentication (local?) use case?

frenck:
Honestly, this is not the goal of this apps. The goal is to offload / proxy Home Assistant. I don't think we should support this usecase (and most certainly not do it using a workaround like is proposed in this PR).

../Frenck  

<p>
  <a href="https://www.linkedin.com/in/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://youtube.com/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://bsky.app/profile/frenck.social"></a>
  &nbsp;&nbsp;
  <a href="https://fosstodon.org/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.instagram.com/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://www.threads.net/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.facebook.com/frenck.dev/"></a>
  &nbsp;&nbsp;
  <a href="https://x.com/frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.tiktok.com/@frenck.nl"></a>
</p>

Blogging my personal ramblings at <a href="https://frenck.dev">frenck.dev</a>

loewexy:
@agners Yes that was my first idea, but i have smart home devices that only accept HTTPS URLs as callback url. But they do not support client authentication. Therefore i also need a local access with HTTPS but without client auth.

I agree that this is not the most elegant solution. But given we have the `customize.servers` option is seems to be intended to have additional servers. However home assistant plugins do not seem to allow to expose configurable additional ports. That would be much cleaner to just go to the settings and add a port here.

Also the possibility to run multiple instances of an addon would be sufficient here. This is my current workaround, I took the original addon modified its name and uploaded it as a custom addon so that i can run 2 instances and have one with client auth on one port and another one without client auth on another port.

agners:
The approach with spare ports is just too hacky, we don't want to go down that road.

With https://github.com/home-assistant/addons/pull/4284 merged, you can now enable SSL on HA Core side, and use this app for client certificate authentication.

Your work around could also be a cleaner, alternative way: Have a separate app with client auth settings nicer integrated :thinking: :shrug:

PR #4412, opened 2026-02-11
otbr: ignore temporary settings files in migration script

sairon:
In some cases, OTBR creates temporary settings files [1]. If the RCP becomes unavailable (seems to happen particularly for network connected adapters), file with a newer timestamp containing ephemeral data is created and incorrectly migrated as configuration for the reconnected adapter later:

```
Settings file for adapter 187a3efffe0182dd already exists at /data/thread/0_187a3efffe0182dd.data but appears to be old, archiving
Wrote new settings file to /data/thread/0_187a3efffe0182dd.data
```

This is how the thread folder looks like after the migration:

```
-rw------- 1 kpt kpt  32 Feb 11 09:57 0_0-tmp.data
-rw------- 1 kpt kpt  73 Feb 11 09:58 0_187a3efffe0182dd.data
-rw------- 1 kpt kpt 278 Feb 11 09:47 0_187a3efffe0182dd.data.backup-20260211095840
```

The new data file is clearly migrated from the 0_0-tmp.data, creating a new configuration out of ephemeral data which is invalid and prevents correct initialization of the OTBR app.

This change adds a check that the file to be migrated is matching a valid name created by [2], effectively ignoring the temporary files.

[1] https://github.com/openthread/openthread/blob/thread-reference-20250612/src/posix/platform/tmp_storage.cpp
[2] https://github.com/openthread/openthread/blob/thread-reference-20250612/src/posix/platform/settings.cpp#L93

## Summary by CodeRabbit

* **Bug Fixes**
  * Migration now properly ignores ephemeral temporary settings files, improving reliability and preventing potential issues during the settings migration process.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.16.3
+- Ignore ephemeral temporary settings files in migration
+
 ## 2.16.2
 - Fix TREL being disabled by default in beta mode
 
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.16.2
+version: 2.16.3
 slug: openthread_border_router
 name: OpenThread Border Router
 description: OpenThread Border Router add-on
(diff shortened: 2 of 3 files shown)


puddly:
Let's go with this approach. The `0_0-tmp.data` file is unexpected and we should be ignoring it. If we run into further issues, we can validate further with https://github.com/home-assistant/addons/pull/4413.

It's a terrible coincidence that both files are TLV format and share type IDs.

PR #4413, opened 2026-02-11
otbr: improve validation of OTBR settings file

sairon:
The current validation only checks if ACTIVE_DATASET (0x1) is present in the settings files. However, this can incorrectly match on data that doesn't contain Thread TLVs, but also for ephemeral data (boot time marker) created by OpenThread's TmpStorage [1], as seen in this hexdump of the file:

```
00000000: 0100 0800 b477 8b69 0000 0000 0200 1000  .....w.i........
```

To make the check more robust, check not only for the key, but also for value length, which should normally be larger than time_t.

[1] https://github.com/openthread/openthread/blob/thread-reference-20250612/src/posix/platform/tmp_storage.cpp#L75


sairon:
This is to mitigate problem stemming from the issue described in #4412. Since these two PRs are in collision, and maybe the validation might not be ideal, I'm submitting it only as draft now, but will update the changelog and version once the other PR is merged or once it's the solution is approved.

PR #4415, opened 2026-02-11
otbr: Only explicitly enable TREL when on beta

puddly:
https://github.com/home-assistant/addons/issues/4409 seems to indicate that running `ot-cli trel enable` during startup actually causes issues for stable users, presumably because the "feature" refactor in ot-br-posix changed how things started/stopped. To maintain the old behavior exactly, we should start TREL only on the beta.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.16.4
+- Only explicitly enable TREL in beta mode to avoid startup issues on stable
+
 ## 2.16.3
 - Ignore ephemeral temporary settings files in migration
 
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.16.3
+version: 2.16.4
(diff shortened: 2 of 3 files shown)


agners:
I was able to reproduce the error reported in #4409 locally as well, but I had to restart the OTBR a couple of times (with beta flag disabled):

```
[13:32:41] INFO: Starting otbr-web...
00:00:00.056 [N] P-Netif-------: Changing interface state to up.
[INFO]-WEB-----: Running 0.3.0-b067e5ac-dirty
[INFO]-WEB-----: Border router web started on wpan0
read: Connection reset by peer
s6-rc: warning: unable to start service otbr-agent-configure: command exited 1
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
/run/s6/basedir/scripts/rc.init: fatal: stopping the container.
```

It should be safe to call `ot-ctrl trel enable`. This seems more like a race condition, maybe more likely with the old OTRB :thinking:

agners:
Should be addressed with #4416.

PR #4416, opened 2026-02-12
openthread_border_router: Fix race condition if web interface is enabled

agners:
Only a single client can use the OTBR CLI unix socket at `/run/openthread-wpan0.sock`. Serialize `otbr-web` service startup to be after `otbr-agent-configure` to avoid concurrent access. See also https://github.com/home-assistant/addons/issues/4409#issuecomment-3890875190.

Also fix `otbr-agent` ready detection, potentially leading to `otbr-agent` declared ready prematurely, which in turn can lead depending services to fail.

Fixes: #4409

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition that could occur during startup when the web frontend is enabled.

### DIFF
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.16.4
+
+- Fix race condition during startup if web frontend is enabled
+
 ## 2.16.3
 - Ignore ephemeral temporary settings files in migration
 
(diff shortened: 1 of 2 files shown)


puddly:
Good catch! This makes sense.

puddly:
Should we have `otbr-web` depend on `otbr-agent-rest-discovery`? They both would connect to the socket on startup, right?

agners:
> Should we have `otbr-web` depend on `otbr-agent-rest-discovery`? They both would connect to the socket on startup, right?

No, `otbr-agent-rest-discovery` only sends the REST API information to Core, which in turn connects to the REST API, but not the CLI unix socket.

PR #4417, opened 2026-02-12
otbr: Automatically migrate OTBR settings to last-good backup

puddly:
Current TLV parsing issues in Core appear to be the result of the "bad" `tmp` data file being loaded by the migration script and passed through to OTBR. It manages to start up and broadcast the broken 8 byte dataset, which breaks TLV parsing in Core and leaves the Thread network broken.

I've amended the migration script to run a data migration and revert to the last-known backup if we see that the current OTBR data file is invalid.

I'm not sure if this is acceptable but I've written a few unit tests using real data and test the migration script's behavior using them. These do not run as part of CI but they are in a `tests/` folder. Is this fine to include in the addons repo?


openthread_border_router/rootfs/usr/local/bin/migrate_otbr_settings.py
```diff
@@ -74,7 +96,86 @@ def serialize_otbr_settings(settings: list[tuple[OtbrSettingsKey, bytes]]) -> by
 
 def is_valid_otbr_settings_file(settings: list[tuple[OtbrSettingsKey, bytes]]) -> bool:
     """Check if parsed settings represent a valid OTBR settings file."""
-    return {OtbrSettingsKey.ACTIVE_DATASET} <= {key for key, _ in settings}
+    active_dataset = get_active_dataset(settings)
+    return active_dataset is not None and len(active_dataset) >= MIN_ACTIVE_DATASET_SIZE
```
agners, line 100: I guess instead of going by length we could just test for some usually available tags 🤔 

I'd go for these five, they are pretty much required:

  1. Network Key
  2. Network Name
  3. Extended PAN ID
  4. PAN ID
  5. Channel

agners (commented):
Generally looks good, I wonder if we have enough such cases which makes it worthwhile to add this complexity 🤔 

Not sure if the "Reset border router" command via Thread configuration in HA is possible in this situation, if not users could also just uninstall the add-on and reinstall it to clear the internal settings, then let the OTBR integration setup a "new" border router 🤔.

I guess we can also add this, and remove it after a couple releases to not have the long term maintenance burden of the complexity 🤷 .

PR #4428, opened 2026-02-16
fix(docs): Update DOCS.md with Thread creds sync instructions

tyeth:
Hey, I wasted a lot of time on setup today, because everything was fine except the credentials sync in the android companion app [hidden in troubleshooting].
That piece of information is available on the thread integration's docs page, but not the open thread border router, which is where the matter signup process fails [android says you need a thread border router - which you have already setup via HA].
Adding to the docs page of the OTBR app is the first place people will look.
It might make more sense to have the how-to-use section containing the OTBR/mobile  guidance, at least until the home assistant companion app can do the sync automatically as you add your network/HA-instance in the mobile app.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for synchronizing Thread network credentials across Android and iOS mobile devices. New documentation provides detailed step-by-step in-app navigation instructions specific to both platforms, helping users successfully configure Thread integration and transfer network credentials to their Android smartphones and iPhones.

### DIFF
--- a/openthread_border_router/DOCS.md
+++ b/openthread_border_router/DOCS.md
@@ -46,6 +46,19 @@ create a new integration named "Open Thread Border Router". With Home Assistant
 Core 2023.3 and newer the OTBR will get configured automatically. The Thread
 integration allows to inspect the network configuration.
 
+You will still need to go to the Thread settings page and select your preferred
+network, choose the option for **Use router for Android and iOS credentials**, 
+and then you can add devices using your mobile via the Home-Assistant companion
+app. **Don't forget to sync the credentials** for the Thread network with your
+companion app.
+
+On an Android phone in the HA app go to **Settings** > **Companion App** >
+**Troubleshooting** > **Sync Thread credentials**.
+
+Or for an iPhone, go to **Settings** > **Devices & services**, select the
+Thread integration, select **Configure** under Services. At the bottom of the
+preferred network box, select **Send credentials to phone**.
+
 ### Web interface (advanced)
 
 There is also a web interface provided by the OTBR. However, the web


agners:
Thank you for your contribution.

However, I'd prefer if we link to the [Thread integration documentation instead](https://www.home-assistant.io/integrations/thread/#to-make-home-assistant-your-first-thread-network), this will make it easier to update this documentation once it works differently.

tyeth:
Appreciate that, any improvement on the current situation would be good.
 I found the threads docs page too long [even this one], but to be honest
it feels a poor experience due to the companion app not syncing for you,
after that you're just scrambling for why the open thread border router
isn't found by Android.

Also reading that if the HA instance had external Bluetooth adapter then
could commission thread/matter devices itself without companion app, so why
can't it use the pi5 Bluetooth adapter for that [is binary blob use
restricted or just lacking driver/HA support]?

agners (changes requested):
> Appreciate that, any improvement on the current situation would be good. I found the threads docs page too long [even this one], but to be honest it feels a poor experience due to the companion app not syncing for you, after that you're just scrambling for why the open thread border router isn't found by Android.

We had the automatic sync at one point, but it was slow. Since WiFi based Matter devices and users of other BRs didn't require the sync, we opted that direction. But it is less then ideal, and we should reconsider.

I agree, the Thread docs are long. Ultimately, this is a work around for now. I still would prefer to not duplicate documentation here, but simply redirect users to the docs. But it should be prominent, since lots of users fall into it right now, so I'd suggest to add an [alert block](https://docs.github.com/de/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) at the top of the section:


> [!IMPORTANT]
> If you use the Android companion app, you have to synchronize the Thread credentials manually. Refer to the [Thread Case 1 documentation](https://www.home-assistant.io/integrations/thread/#case-1-making-home-assistant-your-first-thread-network).

Issue #4425, opened 2026-02-16
Let Encrypt: After update to 6.0, `Could not find a suitable TLS CA certificate bundle`


mateuszdrab:
### Describe the issue you are experiencing

I use Lets Encrypt with Step ACME Subordinate CA signed by my root CA.

Since update to v6, the renewal fails with  `OSError: Could not find a suitable TLS CA certificate bundle`

I have multiple HA instances and tried the upgrade on my test one, where I faced the above issue.

Hopefully, it's a small glitch after the major overhaul of the app.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Let's Encrypt

### What is the version of the app?

6.0

### Steps to reproduce the issue

Use custom ACME endpoint with the CA's certificate provided in `acme_root_ca_cert` and endpoint in `acme_server`

### System Health information

N/A

### Anything in the Supervisor logs that might be useful for us?

```txt

```

### Anything in the app logs that might be useful for us?

```txt
[17:01:12] INFO: Detecting existing certificate type for <REDACTED_DOMAIN>
Saving debug log to /var/log/letsencrypt/letsencrypt.log
[17:01:13] INFO: Existing certificate using 'ecdsa' key type.
Root logging level set at 0
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Requested authenticator standalone and installer None
Single candidate plugin: * standalone
Description: Runs an HTTP server locally which serves the necessary validation files under the /.well-known/acme-challenge/ request path. Suitable if there is no HTTP server already running. HTTP challenge only (wildcards not supported).
Interfaces: Authenticator, Plugin
Entry point: EntryPoint(name='standalone', value='certbot._internal.plugins.standalone:Authenticator', group='certbot.plugins')
Initialized: <certbot._internal.plugins.standalone.Authenticator object at 0x7efda726b0e0>
Prep: True
Selected authenticator <certbot._internal.plugins.standalone.Authenticator object at 0x7efda726b0e0> and installer None
Plugins selected: Authenticator standalone, Installer None
[...]

PR #4430, opened 2026-02-17
letsencrypt: Fix issuing certificates when `acme_root_ca_cert` is set

Cethric:
After the migration from Alpine Linux to Debian Linux the path to the CA bundle changed

#4425

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSL/TLS certificate validation by updating the system's certificate bundle configuration to ensure proper certificate verification across different environments.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.0.3
+
+- Fix issuing certificates when `acme_root_ca_cert` is set
+
 ## 6.0.2
 
 - Fix propagation seconds not being correctly applied to certbot-dns-multi providers
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 6.0.2
+version: 6.0.3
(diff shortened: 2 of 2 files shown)


letsencrypt/rootfs/etc/services.d/lets-encrypt/run
```diff
@@ -547,7 +547,7 @@ if bashio::config.has_value 'acme_server'; then
           | sed 's/----- /-----\n/g' | sed 's/ -----/\n-----/g' \
           | openssl x509  > /usr/local/share/ca-certificates/acme_root_ca.crt \
           && update-ca-certificates --fresh \
-          && export REQUESTS_CA_BUNDLE=/etc/ssl/cert.pem
+          && export REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
```
sairon: We can just remove it here, as the Debian image already has this environment default which is also consumed by `requests`: https://github.com/home-assistant/docker-base/blob/3a3190763da59db5973c35df7148f4dc2c5da969/debian/Dockerfile#L11

Cethric: Thanks for the update, I am still finding my way around the project.

PR #4438, opened 2026-02-18
letsencrypt: add check for `dns.provider`, cleanup build cache files

### LINKED ISSUE #4436: letsencrypt 'dns.provider' is sort of mandatory but too easy to overlook

I upgraded to 6.0.0 and cheerfully configured via the GUI.  The resulting YAML is:

```yaml
keyfile: privkey.pem
certfile: fullchain.pem
challenge: dns
dns:
  lego_provider: hurricane
  lego_env:
    - HURRICANE_TOKENS=[redacted]
domains:
  - [redacted]
email: [redacted]
dry_run: true
```

Spot the error?  The logs are supremely unhelpful because the script actually almost works when configured like this, except I get an inscrutable error about:

`hurricane: some credentials information are missing: HURRICANE_TOKENS`

But HURRICANE_TOKENS is there, and it's even correct!

The issue is that `dns.provider` is missing, and the script does not notice and even makes it through part, but not all, of the lego configuration.  The failure mode is that the dns-multi configuration selects `hurricane` but entirely ignores `lego_env`.  Whoops.  If I set `dns.provider`, then it works.

IMO there should be a better error message about the incorrect configuration, and maybe `provider` could be moved in the UI so that, instead of being in the middle of a giant list of options, it's somewhere obvious.  And maybe it should even default to lego.

Home Assistant OS

CEC Scanner

6.0.2

1. Configure the addon to use lego and set a lego provider and the relevant environment.
2. Fail to see `provider` in the middle of the giant option list.
3. Run it and get very confused by the failure.
4. Read the source code to see that processing `lego_env` is conditional on `dns.provider` being set correctly even though handling `lego_provider` is not.

### System Health information

it's all good.

```txt

```

```txt

```

### Additional information

_No response_

sairon:
Fail if DNS challenge is selected and the provider is not configured.

Also clean up the cache files created by pip and Go build to significantly reduce image bloat.

Fixes #4436

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS challenge now requires a DNS provider to be configured, preventing misconfiguration.

* **Chores**
  * Cleaned up build/cache files to reduce image size.
  * Release version bumped to 6.0.4.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 6.0.4
 
 - Require `dns.provider` if DNS challenge is set
-- Cleanup cache files from the build to reduce image size
+- Clean up cache files from the build to reduce image size
 
 ## 6.0.3
 


WaizenErnter:
Hey, I installed Let's Encrypt 6.0.4 but the filed DNS Provider wasn't required to save the config/start the Addon. Lets encrypt just brought up an error in the log about dns.provider missing. I think the config field should be marked as required.

PR #4453, opened 2026-02-23
letsencrypt: Add force_renew option

mateuszdrab:
Adds a `force_renew` boolean config option to the Let's Encrypt addon that passes `--force-renewal` to certbot, forcing certificate renewal regardless of expiry.

This is copilot-ed - I hope that's not an issue.

I need this flag to invoke a forced renewal of the certificates, this is due to change of the signing certificate of my private ACME server which expired; however, normal renewal attempt doesn't detect that and therefore the pem files are not updated with the new signing cert.

## Changes
- **`config.yaml`**: Added `force_renew: bool?` to schema; bumped version to `6.0.5`
- **`run`**: Reads `force_renew` and appends `--force-renewal` to `ADDITIONAL_ARGS`; applies unconditionally (both default LE and custom ACME servers)
- **`translations/en.yaml`**: Added UI label/description for `force_renew`

When enabled, certbot is invoked with:
```
certbot certonly ... --force-renewal ...
```

---

🔒 GitHub Advanced Security automatically protects Copilot coding agent pull requests. You can protect all pull requests by enabling Advanced Security for your repositories. [Learn more about Advanced Security.](https://gh.io/cca-advanced-security)

## Summary by CodeRabbit

* **New Features**
  * Added a "Force Renew" option to force certificate renewal even if not near expiry; available alongside existing renewal options.

* **Documentation**
  * Updated docs with usage examples and a note to reset the option after renewal.

* **Translations**
  * Added translatable name and description for the "Force Renew" option.

* **Changelog**
  * Added 6.1.0 entry documenting the new option.

### DIFF
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 6.0.5
+version: 6.1.0
 breaking_versions: [5.3.0, 6.0.0]
 slug: letsencrypt
 name: Let's Encrypt
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.1.0
+
+- Add `force_renew` option to force certificate renewal regardless of expiry date
+
 ## 6.0.4
 
 - Require `dns.provider` if DNS challenge is set
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -228,6 +228,14 @@ automation every night via the [Supervisor app restart action][supervisorrestart
 In this example, the automation will run every day at the chosen time, checking
 if a renewal is due, and will request it if needed.
 
+To force a certificate renewal regardless of the expiry date, set the `force_renew` option to `true`:
+
+```yaml
+force_renew: true
+```
+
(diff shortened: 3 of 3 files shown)


letsencrypt/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 6.0.4
+version: 6.0.5
```
sairon: `6.1.0` will be semantically better, as you're adding a new feature, not making a fix.

sairon (changes requested):
Generally LGTM, but also please update the `CHANGELOG.md` and document the option in `DOCS.md` in the **Create & renew certificates** section.

letsencrypt/config.yaml
mateuszdrab: All done, thanks for your comments

GoetzGoerisch:
Interesting to see that the maintainers have now decided to include such an option after it was REJECTED a year ago.
Although the potential missconfiguration is still an issue.

#3976

mateuszdrab:
Oh wow! 
I have some ideas how to improve this so that force option does not stay on forever.
To be honest, I don't see why your PR didn't make it in. This option is very handy with custom ACME CAs.

Looks like most of the options from your PR made it into the add-on anyway.

PR #4473, opened 2026-03-04
deCONZ: Drop unsupported architectures

mattreim:
Support for the armhf, armv7, and i386 architectures has been removed. The `ingress.html` file was automatically formatted, but I don't think a new version is necessary.

## Summary by CodeRabbit

* **Chores**
  * Bumped version to 8.7.0
  * Dropped support for 32-bit ARM and 32-bit x86 architectures
  * Simplified architecture-specific package configuration; only amd64 and aarch64 are now supported

### DIFF
--- a/deconz/CHANGELOG.md
+++ b/deconz/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.7.0
+
+- Drop unsupported architectures
+
 ## 8.6.1
 
 - Align ingress colors with Home Assistant (toolbar color)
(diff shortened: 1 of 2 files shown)


agners (changes requested):
> The `ingress.html` file was automatically formatted, but I don't think a new version is necessary.

But there are changes... 🤔  I'd rather prefer if you undo the change, so its not being touched in this PR.

Other than that, LGTM!

mattreim:
No problem, the formatting has been reset.

PR #4478, opened 2026-03-08
matter-server: Allow ENV vars and matterjs-server versions

Apollon77:
This PR enhances the Matter server addon to allow specifying ENV vars (to tweak matterjs internal settings and allow pre-tests of features) and also allows to specify a matterjs version to install in case of issues but guards that  for 0.x and 1.x (ok we are still at 1.x but to be a bit future proof)

## Summary by CodeRabbit

* **New Features**
  * Release 8.3.0: add matter_server_env_vars to export custom environment variables before starting the JavaScript/Beta Matter Server.
  * matter_server_version now respected in JavaScript/Beta mode for 0.x.y and 1.x.y values (other values fall back to latest).

* **Documentation**
  * Updated docs with usage examples and guidelines for the new options.

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -104,7 +104,7 @@ jobs:
 
       - name: Login to DockerHub
         if: env.BUILD_ARGS == '--docker-hub-check'
-        uses: docker/login-action@v3.7.0
+        uses: docker/login-action@v4.0.0
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
--- a/deconz/CHANGELOG.md
+++ b/deconz/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.7.0
+
+- Drop unsupported architectures
+
 ## 8.6.1
 
 - Align ingress colors with Home Assistant (toolbar color)
--- a/deconz/Dockerfile
+++ b/deconz/Dockerfile
@@ -41,10 +41,7 @@ RUN \
 ARG DECONZ_VERSION
 RUN \
     set -x \
-    && if [ "${BUILD_ARCH}" = "armhf" ]; \
-        then \
-            curl -q -L -o /deconz.deb http://deconz.dresden-elektronik.de/raspbian/stable/deconz-${DECONZ_VERSION}-qt5.deb; \
-        elif [ "${BUILD_ARCH}" = "aarch64" ]; \
+    && if [ "${BUILD_ARCH}" = "aarch64" ]; \
         then \
             curl -q -L -o /deconz.deb http://deconz.dresden-elektronik.de/debian/stable/deconz_${DECONZ_VERSION}-debian-buster-stable_arm64.deb; \
         else \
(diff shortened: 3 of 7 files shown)


matter_server/translations/en.yaml
```diff
@@ -26,6 +26,12 @@ configuration:
       Home Assistant Companion app commissioning method is recommended as it is
       better tested and allows to commission directly in proximity of the
       device itself.
+  matter_server_env_vars:
```
agners, line 29: Is there an advantage of using environment variable instead of passing arguments like we did with the Python server? 🤔

Apollon77: The idea is to have cli for the basic config. Matter.js allows to tweak a lot of its internals via env vars too. But these are more pro use cases but can be handy for tests by power users. Duplicating all these (or a subset) as matter server CLI options does not make too much sense for me. So I would keep cli for common case but allow using env for advanced cases.

agners (approved):
LGTM!

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 8.2.3
```
agners: Noticed just before merge, this adds new options, I think it warrants a minor version bump:


```suggestion
## 8.3.0
```

matter_server/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 8.2.2
+version: 8.2.3
```
agners:
```suggestion
version: 8.3.0
```

matter_server/config.yaml
```diff
@@ -28,12 +28,15 @@ options:
...
   enable_test_net_dcl: false
+  matter_server_env_vars: []
 schema:
   log_level: list(verbose|debug|info|warning|error|critical)
   log_level_sdk: list(automation|detail|progress|error|none)?
   beta: bool?
   enable_test_net_dcl: bool?
   bluetooth_adapter_id: int?
+  matter_server_env_vars:
+    - match(^[A-Za-z_][A-Za-z0-9_]*=.*$)?
```
agners: Maybe after `matter_server_args` so it's alphabetically, or at the end.

matter_server/config.yaml
```diff
@@ -28,12 +28,15 @@ options:
   log_level_sdk: error
   beta: false
   enable_test_net_dcl: false
+  matter_server_env_vars: []
```
agners: I don't think this is required for an optional field 🤔

matter_server/config.yaml
Apollon77: Updated in 0571b1f: addon version is now set to 8.3.0 in config.yaml.

matter_server/config.yaml
Apollon77: Adjusted in 0571b1f: removed matter_server_env_vars from default options since it is optional.

matter_server/CHANGELOG.md
Apollon77: Updated in 0571b1f: release bump is now 8.3.0 in the changelog heading.

matter_server/config.yaml
Apollon77: Adjusted in 0571b1f: moved matter_server_env_vars to after matter_server_args in the schema.

PR #4480, opened 2026-03-09
Allow spaces in configured Samba username

madbrain76:
Windows user names can have spaces, and mine does. The fix is to optionally create a username map. I tested this fix.

## Summary by CodeRabbit

* **New Features**
  * Usernames can now include spaces for more descriptive account naming. The system automatically manages all necessary configuration and internal mapping to ensure seamless authentication, maintain reliable access control, and enable proper user management regardless of spaces appearing in the username.

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 12.6.1
 
-- Always create the Samba username map, using an empty file unless the configured username contains spaces
+- Support spaces in configurable username
 
 ## 12.6.0
 


samba/rootfs/etc/s6-overlay/s6-rc.d/init-smbd/run
```diff
@@ -37,14 +40,30 @@ fi
...
-jq ".interfaces = $(jq -c -n '$ARGS.positional' --args -- "${interfaces[@]}") |
-    .enabled_shares.[] |= ascii_downcase" /data/options.json \
+# Use an internal Samba username when the configured username contains spaces.
+config_username=$(bashio::config 'username')
+samba_username=$(printf '%s' "${config_username}" | tr '[:space:]' '_')
+if [[ "${config_username}" =~ [[:space:]] ]]; then
+    username_has_space=true
+    {
+        printf '%s = \"%s\"\n' "${samba_username}" "${config_username//\"/\\\"}"
+    } >/etc/samba/smbusers
```
agners, line 49: Is there a downside if we'd just always use a map? 🤔 Would make the script a bit cleaner.

madbrain76: > Is there a downside if we'd just always use a map? 🤔 Would make the script a bit cleaner.

I just tested this, and there is no downside. An empty map works.

madbrain76:
> Also update the version in `config.yaml` and update the `CHANGELOG.md` file accordingly.

For config.yaml, I tried, but this breaks the installation from /addons for me, because HA tries to pull the new 12.6.1 image from Docker, and that image doesn't exist yet.

agners:
> For config.yaml, I tried, but this breaks the installation from /addons for me, because HA tries to pull the new 12.6.1 image from Docker, and that image doesn't exist yet.

Yeah for local test the easiest is to temporary remove the `image` tag, then Supervisor will build the app.

I see that you did adjust the PR accordingly, so this should be fine for the PR.

samba/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.6.1
+
+- Always create the Samba username map, using an empty file unless the configured username contains spaces
```
agners: The user visible change here is really that we support spaces in the configured username, so:


```suggestion
- Support spaces in configurable username
```

agners (changes requested):

PR #4491, opened 2026-03-14
Add Windows automatic discovery support with WSDD

madbrain76:
This patch allows the HAOS system to automatically show up in the "Network" tab on Windows hosts.

To minimize impact, I chose to make the the new service startup failure non-fatal. It will just log a warning in this case. I can change it back. I could make it fatal if that's preferred.

## Summary by CodeRabbit

* **New Features**
  * Added Windows automatic discovery support (WSDD) for seamless device discovery on Windows networks.

* **Chores**
  * Included WSDD in the runtime image and startup services.
  * Bumped app version to 12.7.0.

* **Stability**
  * Improved service lifecycle handling so WSDD termination won’t cause the container to fail unexpectedly.
  * Added network interface detection with clear warnings when no interfaces are enabled.


samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/finish
```diff
@@ -0,0 +1,25 @@
...
+readonly service="wsdd"
+
+bashio::log.info \
+  "Service ${service} exited with code ${exit_code_service}" \
+  "(by signal ${exit_code_signal})"
+
+if [[ "${exit_code_service}" -eq 256 ]]; then
+  if [[ "${exit_code_container}" -eq 0 ]]; then
+    echo $((128 + exit_code_signal)) > /run/s6-linux-init-container-results/exitcode
+  fi
```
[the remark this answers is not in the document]
agners: Yeah I think I agree with this.

samba/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 12.6.1
+version: 12.6.2
```
agners: Let's make this a minor bump, since it adds a new feature.

```suggestion
version: 12.7.0
```

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/run
```diff
@@ -0,0 +1,36 @@
...
+for interface in $(bashio::network.interfaces); do
+    interface_enabled=$(bashio::network.enabled "${interface}")
+    if bashio::var.true "${interface_enabled}"; then
+        interfaces+=("-i" "${interface}")
+    fi
+done
+
+if [ ${#interfaces[@]} -eq 0 ]; then
+    bashio::log.warning 'WSDD disabled: no supported interfaces found to bind on.'
+    exec tail -f /dev/null
```
agners: Hm, this probably can't really happen since `init-smbd` which is a dependency of this service bails out with `bashio::exit.nok`. I wonder if we can reuse the interface list from there 🤔 . 

What we could do is exit with an exit code not used by `wsdd` itself, and then exit the finish script with 125 (permanent failure) if that exit code was returned by the run script.

agners (changes requested):

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/run
madbrain76: > Hm, this probably can't really happen since `init-smbd` which is a dependency of this service bails out with `bashio::exit.nok`. I wonder if we can reuse the interface list from there 🤔 .
> 
> What we could do is exit with an exit code not used by `wsdd` itself, and then exit the finish script with 125 (permanent failure) if that exit code was returned by the run script.

Are you sure ? I think it might be possible, if you have other apps listening on some but not all relevant ports. For example, if one is mixing the official Samba share add-on with others (such as forks) accidentally, as I had been doing while testing, trying to decide which upstream to feed my changes. It's more likely that it would fail binding samba first, though. But in theory one could have a completely standalone wsdd add-on also, and it could fail on the second bind, too.

agners: > Are you sure ? I think it might be possible, if you have other apps listening on some but not all relevant ports.

Well, then the service would fail. What I meant here is this if statement. Unless there is really an interface change between the time `init-smbd` and this service comes up, it will be the same.

Maybe `init-smbd` should simply be named `init-samba`, and then we create a default file we can source in `wsdd` service, so we avoid code duplication :thinking:

madbrain76: You are right. I decided to refactor this logic, but without renaming init-smbd to init-samba .

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/finish
```diff
@@ -0,0 +1,29 @@
...
+
+bashio::log.info \
+  "Service ${service} exited with code ${exit_code_service}" \
+  "(by signal ${exit_code_signal})"
+
+if [[ "${exit_code_service}" -eq 256 && "${exit_code_signal}" -eq 15 ]]; then
+  if [[ "${exit_code_container}" -eq 0 ]]; then
+    echo $((128 + exit_code_signal)) > /run/s6-linux-init-container-results/exitcode
+  fi
+  exec /run/s6/basedir/bin/halt
```
agners, line 22: This part feels a bit overengineered to me, but I am fine with it. I mostly meant we should not shutdown the container on regular failure, but what coderabbit suggested was a bit more elaborate 😅

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/finish
```diff
@@ -0,0 +1,29 @@
...
+elif [[ "${exit_code_service}" -eq 256 ]]; then
+  bashio::log.warning \
+    "Optional service ${service} exited by signal ${exit_code_signal}; continuing without Windows discovery."
```
agners, line 25: In this...

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/finish
```diff
@@ -0,0 +1,29 @@
...
+elif [[ "${exit_code_service}" -ne 0 ]]; then
+  bashio::log.warning \
+    "Optional service ${service} failed; continuing without Windows discovery."
```
agners, line 29: ... and this case s6 will simply restart the service again. If we want it to remain stopped we need to exit the finish script with exit code `125` (permanent failure). And I'd make use of that here, since that is what the warning suggests is what we want to happen.

samba/rootfs/etc/s6-overlay/s6-rc.d/wsdd/run
agners: With the exit code 125 above in place, we can use `exit 1` here. This will simply stop the service with permanent failure. No `exec tail -f /dev/null` workaround required.

agners (changes requested):

madbrain76:
Is there something left to fix on this PR ? It looks like I accidentally had left it in draft mode.

madbrain76:
Gentle ping @agners  . Please let me know if you need any more changes before landing this, and I'll be happy to make them.

agners (approved):
LGTM, thank you for your patience!

madbrain76:
Thank you !

Issue #4483, opened 2026-03-10
LE: Cert request via `dns-lego` for provider `ArtFiles` fails


Eagle3386:
### Describe the issue you are experiencing

DNS provider ArtFiles isn't recognized by `dns-lego` although it should be as described below.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Let's Encrypt

### What is the version of the app?

6.1.1

### Steps to reproduce the issue

1. Configure for provider `artfiles`
2. Start the app.
3. Switch to `Log`-tab.

### System Health information

Clicking `Copy` stated via toast it copied the information, but the clipboard remained filled with other data (app's log, see below).
So, here's my manual copy-paste-output:

Version	core-2026.3.1
Installation type	Home Assistant OS
Development	false
Supervisor	true
Docker	true
Container architecture	aarch64
User	root
Virtual environment	false
Python version	3.14.2
Operating system family	Linux
Operating system version	6.12.47-haos-raspi
CPU architecture	aarch64
Timezone	Europe/Berlin
Configuration directory	/config
Home Assistant Community Store
GitHub API	ok
GitHub Content	ok
GitHub Web	ok
HACS Data	ok
GitHub API Calls Remaining	5000
Installed Version	2.0.5
Stage	running
Available Repositories	2877
Downloaded Repositories	11
Home Assistant Cloud
Logged In	false
Reach certificate server	ok
Reach authentication server	ok
Reach Home Assistant Cloud	ok
Home Assistant Supervisor
Host operating system	Home Assistant OS 17.1
Update channel	stable
Supervisor version	supervisor-2026.02.3
Agent version	1.8.1
Docker version	29.1.3
Disk total	28.0 GB
Disk used	17.3 GB
Nameservers	10.x.x.x
Healthy	true
Supported	true
host_connectivity	true
supervisor_connectivity	true
ntp_synchronized	true
virtualization	
Board	yellow
Supervisor API	ok
Version API	ok
Installed apps	Let's Encrypt (6.1.1), Studio Code Server (6.0.1), ESPHome Device Builder (2026.2.4), openWakeWord (2.1.0), Piper (2.2.2), Whisper (3.1.0), Grocy (0.24.0), Terminal & SSH (10.0.2), File editor (5.8.0), Silicon Labs Flasher (0.4.0)
Dashboards
Dashboards	3
Resources	0
Views	1
Mode	storage
[...]

PR #4497, opened 2026-03-17
letsencrypt: Update CertBot-DNS-Multi to 4.32.0

Eagle3386:
Update `dns-lego` to add DNS provider `ArtFiles`.

Fixes: #4483

## Summary by CodeRabbit

* **New Features**
  * Added support for multiple new DNS providers (including ArtFiles, alidns line-record, Bluecat v2, DDNSS, DNSExit, FusionLayer NameSurfer, Leaseweb, TodayNIC).

* **Chores**
  * Upgraded DNS plugin tooling and ACME library to broader provider support and compatibility.

* **Documentation**
  * Published changelog entry for version 6.2.0 and bumped addon configuration version.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -2,15 +2,7 @@
 
 ## 6.2.0
 
-- Update certbot-dns-multi to 4.32.0 (see [lego's changelog](https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md) for details), most notably to add these DNS providers:
-  - alidns (line record option)
-  - ArtFiles
-  - Bluecat v2
-  - DDNSS
-  - DNSExit
-  - FusionLayer NameSurfer
-  - Leaseweb
-  - TodayNIC/时代互联
+- Update certbot-dns-multi to 4.32.0, adding several DNS providers (see [lego's changelog](https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md) for details).
 
 ## 6.1.1
 


Eagle3386:
> Hi, thank you for your contribution. This needs a version bump in `config.yaml` and an update to `CHANGELOG.md`.

Hi & thanks for the reminder - totally forget that, sorry about that!

Here's hope, my "introduction" of [semantic versioning](https://semver.org) is ok!?
As a user, I always got confused why some additions increased the minor version while others of equal impact resulted in an increase of the patch version number - using SemVer solves that, but I'm almost certain that you knew that pattern already. 😅

agners:
> > Hi, thank you for your contribution. This needs a version bump in `config.yaml` and an update to `CHANGELOG.md`.
> 
> Hi & thanks for the reminder - totally forget that, sorry about that!
> 
> Here's hope, my "introduction" of [semantic versioning](https://semver.org) is ok!? As a user, I always got confused why some additions increased the minor version while others of equal impact resulted in an increase of the patch version number - using SemVer solves that, but I'm almost certain that you knew that pattern already. 😅

Yeah we more or less follow semantic versioning. Sometimes previous precedence takes over (like adding a new platform we've let through as a patch since it was always like that and there were so many).

At times it is also a question of interpretation: As an App, if the base OS changes, what type of bump is that? :thinking:  There is no visible functional change for the user.

But anyhow, I am fine using minor here, since that is what certbot multi used.

letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 6.2.0
+
+- Add ArtFiles DNS provider support
+- Update [lego ACME library](https://go-acme.github.io/lego/) to latest
```
sairon: Neither of these is entirely true. The ArtFiles DNS support is only added through the Lego library, not explicitly like other providers, so the first line can be confusing. Also there are many more providers added via this release.

For the second one, it's a transitive dependency, I'd stay consistent with older release notes:
```suggestion
- Update certbot-dns-multi to 4.32.0
```

agners: This is very not helpful for users, IMHO. Can we maybe at least link to useful changelogs?

agners:
```suggestion
- Update certbot-dns-multi to 4.32.0 (refer to [Lego changelog](https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md) for list of changes)
```

sairon: I see two minor issues there:
- `certbot-dns-multi` not always maps to Lego releases 1:1
- Not all changes in the Lego apply to certbot-dns-multi (as large part of the certificate issue logic is handled by Certbot and not Lego)

Also, if you check older release notes, this establishes a pattern we've never used for dependency bumps.

But eventually creating the changelog should be automated and cross-linked with the PRs, so this shouldn't be a burden for long, so I don't care much.

agners: > * `certbot-dns-multi` not always maps to Lego releases 1:1
>  * Not all changes in the Lego apply to certbot-dns-multi (as large part of the certificate issue logic is handled by Certbot and not Lego)

Hm, yeah I see that sucks.

> But eventually creating the changelog should be automated and cross-linked with the PRs, so this shouldn't be a burden for long, so I don't care much.

Right, then of course the discussion becomes what should the PR include... In git convention is what and why (never how), so we'd likely have both captured, the bump as well as why that bump got made :thinking: :shrug:. We'll definitely have PR which just mentions the provider the author is interested in, and then bumps `certbot-dns-multi`. Will we accept that then? :thinking:

agners: As for this PR, 

```suggestion
- Update certbot-dns-multi to 4.32.0 (adds ArtFiles DNS provider support among others)
```

Eagle3386: @sairon
1. The mapping has nothing to do, it's simply a coincidence - and Stefan's suggestion of linking to `lego`'s release notes is even better than my original approach (I originally linked to the `lego` project just to provide a "hint" to go there for the changelog).
2. I _strongly_ disagree with boiling it down to "Updated certbot-dns-multi to x.y.z" - as a user, each single time I'm confused as well as left with almost useless information when updating the `letsencrypt` HA app as providing "updated dependency x" does _not_ provide users with _why_ an updated is useful - for the purpose of my PR, the solely reason is to get my DNS provider, ArtFiles, finally supported (I'm 1 of 2 people who worked on the support). But I'd _happily_ add all the other, newly supported providers as well. Because, after all, this "app's" solely purpose is to enable LE usage on HA systems, hence added DNS providers are one of _the_ most notable changes a release should point out.
3. Who does the "cert issue logic" (and even that for renew) is irrelevant, it's `lego` who enabled broader DNS provider support & hopefully allows for a less catastrophic UI design of an almost endlessly scrollable configuration input field list. Again, this is _not_ a "dependency bump", it's a "feature bump" & given the quite big amount of newly supported DNS providers, it more than justifies an increase of the minor version.

The only thing I agree with you on is your suggestion of an automated changelog - it should, for example, simply include `lego`'s added/changed/deleted/fixed DNS providers & modify `letsencrypt`'s version accordingly as well, e.g., added/changed providers increase the minor, deleted ones increase the major & fixed just increment the patch version number.

So, for now I go ahead & implement Stefan's suggestion of linking to `lego`s changelog & announce all added DNS providers as that's what probably _most_ users like to know first.

letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 6.2.0
+
+- Update certbot-dns-multi to 4.32.0 (see [lego's changelog](https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md) for details), most notably to add these DNS providers:
+  - alidns (line record option)
```
sairon: This is not new provider, this is new option for Alibaba Cloud DNS.

letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## 6.2.0
+
+- Update certbot-dns-multi to 4.32.0 (see [lego's changelog](https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md) for details), most notably to add these DNS providers:
```
sairon: The bump is from 4.30.1 so the list below is missing all providers added in v4.31.0: https://github.com/go-acme/lego/blob/v4.32.0/CHANGELOG.md#v4310

Eagle3386:
> I understand that the high-level view "certbot-dns-multi updated" may be too vague, however, I don't want to duplicate upstream changelog here either - below you can see why: it's easy to get it wrong and it creates unnecessary burden. In the end, if you're updating an already installed app, you most often don't really care _what providers were added_ because you're already using some in your setup. You care about that when you want to install it, so it should be easy to get that information from docs/readme (again, ideally through a link, so you don't need to keep it in sync).
> 
> In the changelog I'll prefer sticking to a reference instead of extensive lists.

I'll give up. Just accept the once-more-updated PR, so I can use this "app" as my DNS provider is _finally_ supported then & we're done. Thanks.

sairon:
@Eagle3386 Sorry for the frustration. We plan on changing the whole way how the official apps are maintained and released, so I hope this will be less frustrating for all parties in the future.

PR #4500, opened 2026-03-18
Improve firewall and NAT64 rules / enable RCP recovery in beta

### LINKED ISSUE #4488: OTBR finish script does not clean up NAT64 iptables rules, causing duplicates on restart

The `otbr-agent` run script adds IPv4 iptables rules when NAT64 is enabled in the addon configuration (`bashio::config.true 'nat64'`), but the finish script never removes them. Every OTBR restart adds another copy of these rules.

The run script (`/etc/s6-overlay/s6-rc.d/otbr-agent/run`) adds these IPv4 iptables rules when NAT64 is enabled:

```bash
iptables -t mangle -A PREROUTING -i ${thread_if} -j MARK --set-mark 0x1001
iptables -t nat -A POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
iptables -t filter -A FORWARD -o ${backbone_if} -j ACCEPT
iptables -t filter -A FORWARD -i ${backbone_if} -j ACCEPT
```

The finish script (`/etc/s6-overlay/s6-rc.d/otbr-agent/finish`) only cleans up the ip6tables chains (`OTBR_FORWARD_INGRESS`, `OTBR_FORWARD_EGRESS`) and ipsets. It does not remove the IPv4 mangle/nat/filter rules.

Every OTBR restart adds another copy of these 4 rules. After N restarts, there are N duplicate copies.

On systems using `iptables-nft`, the accumulated duplicate rules increase the nftables ruleset size. Each rule addition triggers an atomic nftables table replacement, which can cause brief FIB routing cache corruption on the host (observed as `broadcast` route cache entries for external IPs, leading to `ENETUNREACH` errors for other services on the same host).

The finish script should remove the NAT64 iptables rules, similar to how it removes the ip6tables chains. The run script would need to store `backbone_if` (e.g. in `/tmp/otbr-backbone-if`) so the finish script can read it, since `backbone_if` is not available in the finish script's scope. Then add cleanup logic:

```bash
# Clean up NAT64 iptables rules (backbone_if read from /tmp/otbr-backbone-if written by run)
if [ -f /tmp/otbr-backbone-if ]; then
    backbone_if=$(cat /tmp/otbr-backbone-if)
    while iptables -t mangle -C PREROUTING -i $thread_if -j MARK --set-mark 0x1001 2>/dev/null; do
        iptables -t mangle -D PREROUTING -i $thread_if -j MARK --set-mark 0x1001
    done
    while iptables -t nat -C POSTROUTING -m mark --mark 0x1001 -j MASQUERADE 2>/dev/null; do
        iptables -t nat -D POSTROUTING -m mark --mark 0x1001 -j MASQUERADE
    done
    while iptables -t filter -C FORWARD -o $backbone_if -j ACCEPT 2>/dev/null; do
        iptables -t filter -D FORWARD -o $backbone_if -j ACCEPT
    done
    while iptables -t filter -C FORWARD -i $backbone_if -j ACCEPT 2>/dev/null; do
        iptables -t filter -D FORWARD -i $backbone_if -j ACCEPT
    done
    rm -f /tmp/otbr-backbone-if
fi
```

Home Assistant OS

OpenThread Border Router

2.16.5

1. Enable NAT64 in the OpenThread Border Router addon configuration
2. Restart the addon multiple times
3. Run `iptables-save -t mangle` and observe duplicate `PREROUTING -i wpan0 -j MARK --set-mark 0x1001` rules

```txt

```

```txt

```

### Additional information

_No response_

agners:
## Summary                                                                       
                                                                                 
- Replace broad `FORWARD` iptables rules that accepted all traffic on the infra interface with a dedicated `OTBR_FORWARD_NAT64` chain. Outbound Thread traffic is matched by fwmark, return traffic is scoped to established connections only.
- Add proper teardown of NAT64 iptables rules in `finish` to prevent rule accumulation on container restart (fixes home-assistant/addons#4488).
- Use `${}` style and quoting for variables throughout, resolving all shellcheck exceptions.
- Enable spinel recovery for beta builds.
                                                                                 
## Details      
                                                                                 
The previous NAT64 rules appended directly to the `FORWARD` chain with blanket `-i`/`-o` accepts on the infra interface. This was overly broad (allowed all forwarded traffic through that interface, not just Thread-originated) and had no cleanup path — every restart added duplicate rules.

The new approach mirrors the existing ip6tables pattern: a named chain (`OTBR_FORWARD_NAT64`) with an unconditional jump from `FORWARD`. Inside the chain, outbound traffic is identified by the fwmark set in `mangle/PREROUTING` on the Thread interface, and return traffic is accepted only for established connections via conntrack. Unmatched packets fall through to the rest of `FORWARD` via implicit `RETURN`.

## Summary by CodeRabbit

* **New Features**
  * Beta: enabled automatic recovery from radio transmission timeouts

* **Bug Fixes**
  * Fixed and tightened NAT64 firewall handling with more thorough teardown
  * Narrowed non-firewall forwarding to only Thread and infrastructure interfaces
  * Improved restoration/retry handling to reduce RCP restore failures and improve startup/firewall reliability

* **Chore**
  * Version bumped to 2.16.6

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,6 @@ name: Lint
 
 env:
   HADOLINT_VERSION: v1.17.2
-  SHELLCHECK_OPTS: -e SC1008 -s bash
 
 on:
   pull_request:
@@ -50,6 +49,8 @@ jobs:
 
       - name: Run linter
         uses: ludeeus/action-shellcheck@2.0.0
+        env:
+          SHELLCHECK_OPTS: -e SC1008 -s bash -x
         with:
           # Names of scripts with custom shebangs that the action would not
           # discover on its own:
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.16.6
 
+- Fix and improve NAT64 firewall rules
 - Enable recovery mechanism from "radio tx timeout" errors for beta
 
 ## 2.16.5
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.16.6
+
+- Enable recovery mechanism from "radio tx timeout" errors for beta
+
 ## 2.16.5
 
 - Add `backbone_interface` option to override the network interface used for IPv6 routing.
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix and improve NAT64 firewall rules
 - Enable recovery mechanism from "radio tx timeout" errors for beta
+- Narrow non-firewall forwarding rules to Thread and infrastructure interface
 
 ## 2.16.5
 
--- a/openthread_border_router/CHANGELOG.md
+++ b/openthread_border_router/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Fix and improve NAT64 firewall rules
 - Enable recovery mechanism from "radio tx timeout" errors for beta
-- Narrow non-firewall forwarding rules to Thread and infrastructure interface
+- Narrow non-firewall forwarding rules to Thread interface
 
 ## 2.16.5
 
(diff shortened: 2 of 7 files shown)


agners:
Tested with a tado Thermostat X Matter over Thread device, the device appears on the tado App (which connects through the cloud to the device).

puddly (approved):

PR #4510, opened 2026-03-24
Enhance README with detailed openWakeWord information

slgotting:
Expanded the README to include detailed information about wake word detection, its role in the voice pipeline, common uses, installation instructions, and features of openWakeWord.

### DIFF
--- a/openwakeword/README.md
+++ b/openwakeword/README.md
@@ -1,12 +1,162 @@
-# Home Assistant App: openWakeWord
+# openWakeWord – Home Assistant App
 
-![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield]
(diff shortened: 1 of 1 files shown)


openwakeword/README.md
```diff
@@ -1,12 +1,162 @@
-# Home Assistant App: openWakeWord
+# openWakeWord – Home Assistant App
 
-![Supports aarch64 Architecture][aarch64-shield] ![Supports amd64 Architecture][amd64-shield]
+**openWakeWord** adds wake-word detection to Home Assistant’s voice system. It listens to microphone audio and triggers the voice pipeline when a chosen phrase (the *wake word*) is spoken.
```
agners, line 3: I don't see a good reason why these lines are changed, I'd prefer to keep the title the same style for all apps. Same for the architecture support shields.

agners:
> Expanded the README to include detailed information about wake word detection, its role in the voice pipeline, common uses, installation instructions, and features of openWakeWord.

Also such content should be in the `DOCS.md`, which is shown under Documentation tab in Home Assistant.

PR #4519, opened 2026-03-27
fix(matter_server): skip --paa-root-cert-dir for JS beta server

### LINKED ISSUE #4518: Matter Server: --paa-root-cert-dir warning when using JS beta server

When using the Matter Server add-on with the JavaScript beta server (`beta: true`), the startup produces:

```
Warning: --paa-root-cert-dir is deprecated and no longer supported. This option will be ignored.
```

This is because the `run` script unconditionally passes `--paa-root-cert-dir` to both the Python and JavaScript server implementations, but the JS matter-server no longer supports this flag.

### Root cause

In `matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run`, the `matter_server_args` array includes `--paa-root-cert-dir` for both server paths:

```bash
matter_server_args+=(
  ...
  '--paa-root-cert-dir' "/data/credentials"
  ...
)
```

### Suggested fix

Move `--paa-root-cert-dir` into `extra_args` within the Python-only (non-beta) code path, matching the pattern used for other server-specific options.

### Environment

### Steps to reproduce

2. Start the add-on
3. Observe the deprecation warning in the logs

### Additional context

Related to #4516 (`--log-level-sdk` has the same issue).

markvp:
## Summary

- Moves `--paa-root-cert-dir` from the shared argument list into the Python-only (non-beta) code path via `extra_args`, matching the existing pattern for server-specific options
- The JavaScript Matter Server does not support `--paa-root-cert-dir` and emits `Warning: --paa-root-cert-dir is deprecated and no longer supported. This option will be ignored.` on every startup

## Test plan

- [ ] Start Matter Server add-on with `beta: true` — verify no `--paa-root-cert-dir` deprecation warning in logs
- [ ] Start Matter Server add-on with `beta: false` — verify `--paa-root-cert-dir` is still passed correctly to the Python server

Fixes #4518
Related: #4516, #4517

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Matter Server startup configuration to conditionally apply root certificate directory settings based on the execution environment, preventing potential compatibility issues and ensuring stable operation across different deployment modes.

### DIFF
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -85,6 +85,9 @@ else
   # Non-beta mode: Use Python Matter Server
   bashio::log.info "Using Python Matter Server"
 
+  # PAA root cert dir is only supported by the Python server
+  extra_args+=('--paa-root-cert-dir' "/data/credentials")
+
   if bashio::config.has_value "matter_server_version"; then
     matter_server_version=$(bashio::config 'matter_server_version')
     bashio::log.info "Installing Python Matter Server ${matter_server_version}"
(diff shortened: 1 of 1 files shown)


agners (changes requested):
As discussed in #4517, Please consolidate all Matter Server changes into #4517 (or a new PR).

PR #4521, opened 2026-03-27
fix(matter_server): conditionally pass --ota-provider-dir for JS beta server

### LINKED ISSUE #4520: Matter Server: --ota-provider-dir warning when using JS beta server without test-net DCL

When using the Matter Server add-on with the JavaScript beta server (`beta: true`) and `enable_test_net_dcl: false` (the default), the startup produces:

```
WARN   MatterServer.Ota     OTA provider directory (/config/updates) is configured but --enable-test-net-dcl is not set. Custom OTA files will be ignored.
```

This is because the `run` script unconditionally passes `--ota-provider-dir` to both server implementations, but the JS matter-server requires `--enable-test-net-dcl` to also be set for OTA provider files to be used ([ref](https://github.com/matter-js/matterjs-server/blob/a0f7904a01fb7bc34936c2203c21b78ea726fbfd/README.md?plain=1#L165)).

The Python server does not have this requirement — it uses `--ota-provider-dir` independently to load local OTA update JSON files.

### Root cause

In `matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run`, `--ota-provider-dir` is included in the shared `matter_server_args` array passed to both servers unconditionally.

### Suggested fix

- **JS beta path**: Only pass `--ota-provider-dir` when `enable_test_net_dcl` is also enabled
- **Python path**: Continue passing `--ota-provider-dir` unconditionally (it works independently)

### Environment

- Test-net DCL: disabled (default)

### Steps to reproduce

2. Leave `enable_test_net_dcl` at its default (`false`)
3. Start the add-on
4. Observe the OTA warning in the logs

### Additional context

Related to #4516 and #4518 (other JS-unsupported flags passed unconditionally).

markvp:
## Summary

- Moves `--ota-provider-dir` from the shared argument list into both server-specific code paths via `extra_args`
- **JS beta path**: Only passes `--ota-provider-dir` when `enable_test_net_dcl` is also enabled, since the JS server [requires both flags](https://github.com/matter-js/matterjs-server/blob/a0f7904a01fb7bc34936c2203c21b78ea726fbfd/README.md?plain=1#L165) for custom OTA files to be used
- **Python path**: Continues passing `--ota-provider-dir` unconditionally, as the Python server uses it independently to load local OTA update files

## Test plan

- [ ] Start with `beta: true`, `enable_test_net_dcl: false` — verify no OTA warning in logs, `--ota-provider-dir` is not passed
- [ ] Start with `beta: true`, `enable_test_net_dcl: true` — verify `--ota-provider-dir` is passed alongside `--enable-test-net-dcl`
- [ ] Start with `beta: false` — verify `--ota-provider-dir` is still passed unconditionally to the Python server

Fixes #4520
Related: #4516, #4517, #4518, #4519

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Over-The-Air (OTA) update provider configuration to respect server mode settings. The OTA provider directory is now conditionally applied based on whether the server runs in beta test mode or standard production mode, ensuring consistent update behavior and proper directory handling across all server configurations.

### DIFF
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -61,6 +61,11 @@ if [[ "${use_js_server}" == "true" ]]; then
   find ./node_modules -type d -name "cjs" -path "*/@project-chip/*" -exec rm -rf {} + 2>/dev/null || true
   rm -rf /tmp/* /var/tmp/* /root/.npm /root/.cache 2>/dev/null || true
 
+  # JS server only uses --ota-provider-dir when --enable-test-net-dcl is also set
+  if bashio::config.true "enable_test_net_dcl"; then
+    extra_args+=('--ota-provider-dir' "/config/updates")
+  fi
+
   if bashio::config.has_value "matter_server_env_vars"; then
     env_vars_count=$(bashio::config 'matter_server_env_vars | length')
     for (( i=0; i < env_vars_count; i++ )); do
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
CC @Apollon77

MartinHjelmare:
The fix looks correct. Ingo should confirm.

PR #4544, opened 2026-04-09
[openthread_border_router] - configure wakeup coordinator for 1.4

rwrozelle:


agners:
Thank you for your PR. So this feature is mandatory for 1.4?

Can you also bump the version in `config.yaml` and add an entry to `CHANGELOG.md`.

agners:
From what I am aware, the wakeup protocol is still being specified, so I don't think it make sense to enable this just yet.

rwrozelle:
Hi, it’s a draft PR, I’m testing using ESPHome Openthread configured chips, over there, the admins ignore draft PRs, but it helps people understand someone is looking into an area. They also actively use Discord.  Would it be better to close this and just expose my learnings in HA Discord?

agners:
Ah no problem. I somehow assumed it was set by accident to draft. You can leave the PR in draft or close once your experiments came to a conclusion.

That said, I am curious if this feature can be used, especially with ESPHome :sweat_smile:

PR #4545, opened 2026-04-10
letsencrypt: add httpreq provider support

LordMike:
Adds the `httpreq` lego provider using the certbot multi provider. Modeled after the #4467 PR which does roughly the same.

The setup has been tested locally and succesfully issued a certificate via a local dns01proxy host.

Fixes https://github.com/orgs/home-assistant/discussions/3458

## Release Notes

* **New Features**
  * Added HTTP request DNS provider as a certificate validation option.

* **Documentation**
  * Updated configuration guide with HTTP request DNS provider setup details and credential requirements.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.3.1
+
+- Add HTTP request DNS provider support
+
 ## 6.3.0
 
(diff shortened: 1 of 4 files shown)


sairon (approved):
Looks good, thank you!

LordMike:
@sairon do you by any chance know if it we should just flat out map every lego provider, or do them one-by-one as is now?

sairon:
@LordMike Since lego providers can be used directly through the generic option, I think it's not necessarily needed. I'm not against community additions like this one, where e.g. some custom validation is added as an extra, but I'd be probably against pre-emptive addition of _all_ the providers. Usually when someone adds a provider, it means they're actively using it and can provide feedback or fix if something breaks.

Issue #4431, opened 2026-02-17
Letsencrypt App v6.0 killed dns-desec.io recertification


kneip68:
### Describe the issue you are experiencing

Trying to renew a certificate with dns-desc.io in new version 6 off app. I can see the data for desec has been migrated correctly, with the exception of dns propagation wait time. In all other certbot instances I am running, the minimum setting is 120 secs to work, sometimes need to go to 180 seconds. Even when setting propagation-seconds to 180, the log says that it is waiting for 60 seconds, which I verified as correct by comparing the timestamps of the server answers in verbose mode  log.
I also verified that the _acme-challenge TXT record is present in the desec web ui. So that part of updating the dns server is working as expected.
Any hints on how to address the topic? in the logs I see an advice to extend --dns-multi-propagation-seconds, but I can not identify where to do so.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Let's Encrypt

### What is the version of the app?

6.0.0

### Steps to reproduce the issue

1. Migrate from 5.4.8 to 6.0.0 with dns-desec challenge running since long time
2. Start app when renewal is due
3. See logs: 
`2026/02/17 12:34:07 [DEBUG] POST https://desec.io/api/v1/domains/domain.tld/rrsets/
Waiting 60 seconds for DNS changes to propagate
Certbot failed to authenticate some domains (authenticator: dns-multi). The Certificate Authority reported these problems:
  Domain: homeassistant.heipmann.net
  Type:   dns
  Detail: DNS problem: NXDOMAIN looking up TXT for _acme-challenge.host.domain.tld - check that a DNS record exists for this domain
Hint: The Certificate Authority failed to verify the DNS TXT records created by Lego. Ensure the above domains are hosted by Desec and check https://go-acme.github.io/lego/dns/desec for further details on configuring this provider. Try increasing --dns-multi-propagation-seconds from its current value of 60 as well.
2026/02/17 12:35:09 [DEBUG] GET https://desec.io/api/v1/domains/domain.tld/rrsets/_acme-challenge.host/TXT/`
4. Set parameter propagation-seconds in App UI to 180
5. Restart the service
6. Same log entries as above

## System Information

version | core-2026.2.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.13.11
os_name | Linux
[...]

PR #4550, opened 2026-04-11
letsencrypt: don't override lego's DNS propagation timeout with propagation_seconds

### LINKED ISSUE #4549: Let's Encrypt dns-multi: propagation_seconds incorrectly overrides lego's internal DNS propagation timeout, causing renewal failures on slow providers

The run script writes $`{PROVIDER_ENV_PREFIX}_PROPAGATION_TIMEOUT=${PROPAGATION_SECONDS}` into the dns-multi credentials file, overriding lego's internal nameserver sync timeout with the user-facing `propagation_seconds` value (default: 60s).

These are two distinct concepts at different layers of the stack:

- `${PROVIDER}_PROPAGATION_TIMEOUT` — lego's internal polling timeout: how long lego waits for the DNS provider's nameservers to confirm a newly created TXT record has propagated before declaring success.
- `propagation_seconds` — a certbot-level sleep after lego has already confirmed propagation, before notifying the ACME server to perform verification.

By conflating them, the run script caps lego's sync timeout at `propagation_seconds` (60s by default), silently overriding lego's provider-specific defaults. For example, lego's built-in default for ClouDNS is 120s — reflecting that its nameservers are known to be slow — but the run script overwrites this with 60s.

**Failure mode:** lego polls nameservers and times out before sync completes; it never reaches certbot's post-sync sleep. The renewal fails at the lego layer, not at ACME verification.

There is no correct workaround via configuration: raising `propagation_seconds` to accommodate a slow provider also inflates certbot's post-sync sleep for every renewal, unnecessarily.

**Introduced in 6.3.0** when `certbot-dns-multi` replaced individual certbot DNS plugins. The previous per-provider plugins had no equivalent mechanism to override lego's internal timeouts.

Home Assistant OS

Let's Encrypt

6.3.1

1. Use dns-cloudns provider with a CNAME-delegated auth domain
2. Leave propagation_seconds unset (defaults to 60)
3. Attempt cert renewal
4. Lego nameserver sync stalls and fails before all nameservers confirm propagation

### System Health information

There are currently no repairs pending

```txt

```

```txt

```

### Additional information

Proposed Fix: Remove the `${PROVIDER_ENV_PREFIX}_PROPAGATION_TIMEOUT` line from the credentials file. The `--dns-multi-propagation-seconds` flag (already passed on the certbot command line) correctly handles the certbot-layer sleep; lego's internal timeouts should be left to their provider-specific defaults.

tedski:
Removes the `${PROVIDER_ENV_PREFIX}_PROPAGATION_TIMEOUT` line written into the dns-multi credentials file, and the now-dead `PROVIDER_ENV_PREFIX` variable and all its assignments.

`propagation_seconds` and lego's `*_PROPAGATION_TIMEOUT` are distinct settings at different layers:

- `*_PROPAGATION_TIMEOUT` — lego's internal polling timeout for waiting on the DNS provider's nameservers to confirm a newly created TXT record has propagated
 - `propagation_seconds` — a certbot-level sleep after lego confirms propagation, before notifying the ACME server to verify

The run script was writing `propagation_seconds` (default: 60s) as lego's internal timeout, silently overriding provider-specific defaults. For example, lego's built-in default for ClouDNS is 120s — the script was cutting that in half, causing renewal failures before lego even finished polling.

`propagation_seconds` is already correctly passed to certbot via `--dns-multi-propagation-seconds`. Lego's internal timeouts should be left to their provider-specific defaults.

This issue was introduced in 6.3.0 when `certbot-dns-multi` replaced individual certbot DNS plugins.

resolves #4549 

## Summary by CodeRabbit

* **Refactor**
  * Simplified DNS handling for Let's Encrypt: provider-specific timeout override removed so DNS propagation timeout now follows the underlying ACME tool's configured values and defaults.
* **Chores**
  * Bumped package version to 6.3.2 and added a changelog entry describing the DNS propagation behavior change.

### DIFF
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 6.3.2
 
-- No longer override lego's DNS propagation timeout with `propagation_seconds`.  Allows lego configuration and defaults.
+- No longer override lego's DNS propagation timeout with `propagation_seconds`. Allows lego configuration and defaults.
 
 ## 6.3.1
 
--- a/letsencrypt/CHANGELOG.md
+++ b/letsencrypt/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.3.2
+
+- No longer override lego's DNS propagation timeout with `propagation_seconds`.  Allows lego configuration and defaults.
+
 ## 6.3.1
 
 - Add HTTP request DNS provider support
--- a/letsencrypt/config.yaml
+++ b/letsencrypt/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 6.3.1
+version: 6.3.2
 breaking_versions: [5.3.0, 6.0.0]
 slug: letsencrypt
 name: Let's Encrypt
--- a/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
+++ b/letsencrypt/rootfs/etc/services.d/lets-encrypt/run
@@ -109,9 +109,6 @@ if [ "${CHALLENGE}" == "dns" ]; then
         # Use certbot-dns-multi for all other providers
         bashio::log.info "Using certbot-dns-multi for ${DNS_PROVIDER}"
 
-        # Some providers use a different env var prefix than the uppercased provider name
-        PROVIDER_ENV_PREFIX=""
-
         # Map config provider names to lego provider names
         declare -A LEGO_PROVIDERS=(
             ["dns-bunny"]="bunny"
@@ -233,8 +230,6 @@ if [ "${CHALLENGE}" == "dns" ]; then
         'dns-digitalocean')
             bashio::config.require 'dns.digitalocean_token'
             echo "DO_AUTH_TOKEN=$(bashio::config 'dns.digitalocean_token')" >> "${DNS_MULTI_CREDS}"
-            # Uppercased provider name differs from variable prefix
-            PROVIDER_ENV_PREFIX="DO"
             ;;
 
         # DirectAdmin
@@ -320,8 +315,6 @@ if [ "${CHALLENGE}" == "dns" ]; then
             bashio::config.require 'dns.google_creds'
             # File is copied during container init by file-structure.sh
             echo "GCE_SERVICE_ACCOUNT_FILE=/data/google.json" >> "${DNS_MULTI_CREDS}"
-            # Uppercased provider name differs from variable prefix
-            PROVIDER_ENV_PREFIX="GCE"
             ;;
 
         # Hetzner
@@ -502,8 +495,6 @@ if [ "${CHALLENGE}" == "dns" ]; then
                 AWS_REGION="$(bashio::config 'dns.aws_region')"
             fi
             echo "AWS_REGION=${AWS_REGION}" >> "${DNS_MULTI_CREDS}"
-            # Uppercased provider name differs from variable prefix
-            PROVIDER_ENV_PREFIX="AWS"
             ;;
 
         # SakuraCloud
@@ -552,12 +543,6 @@ if [ "${CHALLENGE}" == "dns" ]; then
             echo "WEBSUPPORT_SECRET=$(bashio::config 'dns.websupport_secret_key')" >> "${DNS_MULTI_CREDS}"
             ;;
         esac
-
-        # Set propagation timeout env var
-        if [ -z "${PROVIDER_ENV_PREFIX}" ]; then
-            PROVIDER_ENV_PREFIX=$(echo "${LEGO_PROVIDER}" | tr '[:lower:]-' '[:upper:]_')
-        fi
-        echo "${PROVIDER_ENV_PREFIX}_PROPAGATION_TIMEOUT=${PROPAGATION_SECONDS}" >> "${DNS_MULTI_CREDS}"
 
         ACME_ARGUMENTS+=("--authenticator" "dns-multi" "--dns-multi-credentials" "${DNS_MULTI_CREDS}" "--dns-multi-propagation-seconds" "${PROPAGATION_SECONDS}")
         if bashio::config.has_value 'dns.dns_multi_nameservers'; then


sairon (commented):
> This issue was introduced in 6.3.0 when `certbot-dns-multi` replaced individual certbot DNS plugins.

A nit, but it was `6.0.0` actually.

> `propagation_seconds` and lego's `*_PROPAGATION_TIMEOUT` are distinct settings at different layers:

This part is not entirely true. They should ideally have the same semantics, or better, the `*_PROPAGATION_TIMEOUT` should be taken into effect when renewing certs through `certbot-dns-multi` as well. Turns out as it's not the case (see #4431), despite `DESEC_PROPAGATION_TIMEOUT=180` set through the mapped option, the lego code path used by `certbot-dns-multi` didn't take it into effect and resulted only in in 60s wait, so there came the fix in https://github.com/home-assistant/addons/pull/4435/changes/3e15a2834cbe272acac08879af13ae2cc6a1b593. 

I don't have any logs available, only the AI summary from the linked issue, but if your fix works, then it means some lego providers also use the `*_PROPAGATION_TIMEOUT` option internally - so `--dns-multi-propagation-seconds` might not be needed there. But this PR will also remove the option to make these timeouts configurable. I wonder whether bumping the minimal propagation seconds like we [already do](https://github.com/home-assistant/addons/blob/master/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L388-L391) for some providers wouldn't be better approach here :thinking:

tedski:
> > This issue was introduced in 6.3.0 when `certbot-dns-multi` replaced individual certbot DNS plugins.
> 
> A nit, but it was `6.0.0` actually.
> 
> > `propagation_seconds` and lego's `*_PROPAGATION_TIMEOUT` are distinct settings at different layers:
> 
> This part is not entirely true. They should ideally have the same semantics, or better, the `*_PROPAGATION_TIMEOUT` should be taken into effect when renewing certs through `certbot-dns-multi` as well. Turns out as it's not the case (see #4431), despite `DESEC_PROPAGATION_TIMEOUT=180` set through the mapped option, the lego code path used by `certbot-dns-multi` didn't take it into effect and resulted only in in 60s wait, so there came the fix in [3e15a28](https://github.com/home-assistant/addons/commit/3e15a2834cbe272acac08879af13ae2cc6a1b593).
>
> I don't have any logs available, only the AI summary from the linked issue, but if your fix works, then it means some lego providers also use the `*_PROPAGATION_TIMEOUT` option internally - so `--dns-multi-propagation-seconds` might not be needed there. But this PR will also remove the option to make these timeouts configurable. I wonder whether bumping the minimal propagation seconds like we [already do](https://github.com/home-assistant/addons/blob/master/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L388-L391) for some providers wouldn't be better approach here 🤔

It is true and I believe #4431 ignored the `lego_env` config that is intended for this.  Allow me to illustrate what I mean.  I will use ClouDNS for this example since it is my provider, but the same holds true for any other provider.  ClouDNS is slow on the update of a new record being added to the zone, this is why lego specifies a longer timeout for this provider (see the [lego documentation](https://go-acme.github.io/lego/dns/cloudns/index.html)).  If you browse around to each of the providers, you'll see Lego upstream has unique values for many providers, likely discovered and added through empirical means.  These are in fact used by this addon and added to the `dns-multi-credentials` file [here](https://github.com/home-assistant/addons/blob/2fecd1f9f66a724b961e955469e83f9335c2de1f/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L179).  The dns-multi plugin's propagation setting is set separately [here](https://github.com/home-assistant/addons/blob/2fecd1f9f66a724b961e955469e83f9335c2de1f/letsencrypt/rootfs/etc/services.d/lets-encrypt/run#L562).

With that in mind, I can handcraft a credentials file similar to one the run script would formulate:

```
❯ cat cloudns.ini
dns_multi_provider = cloudns
CLOUDNS_SUB_AUTH_ID='12345'
CLOUDNS_AUTH_PASSWORD='redacted'
CLOUDNS_PROPAGATION_TIMEOUT=60
```

Then I can run certbot and we can observe the issue I originally encountered.  This will keep the provider-specific propagation timeout (time to wait for the backend to satisfy the api request's intent) and the dns-multi propagation seconds (time to wait for propagation to public nameservers) in lockstep at the default of 60 seconds.

```
❯ certbot certonly -a dns-multi \
  --config-dir . --work-dir . --logs-dir . \
  -d "*.home.tedski.net" \
  --dns-multi-credentials /tmp/certbottest/cloudns.ini \
  --dns-multi-nameservers '1.1.1.1,8.8.8.8' \
  --dns-multi-propagation-seconds 60 \
  --dry-run --verbose
Saving debug log to /private/tmp/certbottest/letsencrypt.log
Plugins selected: Authenticator dns-multi, Installer None
Simulating a certificate request for *.home.tedski.net
Performing the following challenges:
dns-01 challenge for Identifier(typ=IdentifierType(dns), value='home.tedski.net')
2026/04/15 08:01:14 [INFO] Found CNAME entry for "_acme-challenge.home.tedski.net.": "_acme-challenge.acme.tedski.net."
2026/04/15 08:01:16 [INFO] [home.tedski.net] Sync 0/8 complete
2026/04/15 08:01:26 [INFO] [home.tedski.net] Sync 1/8 complete
2026/04/15 08:01:37 [INFO] [home.tedski.net] Sync 1/8 complete
2026/04/15 08:01:48 [INFO] [home.tedski.net] Sync 2/8 complete
2026/04/15 08:01:58 [INFO] [home.tedski.net] Sync 3/8 complete
2026/04/15 08:02:09 [INFO] [home.tedski.net] Sync 3/8 complete
Cleaning up challenges
2026/04/15 08:02:09 [INFO] Found CNAME entry for "_acme-challenge.home.tedski.net.": "_acme-challenge.acme.tedski.net."
nameserver sync on home.tedski.net not complete
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /private/tmp/certbottest/letsencrypt.log or re-run Certbot with -v for more details.
```

Observe that the `Sync` step performed by `lego` only waited 60 seconds before bailing.  Now, let's show the two timeouts separated and how this causes the request to succeed.

We omit the provider-specific timeout from the credentials file in order to reflect the default case (lego uses its default for this provider of 180 seconds) and we keep the dns-multi plugin's propagation timeout at 60.  

```
❯ cat cloudns.ini
dns_multi_provider = cloudns
CLOUDNS_SUB_AUTH_ID='12345'
CLOUDNS_AUTH_PASSWORD='redacted'
```

Then we run the same `certbot` command as above.  Note that the dns-multi propagation seconds is still 60.

```
❯ certbot certonly -a dns-multi \
  --config-dir . --work-dir . --logs-dir . \
  -d "*.home.tedski.net" \
  --dns-multi-credentials /tmp/certbottest/cloudns.ini \
  --dns-multi-nameservers '1.1.1.1,8.8.8.8' \
  --dns-multi-propagation-seconds 60 \
  --dry-run --verbose
Saving debug log to /private/tmp/certbottest/letsencrypt.log
Plugins selected: Authenticator dns-multi, Installer None
Simulating a certificate request for *.home.tedski.net
Performing the following challenges:
dns-01 challenge for Identifier(typ=IdentifierType(dns), value='home.tedski.net')
2026/04/15 08:10:39 [INFO] Found CNAME entry for "_acme-challenge.home.tedski.net.": "_acme-challenge.acme.tedski.net."
2026/04/15 08:10:41 [INFO] [home.tedski.net] Sync 0/8 complete
2026/04/15 08:10:52 [INFO] [home.tedski.net] Sync 0/8 complete
2026/04/15 08:11:02 [INFO] [home.tedski.net] Sync 0/8 complete
2026/04/15 08:11:13 [INFO] [home.tedski.net] Sync 1/8 complete
2026/04/15 08:11:23 [INFO] [home.tedski.net] Sync 2/8 complete
2026/04/15 08:11:34 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:11:45 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:11:55 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:06 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:17 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:27 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:38 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:49 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:12:59 [INFO] [home.tedski.net] Sync 7/8 complete
2026/04/15 08:13:10 [INFO] [home.tedski.net] Sync 8/8 complete
Waiting 60 seconds for DNS changes to propagate
Waiting for verification...
Cleaning up challenges
2026/04/15 08:14:11 [INFO] Found CNAME entry for "_acme-challenge.home.tedski.net.": "_acme-challenge.acme.tedski.net."
Non-standard path(s), might not work with crontab installed by your operating system package manager
The dry run was successful.
```

Observe that the `Sync` step was permitted to go on for more than 60 seconds and more importantly succeed.  However, the propagation wait by the dns-multi plugin remains at 60 seconds.

This shows that the two are separately configurable and that they serve different purposes.  Putting them in lockstep actually makes the addon more confusing because the dns-multi documentation and lego documentation outline these specific tunables and without keeping them separate, it is confusing and surprising to the user.  In my case, I saw the default timeout on lego was 180 seconds and was left wondering why the plugin was bailing at 60.  Locking them in step causes an unnecessary increase in total time for cert generation that yields no value.  In the current version, overriding it overrides both and instead of waiting only 60 seconds, dns-multi waits 180 seconds on top of lego's 180.  That's unnecessary.

Coupling the two seems like it was the wrong solution for #4431 and hardcoding the increased timeouts in the run script ignores what lego already does and makes the addon susceptible to drift from upstream.  My change reverts that workaround in favor of using the existing `lego_env` config mapping for lego-specific tunables directly from their own documentation, while also allowing the use of the dns-multi plugin's specific tunable via the config.   I hope this clarifies things and would be more than happy to provide any more insight or examples to illustrate this.

letsencrypt/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 6.3.2
+
+- No longer override lego's DNS propagation timeout with `propagation_seconds`.  Allows lego configuration and defaults.
```
sairon:
```suggestion
- No longer override lego's DNS propagation timeout with `propagation_seconds`. Allows lego configuration and defaults.
```

sairon (approved):
Right, we still have `lego_env` which can be used to override any arbitrary options. Probably the need for changing the lego default would be fairly niche as well, so it's not a big deal. You're right (and I admitted it before) that coupling them doesn't work for all cases and my earlier assumption it does no harm was wrong, actually it can be the opposite. So let's do what you suggest. Thanks!

PR #4555, opened 2026-04-13
silabs-multiprotocol: disable build, remove unsupported architectures from config

sairon:
silabs-multiprotocol still contains armv7 in its build config. Remove it from there, while also adding an exclude condition to the workflow definition so it isn't re-built once the change is merged.

## Summary by CodeRabbit

* **Chores**
  * Removed `silabs-multiprotocol` add-on from automated build process
  * Removed armv7 architecture support from build configurations

* **Documentation**
  * Updated documentation to reflect removal of armv7 architecture support

### DIFF
--- a/.github/workflows/builder.yml
+++ b/.github/workflows/builder.yml
@@ -42,6 +42,8 @@ jobs:
         run: |
           declare -a changed_addons
           for addon in ${{ steps.addons.outputs.addons }}; do
+            # Skip deprecated add-ons that should no longer be built
+            [[ "$addon" == "silabs-multiprotocol" ]] && continue
             if [[ "${{ steps.changed_files.outputs.all }}" =~ $addon ]]; then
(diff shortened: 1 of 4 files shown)


sairon:
Alternatively, we can remove it from the repo :shrug:

agners:
> Alternatively, we can remove it from the repo 🤷

The problem is removing it from the repository will raise a repair suggesting uninstalling the app. The repair we raise doesn't really consider if ZHA/OTBR is using the app, it will just remove it, and hence break existing setups. According to analytics there are still 4k+ setups with this app.

We've considered adding a repair to Core at one point for that specific use case. I think it would make sense to have that first for a release or two before removing the app here.

agners (approved):
I like this approach, gives us a bit more time to migrate existing users away from multi-protocol 👍 .

PR #4561, opened 2026-04-14
mosquitto: bump components, remove unsupported archs

sairon:
- Remove unsupported architectures (armhf, armv7, i386)
- Update base image to Debian 13 (trixie)
- Update mosquitto to 2.1.2, see [Mosquitto changelog](https://mosquitto.org/ChangeLog.txt) for the list of changes
- Update libwebsockets to 4.5.8
- Update mosquitto-go-auth to 3.0.0
- Fix Home Assistant mqtt service (de)registration in init scripts

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Home Assistant MQTT service registration/deregistration

* **Chores**
  * Bumped add-on to 7.0.0
  * Removed support for armhf, armv7, and i386 (retaining aarch64 and amd64)
  * Base image updated to Debian 13 (trixie)
  * Updated core components: mosquitto → 2.1.2, libwebsockets → 4.5.8, mosquitto-go-auth → 3.0.0
  * Disabled embedded HTTP API and interactive editline support in the build

### DIFF
--- a/mosquitto/rootfs/etc/services.d/mosquitto/finish
+++ b/mosquitto/rootfs/etc/services.d/mosquitto/finish
@@ -5,11 +5,11 @@
 # s6-overlay docs: https://github.com/just-containers/s6-overlay
 # ==============================================================================
 
+bashio::services.delete "mqtt" || true
+
 if [[ "$1" -ne 0 ]] && [[ "$1" -ne 256 ]]; then
-  bashio::log.warning "Halt add-on"
-  bashio::services.delete "mqtt" || true
+  bashio::log.warning "Halt app"
   exec /run/s6/basedir/bin/halt
 fi
 
-bashio::services.delete "mqtt" || true
 bashio::log.info "Service restart after closing"
--- a/mosquitto/rootfs/etc/services.d/mosquitto/finish
+++ b/mosquitto/rootfs/etc/services.d/mosquitto/finish
@@ -1,4 +1,4 @@
-#!/usr/bin/env bashio
+#!/usr/bin/with-contenv bashio
 # vim: ft=bash
 # ==============================================================================
 # Take down the S6 supervision tree when service fails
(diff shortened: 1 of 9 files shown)


mosquitto/rootfs/etc/services.d/mosquitto/finish
```diff
@@ -7,7 +7,9 @@
 
 if [[ "$1" -ne 0 ]] && [[ "$1" -ne 256 ]]; then
   bashio::log.warning "Halt add-on"
+  bashio::services.delete "mqtt" || true
   exec /run/s6/basedir/bin/halt
 fi
 
+bashio::services.delete "mqtt" || true
```
agners: Should we just move this up to before the if?

agners (changes requested):
Nice find, I guess the registration being still there meant Core showed a discovered integration which potentially no longer was able to connect to the MQTT server?

mosquitto/rootfs/etc/services.d/mosquitto/finish
sairon: Shouldn't be needed, as the `if` block `exec`s into the `halt` binary.

sairon:
> Nice find, I guess the registration being still there meant Core showed a discovered integration which potentially no longer was able to connect to the MQTT server?

TBH I haven't checked what the Core behavior exactly is. I think this could be a problem if you started this app, then decided to use another app that registers itself as MQTT broker. But it's been like this for ages and seems no one ran into it, so it's rather an unlikely cause - though the self-conflict was real and I noticed it in the logs.

mosquitto/rootfs/etc/services.d/mosquitto/finish
agners: It's not needed, but it would avoid code duplication...

sairon: Oh, mea culpa. I first added it to the end, then realized it doesn't execute on stop. Since the "Halt add-on" message is purely informative, let's move it up (and also get rid of the "add-on"), right!

agners (approved):

Issue #4564, opened 2026-04-14
Govee2MQTT breaks with mosquito 7.0


cryptonym64:
### Describe the issue you are experiencing

7.0 is denying Govee2MQTT 

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Mosquitto broker

### What is the version of the app?

7.0.0

### Steps to reproduce the issue

Install 7.0
Govee2MQTT light strip goes unavailable
Deny messages found in app log

## System Information

version | core-2026.4.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.14.2
os_name | Linux
os_version | 6.12.77-haos
arch | x86_64
timezone | America/New_York
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 2932
Downloaded Repositories | 27

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | January 29, 2027 at 7:00 PM
relayer_connected | true
relayer_region | us-east-1
remote_enabled | false
remote_connected | false
alexa_enabled | false
google_enabled | true
cloud_ice_servers_enabled | true
remote_server | us-east-1-12.ui.nabu.casa
certificate_status | ready
instance_id | 57c1b4166bfb44ff9c61143dfbea4c80
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 17.2
-- | --
update_channel | stable
supervisor_version | supervisor-2026.04.0
agent_version | 1.8.1
docker_version | 29.3.1
disk_total | 97.7 GB
disk_used | 35.4 GB
nameservers | 192.168.90.1
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | kvm
board | ova
supervisor_api | ok
version_api | ok
[...]

PR #4565, opened 2026-04-14
mosquitto: disable auth_plugin_deny_special_chars to allow '/' in client ids (#4564)

manuveli:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed regression where clients whose IDs contain forward slashes (e.g., `govee2mqtt`) were incorrectly rejected with "ACL denying access to client with dangerous client id" error following the Mosquitto 2.1 upgrade.
  * Restored support for special characters in client IDs and usernames with authentication plugin.

### DIFF
--- a/mosquitto/CHANGELOG.md
+++ b/mosquitto/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.0.1
+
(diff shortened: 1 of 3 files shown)


sairon (approved):
Nice catch, thanks for the fix!

Issue #4531, opened 2026-04-03
Mosquitto: custom config loaded before main config, log_type settings get overridden


balsih:
### Describe the issue you are experiencing

# Description
When using the `customize` feature to load a custom `/share/mosquitto/mosquitto.conf`, 
the file is loaded **before** `/etc/mosquitto/mosquitto.conf`. This means any settings 
in the custom config (e.g. `log_type`) are overridden by the main config generated by 
the add-on, making it impossible to control log verbosity via the customize folder.

Check the **additional information** section at the end for the impact.

## Expected behavior
Custom config settings should take precedence over (or at least not be overridden by) 
the generated main config.

## Actual behavior
From the logs it is clear the custom config is loaded first:
1. Loading config file /share/mosquitto/mosquitto.conf
2. Config loaded from /etc/mosquitto/mosquitto.conf.

The main config is loaded second and overrides `log_type`.

## Suggestion
Either load the custom config **after** the main config, or use `include_dir` at the 
**end** of `/etc/mosquitto/mosquitto.conf` so custom settings win.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Mosquitto broker

### What is the version of the app?

2.0.22

### Steps to reproduce the issue

1. Enable `customize: active: true, folder: mosquitto` in the add-on config
2. Create `/share/mosquitto/mosquitto.conf` with `log_type warning`
3. Restart the add-on
4. Observe logs — informational connect/disconnect messages still appear

## System Information

version | core-2026.4.0
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.14.2
os_name | Linux
os_version | 6.12.47-haos-raspi
arch | aarch64
timezone | Europe/Zurich
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 2929
Downloaded Repositories | 3

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

[...]

PR #143, opened 2017-07-21
Mqtt customize

pvizeli:

PR #4574, opened 2026-04-18
mosquitto: ensure custom config takes precedence over defaults

Boilerplate4u:
## Problem

When users configure a custom config file (`/share/mosquitto/mosquitto.conf`), settings are silently ignored. For example, setting `log_dest topic` and `log_type debug` has no effect and all output still goes to the add-on log.

Upstream issue: https://github.com/home-assistant/addons/issues/4531

## Root cause

The cause is that `log_dest` and `log_type` directives in Mosquitto's config parser are accumulated additively using bitwise OR. Once `log_dest stdout` is set in the default config, no subsequent directive can remove it. The user's custom settings are simply added on top, never replacing the defaults.

The problem cannot be solved by reordering directives in the config file, because once a `log_dest` bit is set it cannot be unset by any subsequent directive. The fix must be applied in `mosquitto.gtpl`: when `customize` is active, the default `log_dest` and `log_type` directives are never used.

## Fix

When `customize` is active, skip the entire `log_dest stdout` and `log_type` block in `mosquitto.gtpl` and load `include_dir` instead. The user's custom config then becomes the sole source of logging configuration. When `customize` is not active, behaviour is identical to today.

## Effect on users

Users can now control logging fully via `/share/mosquitto/mosquitto.conf`. Two verified examples:

Log everything to file, including `MOSQ_LOG_DEBUG` and `MOSQ_LOG_INTERNAL`:

```
#
# Log to file
#
log_dest file /share/mosquitto/mosquitto.log
log_type all
log_timestamp true
log_timestamp_format [%H:%M:%S]
```

Log to topic. Note: `MOSQ_LOG_DEBUG` and `MOSQ_LOG_INTERNAL` are excluded by Mosquitto at the source level ([src/logging.c line 388](https://github.com/eclipse-mosquitto/mosquitto/blob/master/src/logging.c#L388)):

```
#
# Log to topic
# Note: excludes MOSQ_LOG_DEBUG & MOSQ_LOG_INTERNAL.
#
log_dest topic
log_type all
log_timestamp true
log_timestamp_format [%H:%M:%S]
```

## Risk

When `customize` is not active, behaviour is completely unchanged. When `customize` is active, `log_dest stdout` is no longer set by default, meaning the add-on log will be silent unless the user's custom config explicitly sets a `log_dest`. This is intentional and expected — the user has opted in to full control by enabling `customize`. Users who enable `customize` without configuring any `log_dest` will see no log output, which is consistent with how Mosquitto behaves when no logging is configured.

## Changed file

`mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl`

## Deep analysis

Examination of [`src/conf.c`](https://github.com/eclipse-mosquitto/mosquitto/blob/master/src/conf.c) shows that each `log_dest` line is OR-combined into a bitmask using `|=`:

```c
cr->log_dest |= MQTT3_LOG_STDOUT;
cr->log_dest |= MQTT3_LOG_TOPIC;
cr->log_dest |= MQTT3_LOG_FILE;
```

Once a destination bit is set, it cannot be unset by a subsequent `log_dest` line. The only escape is `log_dest none` which resets the entire bitmask. This means that even if `include_dir` were moved to the very end of the generated config, the `log_dest stdout` set in the first line of the template would remain permanently active in the final bitmask — making a position change alone insufficient to fix the problem.

The same additive behaviour applies to `log_type`.

At the end of `config__read()` the accumulated value is applied once:

```c
if(cr.log_dest_set){
    config->log_dest = cr.log_dest;
}
```

There is no mechanism for a later config line to remove an earlier destination.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized configuration template structure for improved clarity. Customization folder includes and debug logging directives are now conditionally placed based on configuration settings.

### DIFF
--- a/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
+++ b/mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl
@@ -1,4 +1,7 @@
 user root
+{{ if .customize }}
+include_dir /share/{{ .customize_folder }}
+{{ else }}
 log_dest stdout
 {{ if .debug }}
 log_type all
@@ -7,6 +10,7 @@ log_type error
 log_type warning
 log_type notice
 log_type information
+{{ end }}
 {{ end }}
 log_timestamp_format %Y-%m-%d %H:%M:%S
 persistence true
@@ -43,10 +47,6 @@ auth_opt_http_port 80
 auth_opt_http_getuser_uri /authentication
 auth_opt_http_superuser_uri /superuser
 auth_opt_http_aclcheck_uri /acl
-
-{{ if .customize }}
-include_dir /share/{{ .customize_folder }}
-{{ end }}
 
 listener 1883
 protocol mqtt


Boilerplate4u:
Mr CodeRabbit - Looks like your training data missed this one, or maybe your LLM trainer didn't teach you how to read! 😄

The AI summary misses the point entirely and is surprisingly poor for a code AI bot. This is not a refactor for 'improved clarity', it is a bug fix. The root cause is that Mosquitto's config parser accumulates `log_dest` values using bitwise OR (`|=`), meaning once `log_dest stdout` is set it cannot be removed by any subsequent directive. Moving or reordering `include_dir` would not have helped. The only correct fix is to conditionally skip the `log_dest stdout` and `log_type` block entirely when `customize` is active. See the Deep analysis section in the PR description for full details.

sairon (changes requested):
This only accounts for customization of `log_*` options (which will then even stay unset/default unless user sets them!). The remaining options which are not interpreted in the bitwise OR manner but with "last value wins" semantics will be always overridden by the upstream config.

We can't really tell what is the most common use case of the customize option, but this change has high chance of breaking some of them instead of only fixing a bug in niche use case which has been around for ~9 years (since #143).

Boilerplate4u:
@[sairon](https://github.com/sairon) Well, it's pretty clear that the current custom config is completely broken.

This fix is simple: if a custom config exists, it's on the user to include the right parameters. If there's no custom config, everything just works like before. It's just a few lines in the beginning of `mosquitto/rootfs/usr/share/tempio/mosquitto.gtpl`:

```
user root
{{ if .customize }}
include_dir /share/{{ .customize_folder }}
{{ else }}
log_dest stdout
{{ if .debug }}
log_type all
{{ else }}
log_type error
log_type warning
log_type notice
log_type information
{{ end }}
{{ end }}
```

What else would you suggest?

sairon:
> This fix is simple: if a custom config exists, it's on the user to include the right parameters.

It's not a simple fix if it meas it's a potential breaking change as well.

> What else would you suggest?

I made the options customizable in #4589.

PR #4576, opened 2026-04-20
mariadb: Redirect LOG FD when suppressing log messages

lmagyar:
Just a quick refactoring. I see you are planning a new release, this can be added.

cc @sairon


sairon (approved):
Nice, thanks! Note that this was adopted from Mosquitto app which still uses the old approach, feel free to open a PR for that as well (or I can batch it with some future changes).

PR #4580, opened 2026-04-20
mosquitto: Redirect LOG FD when suppressing log messages

lmagyar:
Equivalent of #4576 for mosquitto.

Note: not tested.

Update: Version is intentionally not increased, this has no functional effect on anything.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal service management code in the Mosquitto add-on.


sairon:
LGTM but merging the change will trigger a rebuild and re-publish of the previously tagged image without a `config.yaml` change. I will include your change in the next release to make sure it doesn't happen.

sairon (approved):
Bundling with #4589. Thanks!

PR #4591, opened 2026-04-25
Enable kernel oplocks in samba config

jfurtner:
Recent change in Samba means that files changed locally aren't picked up by Samba immediately. `kernel oplocks = yes` fixes this behaviour at the cost of disabling performance related enhancements (around folder content caching) with SMB3, but a better user experience if making changes locally and through the Samba addon.

## Summary by CodeRabbit

* **Configuration Updates**
  * Updated Samba configuration to explicitly enable kernel oplocks globally for improved file locking/caching behavior.
  * Disabled kernel oplocks for the **backup** and **media** shares to preserve their intended behavior.
  * Added a **12.7.1** changelog note covering the global kernel oplocks change (excluding **backup** and **media**).
  * Updated the add-on manifest version to **12.7.1**.


agners (changes requested):
Thank you for your contribution!

This seems very sensible for any config directory. However, I wonder if it is a good idea to enable it for media/backup. Those are not usually edited, and kernel locks adds more network round trips when accessing through smb from what I understand.

Can you scope the change so that those two are excluded.

To deploy this as  a new release please bump the version in `config.yaml` and add an entry to `CHANGELOG.md` in this same PR.

jfurtner:
Made the requested change, now turned off for media and backups.

jfurtner:
Note that this uses the image homeassistant/{arch}-addon-samba, which will need a corresponding version bump, I've looked briefly but don't see how that is made available to supervisor to install.

agners (approved):
LGTM, thank you!

PR #4597, opened 2026-04-29
Mark ingress proxy as trusted in Z-Wave JS app

AlCalzone:
Fixes the issues reported in https://github.com/zwave-js/zwave-js-ui/issues/4535

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a validation error appearing in Z-Wave JS UI logs caused by improper proxy configuration handling.
  * Enhanced application compatibility with proxy layers and ingress endpoints, ensuring proper request processing and improved system stability.

### DIFF
--- a/zwave_js/CHANGELOG.md
+++ b/zwave_js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.3.1
+
+Avoids a validation error in the Z-Wave JS UI logs complaining about incorrect proxy configuration.
+
 ## 1.3.0
 
 ### Z-Wave JS 15.23.2...15.23.5
--- a/zwave_js/config.yaml
+++ b/zwave_js/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 1.3.0
+version: 1.3.1
 slug: zwave_js
 name: Z-Wave JS
 description: Control a Z-Wave network with Home Assistant Z-Wave JS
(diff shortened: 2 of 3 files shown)


AlCalzone:
@agners I figure it would be slightly more correct to pass the actual ingress proxy IP - any clue how I can determine that programmatically?

agners:
> @agners I figure it would be slightly more correct to pass the actual ingress proxy IP - any clue how I can determine that programmatically?

Sorry for the late reply.

Best practice is to *not* use the same port to expose the web interface optionally outside (`ports` key) and ingress. Currently that is the case, the "Z-Wave JS UI web port" 8091 is also used for ingress. Detecting ingress based on source IP works in theory (source IP would be whatever `supervisor` resolves to), but it is not ideal as source IP could potentially be spoofed and makes your code messier than it needs to be.

Instead, set `ingress_port` to 0, which means Supervisor will assign a ingress port on startup. You can then read the dynamically assign port using `$(bashio::addon.ingress_port)`. Then create a separate server. For this server you can assume any connection to be pre-authenticated by Supervisor/Home Assistant Core. You can also read the headers `X-Remote-User-Id`, `X-Remote-User-Name`, and `X-Remote-User-Display-Name` (see also [Authenticating a user when using ingress](https://developers.home-assistant.io/docs/apps/security?_highlight=x&_highlight=remote#authenticating-a-user-when-using-ingress)) to learn what user is accessing Z-Wave JS via Ingress. Today only admin users have access to Ingress. I probably need to expand that section a bit. 


Note that with ESPHome and Music Assistant thins are a bit more complicated since they operate with `host_network: true`. In that mode, no "truly" internal IP is available. The hassio gateway IP used to be open to the public and only recently got protected. But this did not apply for regular apps like Z-Wave JS.

agners:
Btw, there is also [the idea](https://github.com/home-assistant/supervisor/issues/6828) to support ingress via Unix domain socket which is by nature only available local and would secure access quite nicely.

AlCalzone:
I've started implementing the 2nd server approach in Z-Wave JS UI. I'll go with that.
Will mark this as draft for now, so I don't lose track of the explanation above.

PR #4601, opened 2026-05-01
Clarify using the latest voices with Wyoming Piper

AlexGustafsson:
The Wyoming Piper code fetches from the main Huggingface "branch", not
v1.0.0. Though in order to use newly available voices, one might need to
update the voices list - clarify this in the docs.

SEE: https://github.com/rhasspy/wyoming-piper/issues/49

## Summary by CodeRabbit

* **Documentation**
  * Updated voice model setup guidance to reference the latest available voices from the rolling main branch.
  * Added note about using the `update_voices` option to fetch the most recent voice models.

### DIFF
--- a/piper/DOCS.md
+++ b/piper/DOCS.md
@@ -27,7 +27,8 @@ for more information.
 [Listen to voice samples](https://rhasspy.github.io/piper-samples/)
 
 Name of the Piper voice to use, such as `en_US-lessac-medium` (the default).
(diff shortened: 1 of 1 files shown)


agners:
#4767 updated the URL as well. I don't think that the option is that important to document separately, so I am closing this PR.

AlexGustafsson:
Thanks for taking a look!

PR #4603, opened 2026-05-04
rfc2136_algorigthm must now be lower case

mloftis:
Woke up to my HAOS Home Assistant unreachable due to an expired Let's Encrypt cert.  Because somewhere the algorithm was changed to only accept lower case.  This fixes the docs at least so that others setting up new don't fall into the same pit.


mloftis:
I'll try again later with a local clone.  Something went very wrong with Codespaces or the GH fork, I changed one line to lowercase the algorithm in the docs and I've got a ~700 line diff.  Very much not the intention of this PR to touch a big chunk of `DOCS.md`

PR #4619, opened 2026-05-20
[openthread_border_router] Promote Thread 1.4 beta to stable

### LINKED ISSUE #50: Bump OTBR to full Thread 1.4 support
### Roadmap opportunity

This epic tracks the delivery of the following roadmap opportunity: 

- https://github.com/OpenHomeFoundation/roadmap/issues/91

### Description

Aim is to bump OTBR to fully support Thread 1.4

### Deliverables/features

- OTBR release with Thread 1.4 support

### Completion criteria

_No response_

### Additional context

_No response_

puddly:
This PR promotes our current beta to stable and cuts a new beta from today's OTBR-POSIX commit.

The beta cycle was unusually long for this addon so we've decided to un-toggle the beta toggle for everyone upgrading. If you want to keep running the beta, please turn it back on.

For the new beta, I picked the most recent OTBR commit as nothing really stood out as a clean point to cut it. It runs without issues locally.

I've also taken this opportunity to replace the OTBR build scripts with just manual dependency management and CMake invocation. `elf_diff` confirms the resulting binaries are code-wise identical. The only change is a version string changing from `thread-reference-20250612-482-g0c4f22d69` turning into `0c4f22d`, which I think is arguably more readable and less confusing for end users.

---

https://github.com/home-assistant/epics/issues/50

## Summary by CodeRabbit

* **New Features**
  * Built-in mDNS enabled by default; OTBR mDNS local hostname standardized to "<hostname>-otbr".

* **Bug Fixes**
  * Improved REST socket handling for more reliable API behavior.
  * Cleared stale radio source-match entries during restore to avoid table exhaustion.
  * Channel monitoring no longer starts automatically by default.

* **Chores**
  * Version bumped to 3.0.0; serialx updated to 1.8.0.
  * Build/runtime packaging and Docker build flow reworked.
  * Auto-disable beta mode when promoting to stable; mDNS startup/management adjusted.


openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
```diff
@@ -4,22 +4,42 @@
...
+if [[ -f ${marker} ]]; then
+    marker_version=$(< "${marker}")
+fi
+
+beta_active=false
 if bashio::config.true 'beta'; then
-    bashio::log.info "Beta mode enabled, using OpenThread built-in mDNS."
+    if (( marker_version < beta_auto_disable_version )); then
+        bashio::log.warning "Disabling beta mode automatically; the previous beta is now stable."
+        bashio::addon.option 'beta' '^false'
```
agners, line 24: `'^false'`? I guess a typo?

```suggestion
        bashio::addon.option 'beta' 'false'
```

openthread_border_router/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 2.16.8
+version: 3.0.0
```
agners, line 2: Let's also mark this as a breaking change so for folks which have auto-update enabled it doesn't just auto-update.

```suggestion
version: 3.0.0
breaking_versions: [3.0.0]
```

openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
```diff
@@ -4,22 +4,42 @@
...
+declare marker
+declare marker_version
+declare beta_active
+
+beta_auto_disable_version=1
+marker=/data/beta_auto_disabled
+marker_version=0
+if [[ -f ${marker} ]]; then
+    marker_version=$(< "${marker}")
+fi
```
agners, line 18: Yeah I realized that we need this trickery, since we don't really know when the first start after the update is 😢 .

On that note, this came just up on Discord: When running non-beta (and the Core was connected to the REST API), then restarting into the beta version (or, in this PR case, on update from 2.x to 3.0.0), the REST server fails to start with:

```
[WARN]-REST----: REST server failed to start on 172.30.32.1:8081
```

The reason for this is the old custom HTTP library used `SO_REUSEADDR` socket option, and the new cpp-httplib library uses `SO_REUSEPORT`. The two are inherently incompatible, so everytime you switch from one to the other you get the message above (new OTBR) or `[ERR ]-REST----: InitializeListenFd error bind : Address already in use` (old OTBR).

Now we are mainly concerned about the upgrade case. One way to fix this would be to just wait TCP/IP `TIME_WAIT`, which is 60s on Linux. Supposedly there is a way to do force kill `TIME_WAIT` via `ss -K`, but it needs kernel config `CONFIG_INET_DIAG_DESTROY` which it seems we don't have 😢 🤷 . So I guess this means sleep for 60 seconds and write a message so users know this is one time only.

agners (changes requested):

openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
puddly: That's not good. I had a PR for this up a long while ago (https://github.com/openthread/ot-br-posix/pull/3212) but I think I forgot the direction this was fixing. I thought it was beta -> stable. If it's stable -> beta, that's not great.

openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
puddly: It's to keep it as a literal: https://github.com/hassio-addons/bashio/blob/a3d7db5803d3ef2179ddcb3698635819591fed1d/lib/addons.sh#L614

openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
puddly: I've re-submitted the PR upstream and applied the patch locally here.

agners: Is `SO_REUSEPORT` still applied then? Is this not interfering the same way (still two sockets with different socket options)? :thinking:

agners: Ah I see we are overwriting the existing socket options. Yeah I agree `SO_REUSEADDR` is right, and since we used this before, this will avoid the `TIME_WAIT` problem entirely. Nice! :+1:

schdro:
>For the new beta, I picked the most recent OTBR commit as nothing really stood out as a clean point to cut it. It runs without issues locally.

Well, only to mention, IMHO there was a recent CVE-2026-8369 in Openthread NAT64.
Unclear if OTBR addon is affected (e.g. Espressif ESP IDF master deliberately unmarked because they replace with own NAT64).
But from that view new OT beta (containing upstream fix) IMHO *is* interesting.

puddly:
Interesting! We don't enable NAT64 by default so I think the OTBR app wouldn't be affected in its default configuration.

CC @agners

puddly:
I've created a branch (building on this one) that backports the CVE fix to `stable`: https://github.com/puddly/hass-addons/compare/puddly/otbr-promote-stable...puddly:hass-addons:puddly/otbr-cve-2026-8369

openthread_border_router/Dockerfile
```diff
@@ -16,28 +16,47 @@ ENV DOCKER=1
...
+        build-essential \
+        ninja-build \
+        cmake \
+        iptables \
+        libjsoncpp-dev \
+        libnetfilter-queue-dev \
+        nodejs \
+        npm \
+        libprotobuf-dev \
+        protobuf-compiler \
```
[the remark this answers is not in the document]
agners: @Copilot can you recreate the suggestion?

openthread_border_router/rootfs/etc/s6-overlay/scripts/enable-check.sh
```diff
@@ -4,22 +4,42 @@
 # Select OTBR version and enable mDNSResponder for stable mode
```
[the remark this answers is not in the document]
agners: @Copilot can you recreate the suggestion?

openthread_border_router/Dockerfile
```diff
@@ -16,28 +16,47 @@ ENV DOCKER=1
...
     && cd /usr/src/ot-br-posix \
     && git fetch origin ${OTBR_BETA_VERSION} \
     && git checkout ${OTBR_BETA_VERSION} \
-    && git submodule update --init \
-    # Bootstrap downloads a few more repos
-    && ./script/bootstrap \
-    # Remove libsystemd-dev to avoid unnecessary linkage
-    && apt-get purge -y libsystemd-dev
+    && git submodule update --init --recursive --depth 1 \
+    && patch -p1 < /usr/src/0001-rest-SO_REUSEADDR.patch
```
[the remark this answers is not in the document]
agners: It seems to build, I guess it comes in as a dependency already. So why bother.

agners (approved):
LGTM, tested on my setup again, beta option gets disabled automatically on first start. Build looks good and works. Let's ship it!

Issue #4299, opened 2026-01-20
OTBR no serial device for network matter device


fueltank34:
### Describe the issue you are experiencing

When I try to setup my SLZB06-MR3 over PoE
1. No serial device is selectable in the UI (as expected)
2. Switching to yaml and manually specifying /dev/null or anything else does not bypass the serial check

Request:
If user enters in /dev/null in device AND supplies a valid network device address, please let it save the configuration.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which add-on are you reporting an issue with?

OpenThread Border Router

### What is the version of the add-on?

2.15.3

### Steps to reproduce the issue

1. RPI4 with NO USB devices attached, running SSD
2. Switch to OTBR config tab
3. Switch to yaml
4. Set device: to /dev/null or /dev/tty01
5. Enter in network_device as tcp://192.168.X.XXX:6639
6. Click Save configuration

Failed to save add-on configuration, Device '/dev/tty01' does not exist in OpenThread Border Router (core_openthread_border_router). Got {'device': '/dev/tty01', 'baudrate': '460800', 'flow_control': True, 'otbr_log_level': 'notice', 'firewall': True, 'nat64': False, 'network_device': 'tcp://192.168.1.118:7638'}
---
Failed to save add-on configuration, Device '/dev/null' don't match the filter subsystem=tty! in OpenThread Border Router (core_openthread_border_router). Got {'device': '/dev/null', 'baudrate': '460800', 'flow_control': True, 'otbr_log_level': 'notice', 'firewall': True, 'nat64': False, 'network_device': 'tcp://192.168.1.118:7638'}

...

## System Information

version | core-2026.1.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.13.11
os_name | Linux
os_version | 6.12.47-haos-raspi
arch | aarch64
timezone | Australia/Sydney
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 2659
Downloaded Repositories | 3

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
[...]

PR #4620, opened 2026-05-20
OpenThread Border Router: Allow network_device without USB device

jmarcelomb:
I'm trying to use an SLZB-MR4 (network-connected RCP) on my Raspberry Pi 4B. The add-on requires `device` to be set even though I'm using `network_device` — there's no USB serial device to provide, and setting a dummy value feels wrong.

This makes `device` optional in the schema and adds a check that fails clearly if neither `device` nor `network_device` is configured.

## Summary by CodeRabbit

* **New Features**
  * Device configuration is now optional, allowing setups without an explicit device entry.

* **Bug Fixes**
  * Startup now validates configuration and exits early with a clear error when neither device nor network device is provided.
  * When a network device is selected, the runtime automatically uses an appropriate virtual device for operation.

### DIFF
--- a/openthread_border_router/config.yaml
+++ b/openthread_border_router/config.yaml
@@ -39,7 +39,7 @@ ports_description:
   8080/tcp: OpenThread Web port
   8081/tcp: OpenThread REST API port
 schema:
-  device: device(subsystem=tty)
+  device: device(subsystem=tty)?
   baudrate: list(57600|115200|230400|460800|921600|1000000)
   flow_control: bool
   backbone_interface: str?
--- a/openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
+++ b/openthread_border_router/rootfs/etc/s6-overlay/s6-rc.d/otbr-agent/run
@@ -42,10 +42,14 @@ if [ "$(cat /proc/sys/net/ipv6/conf/all/forwarding)" != "1" ]; then
     bashio::log.warning "IPv6 routing/forwarding is not enabled! Make sure the Home Assistant host has IPv6 forwarding enabled."
 fi
 
-device=$(bashio::config 'device')
+if ! bashio::config.has_value 'device' && ! bashio::config.has_value 'network_device'; then
+    bashio::exit.nok "No device configured. Set either 'device' (serial port) or 'network_device' (TCP host:port)."
+fi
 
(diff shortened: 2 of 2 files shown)


SveLil:
@agners Can you take a look and approve? There are quite a lot of people waiting for this (see #4299)

agners:
> @agners Can you take a look and approve? There are quite a lot of people waiting for this (see #4299)

I don't really like it. It means that users who use a USB device get a strange error if they forget to select it. And these users should be our primary concern.

Tunneling Thread radio messages through network is generally a bad idea, it is a timing sensitive protocol and the higher level protocol itself is a network protocol, so why not just install a BR firmware on the device if its network capable!? :shrug: 

What I actually would prefer is have a separate app, which we'd then also could clearly mark as experimental. It would also have the advantage that you can run a proper locally connected radio still.

el-gee:
> why not just install a BR firmware on the device if its network capable!?

devices are starting to offer this, but it's often beta support at best.  Users may have an acceptable network and be willing to accept the risks of tunneling thread radio message between the radio and the OTBR server (a common use case being HA physically running in a suboptimal location for radio placement -- aka a corner in the basement) .  

It feels like there are 2 things worth keeping distinct in discussion:  The oddity of the UI behavior here and the general complications and nuances of the configuration it currently blocks: network only radio devices.

fueltank34:
Blocking useful network functionality because a USB user might misconfigure their settings feels like the wrong trade-off. Instead of restricting choices for users who intentionally deploy network-attached coordinators, shouldn't we just improve the UI validation or error messaging to guide the USB users better?

Network-attached RCPs (like the SLZB-06 or TubeZB coordinators) are incredibly common for centralising radios away from server racks.

Giving users the choice of how they connect their hardware, while keeping the configuration unified in one place, is a massive UX win.

SveLil:
I agree that using a network_device instead of a USB device isn't ideal, but as OTBR suppots this, this app should support the configuration.

zeratul021:
@jmarcelomb @agners any chance we could support this in some mode, possibly with some checkbox/tunable?
It is bit frustrating having to select fake USB devices for something that is already supported to work.

jmarcelomb:
I've created [this](https://github.com/jmarcelomb/addons) fork that I'm using to have no USB selected:
https://github.com/jmarcelomb/addons

You can add this repo to HA and use this 😊

PR #4623, opened 2026-05-23
feat(matter_server): allow configuring custom primary network interface

txptr:
## Description
This PR introduces an optional configuration option `matter_server_primary_interface` to the Matter Server add-on. This allows users to manually specify the primary network interface used by the Matter Server (e.g., `enp0s19`), rather than relying solely on the Supervisor's automatic detection.

## Motivation & Context
In complex network environments (such as setups with multiple network interfaces, untagged/tagged VLANs, or multiple gateways), the automatic network interface detection by the Home Assistant Supervisor can sometimes select the wrong interface for the Matter Server. 

Currently, trying to append `--primary-interface` via `matter_server_args` results in duplicate arguments where the automatically detected one takes precedence, making it impossible for users to override it. This PR solves the issue by offering a dedicated, clean UI configuration field.

## How Has This Been Tested?
- **Empty configuration (Default):** Verified that the add-on falls back seamlessly to the Supervisor's automatic detection. No breaking changes for existing users.
- **Valid interface provided:** Verified that the custom interface is correctly injected and used by the Matter Server.
- **Invalid/Typo interface provided:** Verified that the built-in safety check (`ip link show`) detects the missing interface, logs a proper warning/error, and safely falls back to the system default instead of crashing the add-on loop.

## Checklist
- [x] Follows architecture guidelines for Home Assistant add-ons.
- [x] Includes English translations for the new configuration option.
- [x] Implementation is fully backwards-compatible.
- [x] Input validation included to prevent boot-loops caused by user typos.

## Summary by CodeRabbit

* **New Features**
  * Optional setting to manually specify the Matter Server’s primary network interface or leave empty to use the system default.
  * Added a user-facing name and description for this setting.

* **Bug Fixes / Behavior**
  * Better validation and clearer error logging for a user-specified interface.
  * Improved automatic detection fallback; startup now aborts with an explicit error if no primary interface is found.

### DIFF
--- a/matter_server/config.yaml
+++ b/matter_server/config.yaml
@@ -35,6 +35,7 @@ schema:
   enable_test_net_dcl: bool?
   ble_proxy: bool?
   bluetooth_adapter_id: int?
+  matter_server_primary_interface: str?
   matter_server_args:
     - str?
   matter_server_env_vars:
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -122,18 +122,35 @@ if bashio::config.true "enable_test_net_dcl"; then
     extra_args+=('--enable-test-net-dcl')
 fi
 
-primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
-
-# Try fallback method (e.g. in case NetworkManager is not available)
-# shellcheck disable=SC2086
-if [ -z ${primary_interface} ]; then
-  bashio::log.warning 'Trying fallback method to determine primary interface'
-  primary_interface="$(ip --json route show default | jq --raw-output '.[0].dev')"
+# Check if the user configured a custom primary interface
+if bashio::config.has_value 'matter_server_primary_interface'; then
+    custom_interface=$(bashio::config 'matter_server_primary_interface')
+    
+    # Check if the configured interface actually exists in the system
+    if ip link show "$custom_interface" > /dev/null 2>&1; then
+        bashio::log.info "Using manually configured network interface: ${custom_interface}"
+        primary_interface="$custom_interface"
+    else
+        bashio::log.error "The configured network interface '${custom_interface}' was not found on the host!"
+        bashio::log.warning "Falling back to automatic interface detection."
+    fi
 fi
 
-# shellcheck disable=SC2086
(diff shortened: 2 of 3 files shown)


agners:
> Currently, trying to append `--primary-interface` via `matter_server_args` results in duplicate arguments where the automatically detected one takes precedence, making it impossible for users to override it.

At least for the Python Matter Server this is actually not true, see also https://github.com/home-assistant/addons/pull/4070#issuecomment-3050258473. From my perspective this flag does not make sense for this and the other reasons I brought forward in that PR.

As for the matter.js server I am not sure. But if we add this flag, I'd definitely do it imminent promotion of matter.js to default (#4642).

Apollon77:
Yeah matter.js works a bit different here. For us this is basically the main "MDNS listen and send interface". so yes it is basically also used to handle link local scoping but mainly because we should only get announces from that interface, so yes here it makes more sense.

@txptr we plan the relewae of the server 9.0 for next week and I would like to plan this in for after this. So it would be great if you could then next week after merging the 9.0 PR then add changelog for a 9.0.1

txptr:
@Apollon77 thank you. I will definitely do this.

agners:
@txptr what I would prefer is if we'd have a generic network interface isolation based on macvlan or ipvlan for apps which run on host network. This idea exists since a while, see https://github.com/home-assistant/architecture/discussions/1034. I created a prototype today, and at least in my testing it worked fine with Matter over Thread devices (see https://github.com/home-assistant/architecture/discussions/1034#discussioncomment-17281421).

Would this suite your use case as well?

txptr:
@Apollon77 @agners 

Thank you all for taking time for my pr and your insightful feedback.

I completely agree that the proposed larger overhaul - introducing native macvlan or isolated network support directly into the Supervisor - is conceptually the most elegant and robust long-term solution. It would fundamentally solve mDNS, multicast, and routing issues for complex addons across the entire ecosystem.

However, since implementing such a core architectural change in the Supervisor will understandably take considerable time to specify, develop, and thoroughly test, I would love to propose my current PR as a pragmatic, low-risk bridge solution (quick win).

My patch:
- Fixes an acute pain point for Matter users right now.
- Requires zero changes to the Supervisor core.
- Is fully backward-compatible.

Once the grand network reform for the Supervisor is ready and rolled out, the matter_server_primary_interface configuration flag can easily be deprecated and removed.

I am happy to adjust my PR based on whichever direction the core team decides to prioritize, but I believe delivering this relief to users today would be highly beneficial while we work towards the ideal architecture.

You decide :-)

PR #4625, opened 2026-05-23
Deprecate Assist Microphone app

omaramin-2000:
Added deprecation notice as wyoming satellite is deprecated and link to replacement [Assist Satellite](https://github.com/OHF-Voice/apps/tree/main/assist_satellite) app that uses [Linux Voice Assistant](https://github.com/OHF-Voice/linux-voice-assistant). 

## Summary by CodeRabbit

* **Documentation**
  * Added a prominent deprecation notice and migration guidance; Home Assistant app title now marked as deprecated.

* **Chores**
  * Add-on status set to deprecated in configuration.
  * Build workflow updated to skip deprecated add-ons during change detection.


sairon (changes requested):
Makes sense to deprecate now that upstream has been archived. Do you agree with the proposal, @synesthesiam?

One minor issue though is that changing `config,yaml` triggers a rebuild and re-publishing of the image, which may have side-effects (e.g. new non-pinned dependencies get pulled in). In that case we should either bump version (so we don't have two images with the same tag in the wild) or disable the build like we did in #4555.

synesthesiam:
I'm fine with this being marked as deprecated however you see fit @sairon  :slightly_smiling_face:

sairon (approved):
I adjusted the workflow to skip the build then, from my view, looks good now!

agners (approved):
LGTM!

synesthesiam:
Merge when ready :rocket:

Issue #4493, opened 2026-03-16
Let's Encrypt RFC2136 DNS Challenge post v6 requires RFC2136_TSIG_ALGORITHM syntax change


samualblair:
### Describe the issue you are experiencing

Noticed after upgrade to post v6 broken RFC2136

After digging into this, first had to remove (now unsupported 'sign') line or service wasn't fully starting.
Removed line:
rfc2136_sign_query: false

Then still saw error with 'invalid / unsupported signature'
rfc2136: unsupported TSIG algorithm: HMAC-SHA512.

But this was working before.
Digging in, found changes in back end require change in syntax.
Fixed by changing from:

To:
  rfc2136_algorithm: hmac-sha512.

Please See:
https://github.com/miekg/dns/blob/master/tsig.go#L18

Let's Encrypt RFC2136 DNS Challenge post v6.0 requires RFC2136_TSIG_ALGORITHM syntax change 

	HmacSHA1   = "hmac-sha1."
	HmacSHA224 = "hmac-sha224."
	HmacSHA256 = "hmac-sha256."
	HmacSHA384 = "hmac-sha384."
	HmacSHA512 = "hmac-sha512."

At minimum documentation should be updated to stop reflecting old (this is currently listed):
```
email: your.email@example.com
domains:
  - your.domain.tld
certfile: fullchain.pem
keyfile: privkey.pem
challenge: dns
dns:
  provider: dns-rfc2136
  rfc2136_server: dns-server.dom.ain
  rfc2136_port: '53'
  rfc2136_name: letsencrypt
  rfc2136_secret: "secret-key"
  rfc2136_algorithm: HMAC-SHA512
```

and update to current
```
email: your.email@example.com
domains:
  - your.domain.tld
certfile: fullchain.pem
keyfile: privkey.pem
challenge: dns
dns:
  provider: dns-rfc2136
  rfc2136_server: dns-server.dom.ain
  rfc2136_port: '53'
  rfc2136_name: letsencrypt
  rfc2136_secret: "secret-key"
  rfc2136_algorithm: hmac-sha512.
```

Would be even better for others if this could be caught and patched for them if incorrect.

### What type of installation are you running?

Home Assistant OS

### Which operating system are you running on?

Home Assistant Operating System

### Which app are you reporting an issue with?

Let's Encrypt

### What is the version of the app?

6.1.1

### Steps to reproduce the issue

[...]

PR #4633, opened 2026-05-28
Docs: fix example for ACME `rfc2136_algorithm`

NicolaiSoeborg:
Ref: https://github.com/home-assistant/addons/issues/4493

The switch-case handling "Algorithm" runs on [`CanonicalName(t.Algorithm)`](https://github.com/miekg/dns/blob/master/tsig.go#L44) so no need for the dot in the end.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced RFC2136 DNS challenge documentation with refined guidance on required settings. Clarified that `rfc2136_port` must be quoted, all `rfc2136_*` options are mandatory, and proper key configuration is essential for certbot DNS authentication setup.

### DIFF
--- a/letsencrypt/DOCS.md
+++ b/letsencrypt/DOCS.md
@@ -1358,7 +1358,7 @@ And ensure you have an update policy in place in the zone that uses this key to 
      };
   ```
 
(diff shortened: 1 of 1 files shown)


agners (approved):
Thank you for the fix! 

Since it is docs we don't need a changelog entry/version bump. LGTM!

PR #4635, opened 2026-05-29
openthread_border_router: update beta toggle description

TheJulianJES:
### Background

The stable and beta OTBR versions were both recently bumped with https://github.com/home-assistant/addons/pull/4619. Now, the beta toggle no longer unlocks "Thread 1.4" and the "native OpenThread mDNS", as that's already in the stable version. The beta just pins a newer OTBR commit.

### Change

So, This PR updates the description for the beta toggle. 

### Text choices

I'm not sure if the "experimental" part in the description potentially scares off users (or makes the text unnecessarily longer) – I'd also be fine with just "Enable beta mode to run a newer version of OpenThread [Border Router]." or similar. I think we should just get rid of the 1.4 and mDNS mention.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the description for the beta mode configuration option to better explain that it runs a newer, experimental version of the OpenThread Border Router.

### DIFF
--- a/openthread_border_router/DOCS.md
+++ b/openthread_border_router/DOCS.md
@@ -70,7 +70,7 @@ App configuration:
 | firewall           | Enable OpenThread Border Router firewall to block unnecessary traffic |
 | nat64              | Enable NAT64 to allow Thread devices accessing IPv4 addresses |
 | network_device     | IP address and port to connect to a network-based RCP (see below) |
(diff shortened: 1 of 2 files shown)


puddly (approved):
Good catch, thank you!

agners (approved):
LGTM, thanks!

PR #4642, opened 2026-06-02
matter_server: migrate to matter.js server (9.0.0)

### LINKED ISSUE #4627: Matter server beta not starting during internet outage

I enabled matter.js beta (to fix IKEA firmware otherwise not updating). During a recent unplanned internet outage and restarting Home Assistant Green, I found all my Matter devices disabled. Checking the log I noticed the Matter server stalled at wanting to install the latest matter server from npm. 

Effectively:
[07:34:20] INFO: Installing latest matter-server from npm...
npm error code ETIMEDOUT
...
[07:35:43] ERROR: Failed to install Matter Server from npm. Aborting startup.

I originally posted as an issue at [https://github.com/matter-js/matterjs-server/issues/678#issue-4505740446](url). @Apollon77 kindly replied and advised this is likely because of the way beta works, hence reposting here.

The issue was that I lost access to all my locally based Matter devices because I didn’t have internet (which feels a bit ironic 😊). In this situation I would have expected the system to continue using the most recently installed beta.

Home Assistant OS

Matter Server

8.5 (beta enabled)

1. Matter Server App - Enable "Use the latest beta version"
2. Start server and once installed, Stop server
3. Disconnect internet
4. Start Matter Server

## System Information

version | core-2026.5.3
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.14.2
os_name | Linux
os_version | 6.12.85-haos
arch | aarch64
timezone | Australia/Sydney
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 4999
Installed Version | 2.0.5
Stage | running
Available Repositories | 3126
Downloaded Repositories | 34

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 17.3
-- | --
update_channel | stable
supervisor_version | supervisor-2026.05.0
agent_version | 1.8.1
docker_version | 29.3.1
disk_total | 28.0 GB
disk_used | 13.7 GB
nameservers | 192.168.69.1
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | green
supervisor_api | ok
version_api | ok
installed_addons | Matter Server (8.5.0), Mosquitto broker (7.1.0), MQTT Explorer (browser-1.0.3), Filebrowser (2.63.5), Zigbee2MQTT (2.10.1-1), OpenThread Border Router (2.16.8), Cloudflared (7.0.7), ESPHome Device Builder (2026.5.0), Advanced SSH & Web Terminal (23.0.9), Node-RED (21.0.10)

</details>

<details><summary>Dashboards</summary>

dashboards | 7
-- | --
resources | 22
views | 12
mode | storage

</details>

<details><summary>Network Configuration</summary>

adapters | lo (disabled), end0 (enabled, default, auto), docker0 (disabled), hassio (disabled), vethd0aa429 (disabled), veth0e05b44 (disabled), veth532b0d3 (disabled), veth1c8c5b7 (disabled), veth7b3fd21 (disabled), veth0ffc2de (disabled), wpan0 (disabled), veth12de98b (disabled), vethf09636c (disabled), veth5f3573e (disabled)
-- | --
ipv4_addresses | lo (127.0.0.1/8), end0 (192.168.69.99/24), docker0 (172.30.232.1/23), hassio (172.30.32.1/23), vethd0aa429 (), veth0e05b44 (), veth532b0d3 (), veth1c8c5b7 (), veth7b3fd21 (), veth0ffc2de (), wpan0 (), veth12de98b (), vethf09636c (), veth5f3573e ()
ipv6_addresses | lo (::1/128), end0 (fd0a:5c1d:e2f8:3b47:d8ec:d617:377f:977f/64, fe80::9925:c16e:3659:9c8b/64), docker0 (fe80::348c:5cff:feac:1bf6/64), hassio (fd0c:ac1e:2100::1/48, fe80::501c:73ff:fe25:bd1b/64), vethd0aa429 (fe80::3034:46ff:feee:f547/64), veth0e05b44 (fe80::478:5dff:fece:88a2/64), veth532b0d3 (fe80::1ce3:62ff:feb3:11e7/64), veth1c8c5b7 (fe80::b02e:d9ff:fea6:7349/64), veth7b3fd21 (fe80::8e0:74ff:fecb:4cea/64), veth0ffc2de (fe80::7428:3ff:fedb:97c0/64), wpan0 (fddd:c8c1:74d2:fa0d:0:ff:fe00:fc11/64, fd1b:e245:f84a:1:c4e9:6f89:c446:1449/64, fddd:c8c1:74d2:fa0d:0:ff:fe00:fc38/64, fddd:c8c1:74d2:fa0d:0:ff:fe00:fc10/64, fddd:c8c1:74d2:fa0d:0:ff:fe00:ac00/64, fddd:c8c1:74d2:fa0d:40cb:f240:7b87:db1b/64, fe80::4835:1763:27fb:ceaa/64), veth12de98b (fe80::5091:46ff:fef6:54f/64), vethf09636c (fe80::ec57:65ff:fea5:20d4/64), veth5f3573e (fe80::ef:1cff:fe14:def8/64)
announce_addresses | 192.168.69.99, fd0a:5c1d:e2f8:3b47:d8ec:d617:377f:977f, fe80::9925:c16e:3659:9c8b

</details>

<details><summary>Recorder</summary>

oldest_recorder_run | 14 May 2026 at 3:46 pm
-- | --
current_recorder_run | 24 May 2026 at 7:04 am
estimated_db_size | 553.74 MiB
database_engine | sqlite
database_version | 3.49.2

</details>

```txt

```

```txt

```

### Additional information

_No response_

Apollon77:
## Summary

Promotes the Matter Server add-on to **9.0.0**, dropping the Python Matter Server entirely and building on the official matter.js server image.

- Base image → `ghcr.io/matter-js/matterjs-server:0.8.0` (node 24, `matter-server` pre-installed); the manual Node.js install and build toolchain are removed from the image.
- The bundled server runs as-is. The **Beta** flag installs `matter-server@latest` from npm; `matter_server_version` pins a specific npm version (takes precedence over Beta). A major version >= 3 is ignored as a leftover Python version and the bundled server is used.
- npm installs happen in-place in `/app` with the build toolchain installed at runtime only when needed; any apt/npm failure logs a warning and starts the bundled server instead (no hard fail).
- Drop python-only CLI args (`--log-level-sdk`, `--paa-root-cert-dir`) and options (`log_level_sdk`, `matter_sdk_wheels_version`).
- `log_level` schema updated to the matter.js levels (`debug|info|notice|warn|error|fatal`); legacy `verbose|warning|critical` still accepted and mapped.
- Delete a former beta `.migrations` backup directory on startup.
- Keeps the 8.5.0 `ble_proxy` option and `matter_server_env_vars`; `--ota-provider-dir` stays conditional on `enable_test_net_dcl`.
- Add `MIGRATION_FAQ.md`; update DOCS/CHANGELOG/RELEASE.

## Verification

- `shellcheck` clean on `run`; `config.yaml` / `build.yaml` valid YAML.
- 0.8.0 image config confirmed: entrypoint path, `/app` workdir, and `USER 1000` (hence the `USER root` override) all match the run script.
- Live build / commission / migration test is the maintainer gate.

## Summary by CodeRabbit

* **New Features**
  * Matter.js-based Matter Server now runs by default with automatic data migration (slower first start), plus experimental BLE proxy commissioning support
  * Web UI dashboard adds Thread and Wi‑Fi network visualizations

* **Documentation**
  * Added a migration FAQ and stronger full-backup/first-start guidance
  * Updated configuration, version-selection, and Beta/npm fallback behavior documentation and option text

* **Chores**
  * Updated the add-on to 9.0.0, refreshed server image targets, and adjusted container startup/log/config options (Python-specific items removed)

closes #4627

### DIFF
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -13,7 +13,7 @@
   - Coming from Beta server < 0.7: the first start runs the storage data migration and the second start removes the migration fallback data. Both can take a while — be patient on both starts.
 - New Matter Server features:
   - The Web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
-  - BLE proxy support is enabled automatically when no local Bluetooth adapter ID is configured
+  - BLE proxy support is available via the `ble_proxy` option, letting Home Assistant drive BLE commissioning through its own bluetooth stack
   - Otherwise, it offers the same functionality as the Python server, and behavior within Home Assistant should be unchanged
   - For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
 - The **Beta** option now installs the latest `matter-server` from npm on top of the bundled version
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - **The Matter Server now runs on matter.js. Expect the first start to take noticeably longer while your data migrates automatically — this is normal.**
   - **⚠️ It is recommended to take a backup before updating.**
+  - The new matter server needs roughly twice the RAM of the old one, so please check free resources before updating.
   - The Python Matter Server has been replaced with a matter.js-based implementation; your data migrates automatically with no action required
   - Later starts are much faster thanks to matter.js optimizations
 - New Matter Server features:
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -18,6 +18,8 @@
   - For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
 - The **Beta** option now installs the latest `matter-server` from npm on top of the bundled version
   - npm install failures are logged as a warning and the bundled version is started instead
+  - The **Beta** flag keeps its previous value across this update. If you enabled it for the earlier matter.js beta, it stays on. matter.js is now the default, so you can turn it off — but more betas will follow, so keep it on if you want to keep testing pre-releases.
+- The `bluetooth_adapter_id` option is now deprecated in favor of `ble_proxy`, which lets Home Assistant drive BLE commissioning through its own bluetooth stack. `bluetooth_adapter_id` still works for now.
 - `matter_server_version` now installs that specific `matter-server` version from npm (takes precedence over **Beta**); values with a major version >= 3 are ignored as leftover Python Matter Server versions, and the bundled version is used instead
 - Removed the `log_level_sdk` and `matter_sdk_wheels_version` options (Python Matter Server only)
 - Updated the `log_level` options to the matter.js levels (`debug`, `info`, `notice`, `warn`, `error`, `fatal`); the legacy `verbose`/`warning`/`critical` values are still accepted and mapped automatically
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -2,27 +2,15 @@
 
 ## 9.0.0
 
-- **The Matter Server now runs on matter.js by default. Expect the first start to take noticeably longer while your data migrates automatically — this is normal. Let the migration finish before using Matter.** Details below.
-  - **⚠️ Consider a full backup before updating.**
-  - The add-on now builds on the official `matterjs-server` Docker image
-  - The Python Matter Server has been removed
-  - Your existing data migrates automatically; no action is required
-  - The first start re-discovers and fully interviews every node. Later starts are much faster because each node's last-known IP is reused and usually only a partial interview is needed.
-- **For former Beta testers:**
-  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data. Expect it to take a while.
-  - Coming from Beta server < 0.7: the first start runs the storage data migration, and the second start removes the migration fallback data. Expect both starts to take a while.
+- **The Matter Server now runs on matter.js. Expect the first start to take noticeably longer while your data migrates automatically — this is normal.**
+  - **⚠️ It is recommended to take a backup before updating.**
+  - The Python Matter Server has been replaced with a matter.js-based implementation; your data migrates automatically with no action required
+  - Later starts are much faster thanks to matter.js optimizations
 - New Matter Server features:
-  - The Web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
-  - BLE proxy support is available via the `ble_proxy` option, letting Home Assistant drive BLE commissioning through its own bluetooth stack
-  - Otherwise, it offers the same functionality as the Python server, and behavior within Home Assistant should be unchanged
-  - For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
-- The **Beta** option now installs the latest `matter-server` from npm on top of the bundled version
-  - npm install failures are logged as a warning and the bundled version is started instead
-  - The **Beta** flag keeps its previous value across this update. If you enabled it for the earlier matter.js beta, it stays on. matter.js is now the default, so you can turn it off — but more betas will follow, so keep it on if you want to keep testing pre-releases.
-- The `bluetooth_adapter_id` option is now deprecated in favor of `ble_proxy`, which lets Home Assistant drive BLE commissioning through its own bluetooth stack. `bluetooth_adapter_id` still works for now.
-- `matter_server_version` now installs that specific `matter-server` version from npm (takes precedence over **Beta**); values with a major version >= 3 are ignored as leftover Python Matter Server versions, and the bundled version is used instead
-- Removed the `log_level_sdk` and `matter_sdk_wheels_version` options (Python Matter Server only)
-- Updated the `log_level` options to the matter.js levels (`debug`, `info`, `notice`, `warn`, `error`, `fatal`); the legacy `verbose`/`warning`/`critical` values are still accepted and mapped automatically
+  - The web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
+  - Experimental BLE proxy support via the `ble_proxy` option, for commissioning via BLE proxies from the Matter Server dashboard
+- If you enabled the **Beta** flag for the earlier matter.js beta, it stays on; matter.js is now the default, so you can turn it off — keep it on to keep testing pre-releases
+- For more details, see the [Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
 
 ## 8.5.0
 
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -5,8 +5,8 @@
 - **The Matter Server now runs on matter.js. Expect the first start to take noticeably longer while your data migrates automatically — this is normal.**
   - **⚠️ It is recommended to take a backup before updating.**
   - The new matter server needs roughly twice the RAM of the old one, so please check free resources before updating.
-  - The Python Matter Server has been replaced with a matter.js-based implementation; your data migrates automatically with no action required
-  - Later starts are much faster thanks to matter.js optimizations
+  - The Python Matter Server has been replaced with a matter.js-based implementation (v1.1.0); your data migrates automatically with no action required
+  - Later starts are much faster
 - New Matter Server features:
   - The web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
   - Experimental BLE proxy support via the `ble_proxy` option, for commissioning via BLE proxies from the Matter Server dashboard
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -2,15 +2,15 @@
 
 ## 9.0.0
 
-- **The Matter Server now runs on matter.js by default. Your first start will take noticeably longer while your data migrates automatically. This is normal: please be patient and let it finish.** Details below.
+- **The Matter Server now runs on matter.js by default. Expect the first start to take noticeably longer while your data migrates automatically — this is normal. Let the migration finish before using Matter.** Details below.
   - **⚠️ Consider a full backup before updating.**
   - The add-on now builds on the official `matterjs-server` Docker image
   - The Python Matter Server has been removed
   - Your existing data migrates automatically; no action is required
   - The first start re-discovers and fully interviews every node. Later starts are much faster because each node's last-known IP is reused and usually only a partial interview is needed.
 - **For former Beta testers:**
-  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data, which can take a while. Be patient.
-  - Coming from Beta server < 0.7: the first start runs the storage data migration and the second start removes the migration fallback data. Both can take a while — be patient on both starts.
+  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data. Expect it to take a while.
+  - Coming from Beta server < 0.7: the first start runs the storage data migration, and the second start removes the migration fallback data. Expect both starts to take a while.
 - New Matter Server features:
   - The Web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
   - BLE proxy support is available via the `ble_proxy` option, letting Home Assistant drive BLE commissioning through its own bluetooth stack
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## 9.0.0
+
+- **The Matter Server now runs on matter.js by default. Your first start will take noticeably longer while your data migrates automatically. This is normal: please be patient and let it finish.** Details below.
+  - **⚠️ Consider a full backup before updating.**
+  - The add-on now builds on the official `matterjs-server` Docker image
+  - The Python Matter Server has been removed
+  - Your existing data migrates automatically; no action is required
+  - The first start re-discovers and fully interviews every node. Later starts are much faster because each node's last-known IP is reused and usually only a partial interview is needed.
+- **For former Beta testers:**
+  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data, which can take a while. Be patient.
+  - Coming from Beta server < 0.7: the first start runs the storage data migration and the second start removes the migration fallback data. Both can take a while — be patient on both starts.
+- New Matter Server features:
+  - The Web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
+  - BLE proxy support is enabled automatically when no local Bluetooth adapter ID is configured
+  - Otherwise, it offers the same functionality as the Python server, and behavior within Home Assistant should be unchanged
+  - For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
+- The **Beta** option now installs the latest `matter-server` from npm on top of the bundled version
+  - npm install failures are logged as a warning and the bundled version is started instead
+- `matter_server_version` now installs that specific `matter-server` version from npm (takes precedence over **Beta**); values with a major version >= 3 are ignored as leftover Python Matter Server versions, and the bundled version is used instead
+- Removed the `log_level_sdk` and `matter_sdk_wheels_version` options (Python Matter Server only)
+- Updated the `log_level` options to the matter.js levels (`debug`, `info`, `notice`, `warn`, `error`, `fatal`); the legacy `verbose`/`warning`/`critical` values are still accepted and mapped automatically
+
 ## 8.5.0
 
 - Add `ble_proxy` option to expose the Matter Server's BLE proxy endpoint, so the Home Assistant Matter integration can drive BLE commissioning through Home Assistant's bluetooth stack (including ESPHome BLE proxies). To actually use this BLE proxy you need Home Assistant 2026.06 or later (the Matter integration support landed there) and the Beta Matter Server (>= 0.7.1, JavaScript-based); mutually exclusive with `bluetooth_adapter_id` (the local adapter is ignored with a warning when `ble_proxy` is enabled).
--- a/matter_server/DOCS.md
+++ b/matter_server/DOCS.md
@@ -51,7 +51,7 @@ App configuration:
 | bluetooth_adapter_id | Set BlueZ Bluetooth Controller ID (for local commissioning).                                            |
 | matter_server_args   | Extra command-line arguments passed to the Matter Server at startup (advanced).                         |
 | matter_server_env_vars | Extra environment variables exported before the server starts. Use `KEY=VALUE` entries. |
-| matter_server_version | Install this specific `matter-server` version from npm (advanced; takes precedence over **Beta**). Values with a major version >= 3 are ignored (these are Python Matter Server versions) and the bundled version is used. |
+| matter_server_version | Install this specific `matter-server` version from npm (advanced; takes precedence over **Beta**). Values with a major version >= 3 are ignored (these are Python Matter Server versions); unless **Beta** is set, the bundled version is used. |
 
 ### `matter_server_env_vars`
 
--- a/matter_server/DOCS.md
+++ b/matter_server/DOCS.md
@@ -38,6 +38,12 @@ The server version bundled in the app image is used by default. The **Beta**
 flag installs the latest `matter-server` from npm on top of it, and
 `matter_server_version` installs a specific version (see below).
 
+> [!NOTE]
+> The **Beta** flag keeps its previous value across this update. If you enabled
+> it for the earlier matter.js beta, it stays on after updating to 9.0.0.
+> matter.js is now the default, so you can turn it off — but more betas will
+> follow, so keep it on if you want to keep testing pre-releases.
+
 ## Configuration
 
 App configuration:
@@ -48,7 +54,7 @@ App configuration:
 | beta                 | Install the latest `matter-server` from npm on startup instead of the bundled version. On failure a warning is logged and the bundled version is started. |
 | enable_test_net_dcl  | Enable test-net DCL for PAA root certificates, OTA updates and other device information.                |
 | ble_proxy            | Expose the BLE proxy endpoint so the Home Assistant Matter integration can drive BLE commissioning through Home Assistant's bluetooth stack. Mutually exclusive with `bluetooth_adapter_id`. |
-| bluetooth_adapter_id | Set BlueZ Bluetooth Controller ID (for local commissioning).                                            |
+| bluetooth_adapter_id | **Deprecated** — use `ble_proxy` instead. Set BlueZ Bluetooth Controller ID (for local commissioning). Still works for now. |
 | matter_server_args   | Extra command-line arguments passed to the Matter Server at startup (advanced).                         |
 | matter_server_env_vars | Extra environment variables exported before the server starts. Use `KEY=VALUE` entries. |
 | matter_server_version | Install this specific `matter-server` version from npm (advanced; takes precedence over **Beta**). Values with a major version >= 3 are ignored (these are Python Matter Server versions); unless **Beta** is set, the bundled version is used. |
--- a/matter_server/DOCS.md
+++ b/matter_server/DOCS.md
@@ -26,11 +26,17 @@ Matter Server WebSocket server port field.
 
 ## Server variants
 
-This app runs the Python Matter Server implementation from the
-[home-assistant-libs/python-matter-server][matter_server_repo] repository by default.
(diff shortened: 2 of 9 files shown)


matter_server/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 8.5.0
+version: 9.0.0
```
TheJulianJES, line 2: I think we also want to mark this update as "breaking" internally. As far as I know, it just prevents an automatic update to this version.
```suggestion
version: 9.0.0
breaking_versions: [9.0.0]
```
This is similar to what we did for the OTBR 1.4 bump: https://github.com/home-assistant/addons/pull/4619/changes#r3276709222

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+
+- **The Matter Server now runs on matter.js by default. Your first start will take noticeably longer while your data migrates automatically. This is normal: please be patient and let it finish.** Details below.
+  - **⚠️ Consider a full backup before updating.**
+  - The add-on now builds on the official `matterjs-server` Docker image
+  - The Python Matter Server has been removed
+  - Your existing data migrates automatically; no action is required
+  - The first start re-discovers and fully interviews every node. Later starts are much faster because each node's last-known IP is reused and usually only a partial interview is needed.
+- **For former Beta testers:**
+  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data, which can take a while. Be patient.
+  - Coming from Beta server < 0.7: the first start runs the storage data migration and the second start removes the migration fallback data. Both can take a while — be patient on both starts.
```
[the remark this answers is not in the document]
Apollon77: Done in 90cf8cb — reworded the 9.0.0 migration notes to a direct tone ("Expect the first start to take longer…", "Let the migration finish…") and dropped the "please be patient" phrasing.

matter_server/DOCS.md
```diff
@@ -39,31 +45,21 @@ App configuration:
...
-| matter_server_env_vars | Extra environment variables exported before start (only relevant for the JavaScript Matter Server / Beta mode). Use `KEY=VALUE` entries. |
-| matter_server_version | Custom Matter Server version. In JavaScript/Beta mode, only `0.x.y` or `1.x.y` values are used; other values are ignored and latest is installed. |
-
-### `matter_server_env_vars` (JavaScript server only)
+| ble_proxy            | Expose the BLE proxy endpoint so the Home Assistant Matter integration can drive BLE commissioning through Home Assistant's bluetooth stack. Mutually exclusive with `bluetooth_adapter_id`. |
+| bluetooth_adapter_id | Set BlueZ Bluetooth Controller ID (for local commissioning).                                            |
+| matter_server_env_vars | Extra environment variables exported before the server starts. Use `KEY=VALUE` entries. |
+| matter_server_version | Install this specific `matter-server` version from npm (advanced; takes precedence over **Beta**). Values with a major version >= 3 are ignored (these are Python Matter Server versions) and the bundled version is used. |
```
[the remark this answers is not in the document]
Apollon77: Done in 90cf8cb — added a `matter_server_args` row to the config table.

matter_server/MIGRATION_FAQ.md
```diff
@@ -0,0 +1,37 @@
...
+The new Matter Server is a complete rewrite of the Python-based Matter Server. It is now built on the matter.js library, a JavaScript implementation of the Matter standard. The new Matter Server is more performant, more stable, and more feature-complete than the old one.
+
+Version 9.0 is the first release of the new Matter Server.
+
+## Should I do a backup before installing the new version?
+Yes. We always recommend making a backup before installing a new version of the server. The new version migrates your current data automatically, but it is always better to be safe than sorry. Just tick the "Create backup" checkbox when you update the server.
+
+## What happens on the first start of the new version?
+The first start of the new version **automatically migrates your current data**. This requires no manual intervention but needs some time to complete. **Please be patient.** The log shows the progress of the migration.
+
```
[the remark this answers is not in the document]
Apollon77: Done in 469d210 — applied the direct, action-first tone to the backup, first-start, and "report issues" sections.

matter_server/config.yaml
Apollon77: Done in 469d210 — added `breaking_versions: [9.0.0]`. Thanks, that matches the OTBR/letsencrypt/mariadb precedent and is exactly what we want for this major migration.

matter_server/Dockerfile
```diff
@@ -1,6 +1,9 @@
 ARG BUILD_FROM
 FROM ${BUILD_FROM}
 
+# The matter.js base image runs as an unprivileged user; s6-overlay needs root.
+USER root
```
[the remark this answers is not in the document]
Apollon77: Intentional. Every other add-on in this repo builds on a `ghcr.io/home-assistant/*-base*` image, which runs as root by default, and none of them drop privileges (no `s6-setuidgid`/`su-exec`/`gosu` anywhere in the repo). `matter_server` is the only add-on not on a HA base image — the upstream `matterjs-server` image sets `USER 1000` for standalone `docker run`, so the explicit `USER root` simply brings it in line with every sibling add-on (s6 `/init` needs root). The `# hadolint ignore=DL3002` covers the only privilege linter in this repo's CI (lint.yml runs hadolint/yamllint/shellcheck; Trivy is not a CI gate here). Dropping the server to UID 1000 would also need re-validation of BLE/D-Bus access and `/data`+`/app` ownership, so keeping root is the consistent, tested choice.

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -25,90 +23,75 @@ if ! bashio::config.exists log_level; then
...
-      env_name="${env_entry%%=*}"
-      env_value="${env_entry#*=}"
-
-      if ! [[ "${env_name}" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]]; then
-        bashio::exit.nok "Invalid environment variable name in matter_server_env_vars: ${env_name}"
-      fi
+fi
+if ! bashio::var.has_value "${install_target}" && bashio::config.true "beta"; then
+  install_target="matter-server@latest"
+fi
```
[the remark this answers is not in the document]
Apollon77: By design. A rejected `matter_server_version` (major >= 3) is treated as "unset" and the normal resolution continues. `beta: true` is an independent, explicit opt-in to `@latest`, so it still wins in that case — a user who set `beta: true` asked for latest. The common real-world case (leftover Python version + default `beta: false`) correctly falls back to the bundled server. I clarified the DOCS wording in 469d210: "…unless **Beta** is set, the bundled version is used."

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -25,90 +23,75 @@ if ! bashio::config.exists log_level; then
...
+  bashio::log.info "Installing ${install_target} from npm over the bundled server..."
+  # matter.js native modules may need compiling; the base image purges the
+  # toolchain, so add it back here. A failure here must not abort startup.
+  if ! { apt-get update \
+      && apt-get install -y --no-install-recommends \
+        python3 make gcc g++ libbluetooth-dev libudev-dev \
+      && cd /app \
+      && npm install --omit=dev --foreground-scripts "${install_target}"; }; then
+    bashio::log.warning \
+      "Failed to install ${install_target} from npm. Starting the bundled Matter Server instead."
```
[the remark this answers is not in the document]
Apollon77: Deliberate tradeoff — we chose in-place `/app` over a staging dir to avoid doubling the (large) `node_modules` at runtime. npm resolves and downloads to its cache before mutating `node_modules`, so the realistic failure modes here — invalid/`404` version, registry/network down, missing build toolchain — abort before `/app` is touched, leaving the bundled server intact for the fallback. The only window that could leave `/app` half-updated is a mid-reify kill (OOM/disk-full), which we accept as a rare, beta-only risk in exchange for not doubling image/runtime size.

matter_server/Dockerfile
```diff
@@ -1,6 +1,10 @@
 ARG BUILD_FROM
 FROM ${BUILD_FROM}
 
+# The matter.js base image runs as an unprivileged user; s6-overlay needs root.
+# hadolint ignore=DL3002
+USER root
```
agners, line 7: From my perspective it is fine to use `USER root` if we need root (and we did default to root before since the base container used to be run as root).

However, I'd frame it more around the fact that Noble in it's default configuration uses Bluetooth using raw hci access, which requires root. s6-overlay since v3 can be run without root:


```suggestion
# The matter.js base image runs as an unprivileged user; the app requires root for Bluetooth via Noble raw hci access
# hadolint ignore=DL3002
USER root
```

That said, maybe we should consider using Noble BlueZ backend. We do run BlueZ on HAOS on the same Bleutooth adapter, if HA is actively using it, commissioning might fail....

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+  - The first start re-discovers and fully interviews every node. Later starts are much faster because each node's last-known IP is reused and usually only a partial interview is needed.
+- **For former Beta testers:**
+  - Coming from Beta server 0.7.x/0.8.x: the first start removes the migration fallback data. Expect it to take a while.
+  - Coming from Beta server < 0.7: the first start runs the storage data migration, and the second start removes the migration fallback data. Expect both starts to take a while.
+- New Matter Server features:
+  - The Web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
+  - BLE proxy support is available via the `ble_proxy` option, letting Home Assistant drive BLE commissioning through its own bluetooth stack
+  - Otherwise, it offers the same functionality as the Python server, and behavior within Home Assistant should be unchanged
+  - For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
+- The **Beta** option now installs the latest `matter-server` from npm on top of the bundled version
```
agners: Should we maybe note to existing beta user to consider disable the Beta option if they no longer want to participate?

For the Thread BR we forced it off (in a way reset all beta users back to non-beta). Maybe we should do the same here? 🤔

matter_server/Dockerfile
Apollon77: For me with the introduction of the "BLE proxy" we should more go the way to "deprecate" the "define a BLuetooth adapter" (and so bare usage of HCI) and remove in a "not too far but when we know that BLE proxy is stable" timeframe and do a v10 for this then in some months. Then we could add the deprecation already here but leave it in, so it is "endable".
This then together with intermediate "default enable ble proxy when no adapter is set" as it was alreayd but I reverted because we should get the ble proxy rocking stable first

matter_server/CHANGELOG.md
Apollon77: Nooo ... I think beta people are happy to still do beta here because it was a definitive "opt in" already and valuable feedback, but yes adding a note here I would do, force.drop not

matter_server/Dockerfile
agners: Yeah once BLE proxy is stable we can deprecate bare HCI usage.

I wonder if we get it that stable fast enough. And I wonder if Nomble via BlueZ is maybe more stable than raw hci... From 4 tries I was able to commission once on a RPi here :thinking: .

Anyhow, I think for the 9.0.0 release lets not introduce a change. Let's just stick with what we used with the matter.js server as default during beta.

Apollon77: text adjusted

agners (approved):
LGTM!

matter_server/Dockerfile
```diff
@@ -1,6 +1,12 @@
...
 FROM ${BUILD_FROM}
 
+# The matter.js base image runs as an unprivileged user; s6-overlay needs root.
+# hadolint ignore=DL3002
+# The matter.js base image runs as an unprivileged user; the app requires root for Bluetooth via Noble raw hci access
+# hadolint ignore=DL3002
+USER root
+
 # Default ENV
 ENV \
```
agners, line 10: In my testing I get this error:

```
s6-rc: info: service s6rc-oneshot-runner: starting
s6-rc: info: service s6rc-oneshot-runner successfully started
s6-rc: info: service fix-attrs: starting
s6-rc: info: service banner: starting
s6-rc: info: service fix-attrs successfully started
s6-rc: info: service legacy-cont-init: starting
s6-rc: info: service legacy-cont-init successfully started
/usr/lib/bashio/log.sh: line 107: info: unbound variable
s6-rc: warning: unable to start service banner: command exited 1
/run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
```

The reason seems to be that the matter.js Server we base on now uses `LOG_LEVEL` with a textual value. But bash.io uses the same environment variable but with numeric values 😰 

What we can do here is just clear the environment variable to trigger the default:

```suggestion
ENV \
    LOG_LEVEL="" \
```

agners (changes requested):
Built and tested on a aarch64 platform, and found one minor issue still.

agners (approved):

matter_server/translations/en.yaml
```diff
@@ -53,17 +51,11 @@ configuration:
...
-      format! Use this feature with caution! Make sure you have a recent backup
-      of the app!
-  matter_sdk_wheels_version:
-    name: Matter SDK wheels version
-    description: >-
-      Install custom Python Matter SDK wheels version (not applicable when
-      using Beta flag). NOTE: The API might not be compatible with the Python
-      Matter server.
+      Install a specific `matter-server` version from npm (takes precedence over
+      the Beta flag). Only `0.x.y` or `1.x.y` values are used; values with a
```
TheJulianJES: nit: HA doesn't render text with backticks as code blocks for addons/apps, so IMO they should be removed everywhere in the addon/app configs. At most, we can replace them with normal `"` or `'`.

Like, I don't think this looks good at all – it's probably not what's intended either:
Screenshot showing HA Matter server addon configuration with this text: `0.x.y` or `1.x.y`. The backticks are rendered as-is. They do not cause the text inside to render as a codeblock, which is likely what's intended.

(This is the case in many other locations here, but I've only flagged this example for clarity. It's easy enough to search for the others, or let Claude fix everything.)

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
+## 9.0.0
+
+- **The Matter Server now runs on matter.js by default. Expect the first start to take noticeably longer while your data migrates automatically — this is normal. Let the migration finish before using Matter.** Details below.
+  - **⚠️ Consider a full backup before updating.**
+  - The add-on now builds on the official `matterjs-server` Docker image
```
TheJulianJES: nit: We probably want to use the new "app" terminology.

matter_server/translations/en.yaml
```diff
@@ -3,24 +3,24 @@ configuration:
...
     description: >-
       Enable PAA root certificates, software updates and other device
       information from test-net DCL. This is meant for development and testing
       purposes.
   bluetooth_adapter_id:
-    name: Bluetooth Adapter ID
+    name: Bluetooth Adapter ID (deprecated)
     description: >-
+      DEPRECATED: Use the `Enable BLE proxy` option instead, which lets Home
+      Assistant drive BLE commissioning through its own bluetooth stack
```
TheJulianJES: nit: bluetooth -> Bluetooth
```suggestion
      Assistant drive BLE commissioning through its own Bluetooth stack
```

matter_server/CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+  - npm install failures are logged as a warning and the bundled version is started instead
+  - The **Beta** flag keeps its previous value across this update. If you enabled it for the earlier matter.js beta, it stays on. matter.js is now the default, so you can turn it off — but more betas will follow, so keep it on if you want to keep testing pre-releases.
+- The `bluetooth_adapter_id` option is now deprecated in favor of `ble_proxy`, which lets Home Assistant drive BLE commissioning through its own bluetooth stack. `bluetooth_adapter_id` still works for now.
+- `matter_server_version` now installs that specific `matter-server` version from npm (takes precedence over **Beta**); values with a major version >= 3 are ignored as leftover Python Matter Server versions, and the bundled version is used instead
+- Removed the `log_level_sdk` and `matter_sdk_wheels_version` options (Python Matter Server only)
+- Updated the `log_level` options to the matter.js levels (`debug`, `info`, `notice`, `warn`, `error`, `fatal`); the legacy `verbose`/`warning`/`critical` values are still accepted and mapped automatically
```
TheJulianJES: nit: I do have to say this changelog is huge. Like, this is what a user sees (assuming they're viewing the update entity like most do, and do not go through the addon/app page). On Desktop:

Update entity of Matter addon showing long release notes


Couple of points:
1. I think the points regarding "For former Beta testers" could be removed almost entirely.
   - For end-users, it only matters that the start can take a while, but we already note that in the first line, so I think most people have that expectation if they're upgrading to this version.
   - For all current beta users, the starts are already very long anyway, because Matter.js-Server is reinstalled on every restart at the moment. So they already have that expectation for other reasons.
2. I would opt to not prominently mention the "BLE Proxy" option in the release notes.
   - This will be great once it's finished and when it hopefully works directly using the HA frontend, but currently, it only works when using the Matter Server UI.
   - And for now, you have to manually copy/paste your Thread dataset TLV there.
   - So I'm not sure the release notes are the best option for that. Users might have the expectation, that if this is mentioned in HA addon release notes, for this to work via HA itself.
4. IMO, the other points with `matter_server_version`, `log_level_sdk`, and `matter_sdk_wheels_versions` options could also be removed.
   - It's already mentioned in the beginning that the Python Server is removed, so I don't think we need to call out the removal of these already very technical options again..?
   - And we really don't need to mention the debug option of installing a specific Matter Server version IMO. Calling out the on/off beta option is completely fine, but no one will (or should) install a specific version, unless they're debugging a regression in Matter.js.

There's also a "Migration FAQ" linked. I think that can and should include more information than these release notes, even the more "technical stuff", which we don't need to bore normal users with IMO.

In general, I just think that if we have these huge release notes, that nobody is gonna read anything. IMO, we should mention the most important points only, then refer to the FAQ for more info.

TheJulianJES (commented):
Some optional feedback below – nothing blocking. IMO, we should look at the release notes to possibly trim them a bit. I think there's a lot of unnecessary info for most people in there now. That risks users skipping them entirely.

matter_server/CHANGELOG.md
TheJulianJES: Again, this is just optional feedback. But I think something like this would be enough for most users (and marks the "BLE Proxy" option as experimental, plus subverts expectations that it's available directly in HA now):

Screenshot of proposed shortened Matter release notes on Desktop


All users that are really interested in the technical details can go to the "Migration FAQ".

And for the vast majority of users, they are more likely to read the smaller release notes (IMO), so we have a good place to advertise matter.js but also make it clear everything migrates automatically. We're at least not scaring them away with a huge chunk of text anymore.

<details><summary>Markdown for that: (CLICK TO EXPAND)</summary>
<p>

```markdown
## 9.0.0

- **The Matter Server now runs on matter.js. Expect the first start to take noticeably longer while your data migrates automatically.**
  - **⚠️ It's recommended to take a backup before updating.**
  - The Python Matter Server has been replaced with a matter.js-based implementation
  - Future startups are much faster due to optimizations in matter.js
- New Matter Server features:
  - The web UI dashboard adds Thread and Wi-Fi network visualizations, plus many other improvements
  - Experimental BLE Proxy support is available via the `ble_proxy` option, allowing commissioning via BLE Proxies using the Matter Server dashboard
- If you previously tested matter.js by turning on the **Beta** flag, you can now turn it off if you do not want to receive future matter.js pre-releases
- For more details on the differences, see [our Migration FAQ](https://github.com/home-assistant/addons/blob/master/matter_server/MIGRATION_FAQ.md)
```

</p>
</details>

TheJulianJES: And just as an additional datapoint – this is what the release notes look like on an iPhone 12 Pro:

On the left are the current/long release notes, already scrolled down (skipping the header with icon, old/current version, and so on). There, you just have to scroll a ton – and there will be a lot of users who only look at these on their phones.

On the right are my proposed smaller release notes. These also fill the screen on an iPhone 12 Pro, but you don't have to scroll at all to read them:

| Current/long Matter release notes on iPhone 12 Pro, already scrolled down quite a bit  | Proposed shorter Matter release notes on iPhone 12 Pro, completely fitting on page, no scrolling required |
| ------------- | ------------- |
| Screenshot showing part of long Matter release notes on iPhone 12 Pro  | Screenshot showing entire shortened Matter release notes on iPhone 12 Pro  |

So I really think my suggested release notes are already at the maximum length we should use if we want users to realistically read them, especially on mobile.

Apollon77: Done in 4aca5cb — trimmed the 9.0.0 notes close to your proposed version (lead + backup + auto-migration + faster restarts + dashboard & experimental BLE proxy + a soft beta note + FAQ link). I moved the former-beta migration-timing detail into the Migration FAQ so it is not lost. Thanks for the mockups, very helpful for getting the length right.

matter_server/translations/en.yaml
Apollon77: Done in 4aca5cb — replaced the backticks with double quotes throughout `translations/en.yaml` (`config.yaml` had none). Left backticks in DOCS.md/CHANGELOG.md since those render as Markdown.

matter_server/CHANGELOG.md
Apollon77: The trimmed 9.0.0 notes (4aca5cb) no longer use the word "add-on". I kept "Matter Server" for the server itself rather than "app", since "app" refers to the add-on wrapper and would read oddly in those bullets. If you would like a doc-wide add-on→app rename (DOCS/FAQ/translations), I am happy to do that as a separate pass — just say the word.

Apollon77:
@TheJulianJES Changelog adjusted but also added the RAM notice from yesterday matter sync, so a bit longer than your proposal

matter_server/MIGRATION_FAQ.md
```diff
@@ -0,0 +1,43 @@
+# Python Matter Server v8.x --> matter.js-based Matter Server 9.0+ Migration FAQ
+
+## Why is there a new Matter Server?
+The new Matter Server is a complete rewrite of the Python-based Matter Server. It is now built on the matter.js library, a JavaScript implementation of the Matter standard. The new Matter Server is more performant, more stable, and more feature-complete than the old one.
+
+Version 9.0 is the first release of the new Matter Server.
```
MartinHjelmare: Version 9.0 is the app version, not the server version.

Apollon77: I rephrased to:
Version 9.0 of the Matter App is the first release for all Home Assistant users which uses the new Matter Server.

matter_server/config.yaml
```diff
@@ -25,12 +26,10 @@ map:
     read_only: false
 options:
   log_level: info
-  log_level_sdk: error
   beta: false
   enable_test_net_dcl: false
 schema:
-  log_level: list(verbose|debug|info|warning|error|critical)
-  log_level_sdk: list(automation|detail|progress|error|none)?
```
agners, line 33: This option

matter_server/config.yaml
```diff
@@ -40,7 +39,6 @@ schema:
   matter_server_env_vars:
     - match(^[A-Za-z_][A-Za-z0-9_]*=.*$)?
   matter_server_version: str?
-  matter_sdk_wheels_version: str?
```
agners, line 43: As well as this one should be deleted from the options in a startup script.

matter_server/config.yaml
Apollon77: Done in 148a731 — the startup script now deletes the removed options from a user's stored options on startup, using the pattern from the HA docs:

```bash
for old_key in log_level_sdk matter_sdk_wheels_version; do
  if bashio::jq.exists "${options}" ".${old_key}"; then
    bashio::log.info "Removing obsolete '${old_key}' option"
    bashio::addon.option "${old_key}"
  fi
done
```

So both `log_level_sdk` and `matter_sdk_wheels_version` are cleared and the Supervisor will not warn that they are no longer in the schema.

matter_server/config.yaml
Apollon77: Done in 148a731 — `matter_sdk_wheels_version` is deleted on startup together with `log_level_sdk` via `bashio::addon.option` (guarded by `bashio::jq.exists`), so neither triggers a schema warning. See the snippet in the sibling thread.

agners (approved):
LGTM!

PR #4660, opened 2026-06-22
Create vlc user and run vlc under the new user.

YukonStriker:
Current patching of the vlc binary in DockerFile poses a security risk.

VLC does not run under the root user because the source code explicitly contains a hardcoded security check that blocks execution if your Effective User ID

The Core Reason: Security Against Exploits

Media players process complex, untrusted, third-party media files and network streams. Historically, codecs and media demuxers have been vulnerable to buffer overflow and remote code execution vulnerabilities.

If an attacker crafts a malicious video file or network stream that exploits a bug in VLC while it runs as root, the attacker immediately gains full root control over the entire system. Because Docker containers run as root by default, a compromise inside the container drastically lowers the barrier for a container escape to compromise the underlying host system. To prevent this massive security risk, the VideoLAN developers chose to completely abort application startup if it detects a root environment.

## Release Notes

* **Chores**
  * Improved VLC container security by running the media daemon as a dedicated non-root user.
  * Updated service startup to set the runtime home directory and launch `cvlc` with user switching.
  * Preserved existing VLC runtime behavior, including HTTP/telnet settings, password handling, and audio output configuration.

### DIFF
--- a/vlc/Dockerfile
+++ b/vlc/Dockerfile
@@ -4,7 +4,6 @@ FROM ${BUILD_FROM}
 # Install VLC
 WORKDIR /usr/src
 
-# create vlc user to run vlc without patching vlc to run as root
 RUN \
     set -x \
     && apk add --no-cache \
@@ -12,7 +11,7 @@ RUN \
         nginx \
         pwgen \
         vlc \
-    && addgroup -S vlc && adduser -D -D -G vlc vlc \
+    && addgroup vlc && adduser -D -D -G vlc vlc \
     && sed -i '197s#.*#    dir = dir == undefined ? "file:///media" : dir;#' /usr/share/vlc/lua/http/js/controllers.js
         
 WORKDIR /
--- a/vlc/Dockerfile
+++ b/vlc/Dockerfile
@@ -10,7 +10,7 @@ RUN \
         nginx \
         pwgen \
         vlc \
-    && sed -i 's/geteuid/getppid/' /usr/bin/vlc \
+    && addgroup -S vlc && adduser -D -D -G vlc vlc # create vlc user to run vlc without patching vlc to run as root \
     && sed -i '197s#.*#    dir = dir == undefined ? "file:///media" : dir;#' /usr/share/vlc/lua/http/js/controllers.js
         
 WORKDIR /
--- a/vlc/Dockerfile
+++ b/vlc/Dockerfile
@@ -3,6 +3,8 @@ FROM ${BUILD_FROM}
 
 # Install VLC
 WORKDIR /usr/src
+
+# create vlc user to run vlc without patching vlc to run as root
 RUN \
     set -x \
     && apk add --no-cache \
@@ -10,7 +12,7 @@ RUN \
         nginx \
         pwgen \
         vlc \
-    && addgroup -S vlc && adduser -D -D -G vlc vlc # create vlc user to run vlc without patching vlc to run as root \
+    && addgroup -S vlc && adduser -D -D -G vlc vlc \
     && sed -i '197s#.*#    dir = dir == undefined ? "file:///media" : dir;#' /usr/share/vlc/lua/http/js/controllers.js
         
 WORKDIR /
--- a/vlc/Dockerfile
+++ b/vlc/Dockerfile
@@ -11,7 +11,7 @@ RUN \
         nginx \
         pwgen \
         vlc \
-    && addgroup vlc && adduser -D -D -G vlc vlc \
+    && addgroup vlc && adduser -D -G vlc vlc \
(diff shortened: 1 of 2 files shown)


vlc/Dockerfile
```diff
@@ -3,14 +3,16 @@ FROM ${BUILD_FROM}
 
 # Install VLC
 WORKDIR /usr/src
+
+# create vlc user to run vlc without patching vlc to run as root
```
agners: The location of the comment is a bit confusing. IMHO it is quite obvious what we are doing here, so I'd just drop the comment entirely.

```suggestion
```

vlc/rootfs/etc/s6-overlay/s6-rc.d/vlc/run
```diff
@@ -19,5 +19,5 @@ bashio::log.info "PulseAudio socket available, starting VLC"
 # Send out discovery information to Home Assistant
 /etc/s6-overlay/scripts/vlc-discovery &
 
-# Run daemon
-exec cvlc -I http --no-dbus --extraintf telnet --http-host 127.0.0.1 --http-password "INGRESS" --telnet-password "${PASSWORD}" --aout=pulse
+# Run daemon as vlc user
+exec su -c "cvlc -I http --no-dbus --extraintf telnet --http-host 127.0.0.1 --http-password \"INGRESS\" --telnet-password \"${PASSWORD}\" --aout=pulse" vlc
```
agners: Have you tested this? I am worried that the PulseAudio socket is not accessible by the user 🤔 

Also, there is a s6 built-in way to do this:

```
exec s6-setuidgid vlc cvlc ...
```

vlc/Dockerfile
```diff
@@ -3,14 +3,16 @@ FROM ${BUILD_FROM}
...
 RUN \
     set -x \
     && apk add --no-cache \
         inotify-tools \
         nginx \
         pwgen \
         vlc \
-    && sed -i 's/geteuid/getppid/' /usr/bin/vlc \
+    && addgroup -S vlc && adduser -D -D -G vlc vlc \
```
agners: This makes the group a system group but the user not. I'd go with regular group/user for both (drop -S from `addgroup`.

agners (changes requested):
Thank you for your PR! I agree, this would be a win from security perspective 👍 .

vlc/Dockerfile
YukonStriker: I have removed the comment

vlc/Dockerfile
YukonStriker: I have removed the -S and it still works.

vlc/rootfs/etc/s6-overlay/s6-rc.d/vlc/run
YukonStriker: Thanks you for your feed back. I have been testing this on my HA instance on a Pi4 running HAOS supervied with the add on code in the addons folder and using the UI to install my changes.

I have updated the run to use the s6-setuidgid as you suggested and after also adding an export for the home folder it works. With out adding the export, it was trying to use the root's home folder. I only had to export the home folder and not the user.

Error from log with not exporting the HOME variable

[14:32:28] INFO: PulseAudio socket available, starting VLC
s6-rc: info: service legacy-services successfully started
[0000007f8eba3d40] main libvlc error: cannot open config file (/root/.config/vlc/vlcrc): Permission denied
Failed to create secure directory (/root/.config/pulse): Permission denied
[0000007f8ea76a30] vlcpulse audio output error: PulseAudio server connection failure: Connection refused

Explanation:

Because s6-setuidgid retains root's environment profile (such as HOME=/root), programs that look for configuration or cache paths inside the home directory will throw a Permission denied error. Manually override these right before or alongside the execution:

export HOME=/home/myuser
export USER=myuser

exec s6-setuidgid myuser my-application

vlc/Dockerfile
```diff
@@ -3,14 +3,15 @@ FROM ${BUILD_FROM}
...
+
 RUN \
     set -x \
     && apk add --no-cache \
         inotify-tools \
         nginx \
         pwgen \
         vlc \
-    && sed -i 's/geteuid/getppid/' /usr/bin/vlc \
+    && addgroup vlc && adduser -D -D -G vlc vlc \
```
agners: Only one `-D` required here.
```suggestion
    && addgroup vlc && adduser -D -G vlc vlc \
```

agners (approved):
LGTM, thanks!

YukonStriker:
@agners Should the version in config.yaml have been updated?

agners:
> @agners Should the version in config.yaml have been updated?

Yes, I missed that :see_no_evil: But #4691 updates it and adds this change to the CHANGELOG.md. Best we can do at this point.

YukonStriker:
@agners I just tested out the new 1.1.0 version installing though the UI and all looks and sounds good.

PR #4685, opened 2026-06-28
Matter Server: add time_sync option (auto/on/off)

markvp:
## Breaking change

None. The new `time_sync` option defaults to `auto`, which enables time sync
only when the host clock is already NTP synchronised. On hosts that are not NTP
synchronised, behaviour is unchanged. On NTP-synchronised hosts the server now
pushes the correct time to commissioned Matter devices, which is the intended
improvement and can be turned off with `time_sync: off`.

## Proposed change

Adds a `time_sync` option (`auto` / `on` / `off`, default `auto`) to the Matter
Server add-on. It controls whether the server pushes the current time to
commissioned Matter devices via its `--enable-time-sync` flag:

- `auto` (default): enable only when the host clock is NTP synchronised, checked
  via the Supervisor API (`GET /supervisor/info`, field `ntp_synchronized`).
- `on`: always enable, logging an advisory warning if the host clock is not NTP
  synchronised (the time sent to devices is only as accurate as the host clock).
- `off`: never enable.

The bundled Matter Server base image is bumped to `1.2.5`, which includes the
`--enable-time-sync` flag added in matter-js/matterjs-server#440 (released in
matter-server `1.2.1`).

An earlier revision of this PR gated the flag behind a runtime server-version
check, because the Matter Server CLI uses commander.js (which exits on unknown
flags) and the bundled server predated the flag. Now that the flag ships in the
bundled server, that gate has been removed and `--enable-time-sync` is passed
directly whenever `time_sync` resolves to enabled.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New integration (thank you!)
- [x] New feature (which adds functionality to an existing integration)
- [ ] Deprecation (breaking change to happen in the future)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Companion server pull request: matter-js/matterjs-server#440 (merged; released
  in matter-server `1.2.1`)
- Link to documentation pull request:

## Checklist

- [x] I understand the code I am submitting and can explain how it works.
- [x] The code change is tested and works locally.
- [x] There is no commented out code in this PR.
- [x] The add-on version in `config.yaml` was bumped and `CHANGELOG.md` updated.
- [x] The new option is added to `config.yaml` `options`/`schema` and documented
      in `DOCS.md` and `translations/en.yaml`.
- [x] The startup script passes shellcheck with no new findings, and follows the
      existing bashio conventions used in this add-on.

## Summary by CodeRabbit

* **New Features**
  * Added a `time_sync` configuration option to control whether Matter Server pushes time to devices: `auto` (default), `on`, or `off`.
  * In `auto`, time sync is enabled only when the host clock is NTP-synchronized; `on` forces it and issues a warning if not synchronized.
* **Documentation**
  * Updated configuration docs, user-facing descriptions, and upgrade notes for the new option.
* **Chores**
  * Updated the Matter Server add-on to version 9.1.0 (including the underlying server image update).

### DIFF
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -119,6 +119,57 @@ if bashio::config.true "enable_test_net_dcl"; then
     extra_args+=('--ota-provider-dir' "/config/updates")
 fi
 
+# Resolve whether to enable time sync for nodes with the TimeSynchronization
+# cluster. The 'time_sync' option has three modes:
+#   - on:   always enable (warn if the host clock is not NTP synchronised)
+#   - off:  always leave disabled
+#   - auto: enable only when the host clock is NTP synchronised (default)
+time_sync_mode="$(bashio::string.lower "$(bashio::config 'time_sync' 'auto')")"
+ntp_synchronized="$(bashio::api.supervisor 'GET' '/supervisor/info' '' '.ntp_synchronized')"
+time_sync_enabled=false
+
+case "${time_sync_mode}" in
+  on)
+    time_sync_enabled=true
+    if [ "${ntp_synchronized}" != "true" ]; then
+        bashio::log.warning \
+          "Time sync is enabled but host NTP is not synchronised — time pushed to Matter devices may be inaccurate"
+    fi
+    ;;
+  off)
+    bashio::log.info "Time sync is disabled ('time_sync' set to off)."
+    ;;
+  *)
+    if [ "${ntp_synchronized}" = "true" ]; then
+        time_sync_enabled=true
+        bashio::log.info "Time sync 'auto': host NTP is synchronised, enabling time sync."
+    else
+        bashio::log.info "Time sync 'auto': host NTP is not synchronised, leaving time sync disabled."
+    fi
+    ;;
+esac
+
+if [ "${time_sync_enabled}" = "true" ]; then
+    # The --enable-time-sync flag was added to the Matter Server in 1.2.0 (see
+    # matter-js/matterjs-server#440). The server uses commander, which exits on
+    # unknown options, so only pass the flag when the running server is new
+    # enough; otherwise the option would break startup. A pre-release such as
+    # 1.2.0-alpha is treated as 1.2.0 by comparing the MAJOR.MINOR.PATCH core.
+    time_sync_min="1.2.0"
+    running_version=""
+    if [ -f /app/node_modules/matter-server/package.json ]; then
+        running_version="$(jq --raw-output '.version // empty' /app/node_modules/matter-server/package.json)"
+    fi
+    running_core="${running_version%%-*}"
+    if bashio::var.has_value "${running_core}" \
+        && [ "$(printf '%s\n%s\n' "${time_sync_min}" "${running_core}" | sort -V | head -n1)" = "${time_sync_min}" ]; then
+        extra_args+=('--enable-time-sync')
+    else
+        bashio::log.warning \
+          "Matter Server ${running_version:-(unknown)} does not support time sync (requires >= ${time_sync_min}); '--enable-time-sync' not passed. Update the Matter Server to enable this feature."
+    fi
+fi
+
 primary_interface="$(bashio::api.supervisor 'GET' '/network/info' '' 'first(.interfaces[] | select (.primary == true)) .interface')"
 
 # Try fallback method (e.g. in case NetworkManager is not available)
--- a/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
+++ b/matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
@@ -133,11 +133,11 @@ fi
 
 # Resolve whether to enable time sync for nodes with the TimeSynchronization
 # cluster. The 'time_sync' option has three modes:
-#   - on:   always enable (warn if the host clock is not NTP synchronised)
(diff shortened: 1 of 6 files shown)


markvp:
Thanks @coderabbitai — good catch on the unknown-flag risk. The Matter Server CLI uses commander.js without `allowUnknownOption()`, so it exits on unrecognised flags rather than ignoring them, and the `--enable-time-sync` flag itself is still pending in the server repo.

I've updated the startup script to gate the flag behind a minimum server version: it reads the running `matter-server` version from `/app/node_modules/matter-server/package.json` and only appends `--enable-time-sync` when the version is `>= 1.2.0` (a pre-release like `1.2.0-alpha` is treated as `1.2.0`). On older servers it logs a warning and skips the flag, so startup is unaffected. The host-NTP advisory warning is unchanged.

Note for context: the server-side flag is added in matter-js/matterjs-server#440, which still needs to be merged and released first.

@apollon77 — flagging this for visibility, since this add-on change depends on the `--enable-time-sync` flag (and `ENABLE_TIME_SYNC` env var) from matter-js/matterjs-server#440 landing in a server release. I've gated the add-on on Matter Server `>= 1.2.0`; if that PR ships under a different version, I'll adjust the threshold to match.

Apollon77:
@markvp Honestly I would not add a flag at all and only auto detect it. When it works it automatically works ... fine ... when not then it is logged and will not work ...

markvp:
> @markvp Honestly I would not add a flag at all and only auto detect it. When it works it automatically works ... fine ... when not then it is logged and will not work ...

Unless we want the time syncing but are not explicitly using NTP sync. The accuracy will be lower - hence the warning - but there's no reason why we shouldnt be able to still send the time to devices that consistently lose their time, is there?

perhaps we default to "true" and allow opt-out if desired rather than forcing it no matter what

Apollon77:
I like the idea for the HA addon to rely on te NTP sync status. so when synced then always do, when not then not. I would think about an opt-out if we get negative feedback from users and so also know reasons and such :-)

markvp:
well i can tell you that my NTP sync is broken running HA on a VM on proxmox VE. I still can't get it to sync and it always run about 4s off RTC.... but I'd rather it sync that time to devices than have them have no idea what the time is...

what if we default to on if NTP sync available, default to off if NTP sync not available, but allow explicit true/false that will override the NTP based default?

Apollon77:
Ok when it is like you then we need an App Flag "ignore unsynced NTP and enable regardless" ;-) which is then not doing that check but always enable it. And the default is still "enable when flag is set as synced" ?!

markvp:
Updated based on review feedback: the option is now a three-state `time_sync` choice (`auto` / `on` / `off`, default `auto`) instead of a plain boolean.

- `auto` (default): enable time sync only when the host clock is NTP synchronised (checked via `GET /supervisor/info`, `ntp_synchronized`).
- `on`: always enable, with a warning if the host clock is not NTP synchronised.
- `off`: leave disabled.

This is exposed as a `list(auto|on|off)` so all three states are selectable from the add-on Configuration dropdown (like `log_level`), rather than relying on an optional boolean whose "unset" state is only reachable via the YAML editor.

The version gate is unchanged: `--enable-time-sync` is only passed when the running Matter Server is `>= 1.2.0`, otherwise it is skipped with a warning so startup is unaffected. Still pending the server-side flag in matter-js/matterjs-server#440 (cc @apollon77).

matter_server/rootfs/etc/s6-overlay/s6-rc.d/matter-server/run
```diff
@@ -119,6 +119,57 @@ if bashio::config.true "enable_test_net_dcl"; then
...
+    if [ "${ntp_synchronized}" = "true" ]; then
+        time_sync_enabled=true
+        bashio::log.info "Time sync 'auto': host NTP is synchronised, enabling time sync."
+    else
+        bashio::log.info "Time sync 'auto': host NTP is not synchronised, leaving time sync disabled."
+    fi
+    ;;
+esac
+
+if [ "${time_sync_enabled}" = "true" ]; then
```
agners, line 164: I think I'd just wait until https://github.com/matter-js/matterjs-server/pull/440 is merged and a new release is out, then we can just drop this gate.

agners (changes requested):
Other than the version gate removal LGTM, nice work! If you remove the version gate, just keep the PR in draft until we have a matter.js server release, then add the version bump to this PR.

Apollon77:
@agners How the HA normally gets synced? In Discord we got this idea:
> For users naive on how to check or setup NTP on HA, you might suggest installing the HA Chrony App. Easy defaults to set/sync the HA system clock in the Config page gui of that App.

So yes we should think about guidance if it is not automatic

agners:
> @agners How the HA normally gets synced? In Discord we got this idea:
> 
> > For users naive on how to check or setup NTP on HA, you might suggest installing the HA Chrony App. Easy defaults to set/sync the HA system clock in the Config page gui of that App.
> 
> So yes we should think about guidance if it is not automatic

Home Assistant OS includes systemd-timesyncd and uses `time.cloudflare.com` (or if DHCP provides a NTP server, that one is preferred). Right now it is not possible to explicitly set a specific NTP server, but this is tracked with https://github.com/home-assistant/supervisor/issues/6278.

Apollon77:
@markvp FYI ... I took your changes into https://github.com/home-assistant/addons/pull/4727 and enhanced it with the other parts needed for the 9.1.0 release ...

PR #4690, opened 2026-06-29
Fix devcontainer workspace mount after app rename

felipecrs:
From https://github.com/home-assistant/devcontainer/pull/183#pullrequestreview-4595173690.

## Summary by CodeRabbit

* **Chores**
  * Updated the development workspace configuration to use the newer Home Assistant devcontainer image.
  * Adjusted the workspace folder path to target the correct apps directory while preserving the existing local folder structure and variables.


.devcontainer/devcontainer.json
agners: To force a container update we should also bump to `6-apps`.

agners (changes requested):

agners (approved):
LGTM, thanks!

PR #4699, opened 2026-07-02
nginx_proxy: Set nginx proxy to start in services phase

devn1x:
Hi,

I would like to propose changing the startup phase of the nginx_proxy to 'services' as in other similar projects (e.g. https://github.com/hassio-addons/addon-nginx-proxy-manager/blob/main/proxy-manager/config.yaml#L9). There was already an attempt at this (https://github.com/home-assistant/addons/pull/4164), but I want to try again with a cleaner pull request. Because in my mind, it makes more sense that way and makes the system as a whole feel more responsive when starting up.

I would greatly appreciate it if this could be merged, or let me know your thoughts and/or suggestions. Thanks for what you do!

## Summary by CodeRabbit

* **New Features**
  * Improved container startup behavior so required services are launched automatically on startup.

### DIFF
--- a/nginx_proxy/CHANGELOG.md
+++ b/nginx_proxy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.3.0
+
+- Change startup to `services` phase to start nginx before Home Assistant Core
+
 ## 4.2.0
 
 - update SSL/TLS configuration and add PQC key exchange following Mozilla guidelines
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 4.2.0
+version: 4.3.0
 hassio_api: true
 slug: nginx_proxy
 name: NGINX Home Assistant SSL proxy
--- a/nginx_proxy/config.yaml
+++ b/nginx_proxy/config.yaml
@@ -42,3 +42,4 @@ schema:
     servers: str
   real_ip_from:
     - str
+startup: services


agners:
It just feels wrong to start the proxy before the backend service is up, no?

I'd say that your installation might suffer from the same issue discussed in #4164, where a single integration just delays the whole system startup, preventing the system from entering `RUNNING` state. Fixing such slow integrations seems the better approach to me.

What probably would be ideal to have a startup setting which is tied to the http integration specifically being loaded :thinking:

agners:
Ok I've checked how Core not being running is handled, and NGINX just raises a 502 Bad Gateway, I guess not the worst outcome.

I think what I am more concerned is missconfiguration: If users set a port for the NGINX proxy which Core uses as its port, then you end up with a NGINX proxy running and occuping a port and no chance to reconfigure the NGINX proxy :cold_sweat:

agners:
This came up in discussion with @edenhaus. One additional concern here is that if we make the App load before Core, it is very easy to accidentally create port conflicts. E.g. with the change to default port 80 (https://github.com/home-assistant/core/pull/174278), and if the user simply mindlessly installs NGINX proxy with `startup: service`, he would end up with a 502 on next startup  :see_no_evil: (since app defaults to start on boot).

With the current startup order, this can't happen since Core claims the port first.

On the other hand, since with the 5 minute time which auto-reverts a new HTTP config (https://github.com/home-assistant/core/pull/174428), the NGINX proxy might not be started on time. Hm, actually, on the other hand, config changes only happen at runtime, not startup time, where we leave the NGINX proxy running. So I think this is a non-issue.

Anyhow, to conclude: I think making NGINX proxy `startup` to `services` seems to make it too easy to create port issues. A possible solution to make the proxy available earlier is to wire up a new startup signal from Core to Supervisor (http/core integration loaded), and introduce a new `startup` option (or make the existing `services` be tied to this, technically that would be a breaking change however).

PR #4700, opened 2026-07-02
Drop the note that main release 9.0.0 should be considered too

agners:
The update entity presenting the update automatically filters the CHANGELOG.md by heading according to version upgrade. So users who update from 8.x.x will see the relevant release notes already. No additional note required. The additional note is actually a bit confusing for users who already upgraded to 9.x.x.

## Summary by CodeRabbit

* **Documentation**
  * Cleaned up repeated upgrade warning entries in the release notes for several recent versions.
  * Removed redundant warning bullets in the 9.0.2–9.0.3 sections while keeping the version update entries intact, making the changelog easier to scan.

### DIFF
--- a/matter_server/CHANGELOG.md
+++ b/matter_server/CHANGELOG.md
@@ -7,18 +7,15 @@
 
 ## 9.0.3
 
-- **⚠️ When upgrading from 8.x, please also consider the release notes for 9.0.0.**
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
I agree those notes should be dropped. (nit: PR title could be adjusted to mention Matter)

agners:
That is how an upgrade straight from 8.5.0 gets rendered:

image

Apollon77 (approved):

PR #4702, opened 2026-07-02
Feature/nginx max body size - Add the ability to configure max body size from the UI to allow uploading files larger than 1MB

mediocaballero:
Add the ability to configure max body size from the UI to allow uploading files larger than 1MB

## Summary by CodeRabbit

* **New Features**
  * Added a UI option to configure the Nginx proxy maximum client request body size (in megabytes).
  * Allows larger uploads through the proxy (for example, KNX project files).
  * Uses a default limit of 1 MB when not configured.
* **Documentation**
  * Updated the changelog and on-screen configuration text to reflect the new upload-size setting.


nginx_proxy/config.yaml
```diff
@@ -20,6 +20,7 @@ options:
   keyfile: privkey.pem
   cloudflare: false
   use_ssl_backend: false
+  client_max_body_size: 1m
```
agners: This also need to be part of the schema below.

Defining it as a string and expect the user to know what suffix NGINX config is using is probably not very user friendly. So maybe better to make this `client_max_body_size_megabytes` and then append `m` automatically and define it as a `int` in the schema. 

The options should then also get an entry in `nginx_proxy/translations/en.yaml`, along with an appropriate description.

agners (changes requested):
Any change to an app needs to include a version bump in `config.yaml` and an entry in `CHANGELOG.md`.

nginx_proxy/config.yaml
mediocaballero: Sure, nice idea. Updating PR with this suggestion, also schema and translation. Thanks for the review!

mediocaballero: Oh, I also bumped the release, hope that's right (maybe such a small change could be a fix bump).

PR #4468, opened 2026-03-02
samba: 12.6.1

iamGavinJ:
- Added `mdnsd` for macOS discoverability
- Renamed `smb.conf` template for disambiguation with new templates
- Added config option, and `nmbd` startup logic, to disable NetBIOS
- Colocated mapped mounts to a `/smbshare` prefix to avoid potential confusion.
- Updated `config.yaml` `map:` to new object syntax
- Added config options to override default ports (445/139)

## Summary by CodeRabbit

* **New Features**
  * mDNS advertising for macOS Finder discoverability
  * Option to enable/disable NetBIOS and to override SMB (445) and NetBIOS (139) ports
  * Shares now colocate under /smbshare
  * Release bumped to 12.6.1

* **Refactor**
  * Configuration map converted to structured object syntax; share templating reworked

* **Documentation**
  * Docs updated and new translation entries for the NetBIOS option

* **Chores**
  * Startup/init and base image setup expanded to support new services

PR #4705, opened 2026-07-03
Samba: Add option to disable NetBIOS (nmbd)

atx32:
Added config option and associated files to allow user to disable `nmbd` service startup. NetBIOS is not needed to access SMB shares with current operating systems. Added parameter to `config.yaml` and implemented the enable/disable function by adding an `enable-check.sh` script and associated `Dockerfile` variable.

Description of netbios config parameter taken from https://github.com/home-assistant/addons/pull/4468 since this text had already been reviewed. Made other updates to documentation to include a full sample config and update listing order.

I did not bump the revision number or add to the changelog but can do that if needed.

## Summary by CodeRabbit

* **New Features**
  * Added a new NetBIOS over IP setting for the Samba add-on to enable or disable NetBIOS.
  * When NetBIOS is disabled, the NetBIOS service is no longer started and access is restricted to SMB on port 445 (with port 139 blocked).
  * Updated “local master” behavior so it depends on whether NetBIOS is enabled.
* **Documentation**
  * Refreshed the Samba share configuration guide with updated YAML examples and expanded descriptions for workgroup, NetBIOS, and local master.


samba/config.yaml
```diff
@@ -34,6 +33,8 @@ options:
     - ssl
   compatibility_mode: false
   apple_compatibility_mode: true
+  netbios: true
```
agners, line 36: Nice, yeah defaulting to true first makes sense. Maybe we can ship a breaking change where we swap the default to false in the future.

One thing which Claude pointed me at, the smb.conf needs apparently this as well to disable netbios entirely:

```
disable netbios = yes
smb ports = 445
```

I haven't fact checked it, but it seems legitimate. There is also `netbios name` set in the config, which we probably can get rid off if this setting is false 🤔 ?

samba/rootfs/etc/s6-overlay/scripts/enable-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/with-contenv bashio
```
agners: Other scripts use:

```suggestion
#!/command/with-contenv bashio
```

agners (changes requested):
Thank you for your contribution!

> I did not bump the revision number or add to the changelog but can do that if needed.

Yes, please also bump the version in `config.yaml` and add an entry to `CHANGELOG.md`.

samba/config.yaml
atx32: From my testing it appears that `disable netbios` has no impact and the `smb ports` setting is all that is needed. Adding this as a conditional configuration to `smb.gtpl`. For `netbios name` I'd rather leave it alone as it does no harm if always present.

samba/rootfs/etc/s6-overlay/scripts/enable-check.sh
atx32: All the other addon enable-check scripts use `/usr/bin` (openthread_border_router, deconz, etc.) If the preference is still to use `/command` I can change that.

atx32: I see that with s6-overlay v3 I should use what you suggested.

atx32:
Should all be good now. I would not consider changing the `netbios` default to false until mDNS advertising is implemented.

agners (approved):
LGTM, thanks!

agners:
> Should all be good now. I would not consider changing the `netbios` default to false until mDNS advertising is implemented.

There is also wsdd implemented now. Shouldn't that make the Samba server discoverable for almost all Windows systems already? :thinking:

atx32:
> There is also wsdd implemented now. Shouldn't that make the Samba server discoverable for almost all Windows systems already? 🤔

For Windows I agree. But in some other cases (VLC on FireOS as a personal example) shares aren't discoverable without nmbd running or the service advertised with mDNS.

PR #4717, opened 2026-07-11
Samba: Normalize stored enabled_shares values to lower case

kingpanther13:
## Summary

Step 1 of the plan agreed in home-assistant/supervisor#7023: prepare `enabled_shares` for a strict `list()` schema by normalizing stored values to lower case at startup. Values have always been validated case-insensitively and lower-cased at runtime before rendering smb.conf, so this changes no behavior — it only persists the canonical form.

On startup the app reads its stored options via the Supervisor API and, only when an `enabled_shares` value is not already lower case, writes them back lower-cased. On API failure it logs a warning and continues to boot, retrying on the next start.

Step 2 (separate, breaking): home-assistant/addons#4718 switches the schema from `match()` to `list()`, which also fixes the empty suggestion list in the configuration UI (home-assistant/frontend#51510).

## Summary by CodeRabbit

* **New Features**
  * Added automatic normalization of stored share names to lowercase during startup.
  * If updating stored configuration fails, a warning is logged and the update will be retried on the next startup.

* **Documentation**
  * Updated release notes with a new 12.8.1 entry describing the startup-time normalization behavior.

* **Chores**
  * Updated the Samba add-on version to 12.8.1.

### DIFF
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 12.7.2
+
+- Normalize stored `enabled_shares` values to lower case at startup. Values
+  were already handled case-insensitively at runtime; this persists the
+  canonical form in preparation for a stricter schema in a future release.
+
 ## 12.7.1
 
 - Enabled kernel oplocks in smb.conf to ensure changes made to files on disk are available immediately via SMBD. This covers all shares except backup, and media as the contents shouldn't be changed by the server once they're written in those shares.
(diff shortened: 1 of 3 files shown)


agners:
This writes back all options, including the newly introduced `netbios` in PR #4705. In this case this is a side effect we prefer, since we want `netbios` to be enabled in settings for existing installations. This will allow us to flip the default to `false`, later, disabling it by default for new users without affecting existing installations.

Just wanted to point this out for this PR. I'd like to test this PR again before merging, probably later today or tomorrow.

agners (approved):
LGTM!

PR #4727, opened 2026-07-15
Matter Server: bump to 1.2.6, add time_sync and default_fabric_label options (9.1.0)

Apollon77:
## Breaking change

None. The new `time_sync` option defaults to `auto`, which enables time sync only when the host clock is NTP synchronized — on hosts that are not NTP synchronized, behaviour is unchanged. The new `default_fabric_label` option is optional and inactive unless set.

## Proposed change

Takes over and finalizes #4685 (thanks @markvp for the groundwork!) and adds a few things on top:

- Bump the bundled Matter Server image from `1.1.7` to [`1.2.6`](https://github.com/matter-js/matterjs-server/blob/main/CHANGELOG.md), which brings Matter 1.6.0 support, time synchronization to devices, experimental ICD battery-saving support and many optimizations and fixes.
- Add a `time_sync` option (`auto` / `on` / `off`, default `auto`) controlling the server's `--enable-time-sync` flag (from #4685):
  - `auto` (default): enable only when the host clock is NTP synchronized (checked via the Supervisor `/host/info` `dt_synchronized` field).
  - `on`: always enable, logging a warning if the host clock is not NTP synchronized.
  - `off`: never enable.
- Add a `default_fabric_label` option (max 32 characters) wired to the server's `--default-fabric-label` flag. Use this if you use the Matter Server from multiple Home Assistant instances, which otherwise each keep pushing their own instance name as the label. When set, changing the label via the WebSocket API is blocked.
- Quote the final `extra_args` expansion in the startup script so option values containing spaces (e.g. a fabric label like "Living Room") are passed as a single argument.

## Type of change

- [ ] Dependency upgrade
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New integration (thank you!)
- [x] New feature (which adds functionality to an existing integration)
- [ ] Deprecation (breaking change to happen in the future)
- [ ] Breaking change (fix/feature causing existing functionality to break)
- [ ] Code quality improvements to existing code or addition of tests

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Supersedes #4685
- Link to documentation pull request:

## Checklist

- [x] I understand the code I am submitting and can explain how it works.
- [x] The code change is tested and works locally.
- [x] There is no commented out code in this PR.
- [x] The add-on version in `config.yaml` was bumped and `CHANGELOG.md` updated.
- [x] The new options are added to `config.yaml` `options`/`schema` and documented in `DOCS.md` and `translations/en.yaml`.
- [x] The startup script passes shellcheck with no new findings.

## Summary by CodeRabbit

- **New Features**
  - Added configurable device time synchronization with `auto`, `on`, and `off` modes.
  - Added an option to set a `default_fabric_label`, including protection against WebSocket label changes when configured.
  - Added support for Matter 1.6.0.
- **Documentation**
  - Updated configuration docs and the changelog for the 9.1.0 upgrade, including new options and upgrade guidance.
- **Improvements**
  - Updated the Matter Server base image to a newer version.
  - Improved startup argument handling for more reliable extra argument boundaries.

### DIFF
--- a/matter_server/DOCS.md
+++ b/matter_server/DOCS.md
@@ -54,7 +54,7 @@ App configuration:
 | beta                 | Install the latest `matter-server` from npm on startup instead of the bundled version. On failure a warning is logged and the bundled version is started. |
 | enable_test_net_dcl  | Enable test-net DCL for PAA root certificates, OTA updates and other device information.                |
 | time_sync            | Whether the Matter Server pushes the current time to Matter devices: `auto` (default; on only when the host clock is NTP synchronized), `on` (always; warns if NTP is not synchronized), or `off`. |
-| default_fabric_label | Pin the fabric label (shown on Matter devices, max 32 characters) to this value. Use this if you use the Matter Server from multiple Home Assistant instances, which otherwise each keep pushing their own instance name as the label. When set, changing the label via the WebSocket API is blocked. |
+| default_fabric_label | Pin the fabric label (shown on Matter devices, max 32 characters) to this value. By default the Home Assistant home name is used. This is useful if the Matter Server is used from multiple Home Assistant instances, which otherwise each keep pushing their own home name as the label. When set, changing the label via the WebSocket API is blocked. |
 | ble_proxy            | Expose the BLE proxy endpoint so the Home Assistant Matter integration can drive BLE commissioning through Home Assistant's bluetooth stack. Mutually exclusive with `bluetooth_adapter_id`. |
 | bluetooth_adapter_id | **Deprecated** — use `ble_proxy` instead. Set BlueZ Bluetooth Controller ID (for local commissioning). Still works for now. |
 | matter_server_args   | Extra command-line arguments passed to the Matter Server at startup (advanced).                         |
--- a/matter_server/DOCS.md
+++ b/matter_server/DOCS.md
@@ -53,6 +53,8 @@ App configuration:
 | log_level            | Logging level of the Matter Server component.                                                           |
 | beta                 | Install the latest `matter-server` from npm on startup instead of the bundled version. On failure a warning is logged and the bundled version is started. |
 | enable_test_net_dcl  | Enable test-net DCL for PAA root certificates, OTA updates and other device information.                |
+| time_sync            | Whether the Matter Server pushes the current time to Matter devices: `auto` (default; on only when the host clock is NTP synchronized), `on` (always; warns if NTP is not synchronized), or `off`. |
+| default_fabric_label | Pin the fabric label (shown on Matter devices, max 32 characters) to this value. Use this if you use the Matter Server from multiple Home Assistant instances, which otherwise each keep pushing their own instance name as the label. When set, changing the label via the WebSocket API is blocked. |
 | ble_proxy            | Expose the BLE proxy endpoint so the Home Assistant Matter integration can drive BLE commissioning through Home Assistant's bluetooth stack. Mutually exclusive with `bluetooth_adapter_id`. |
 | bluetooth_adapter_id | **Deprecated** — use `ble_proxy` instead. Set BlueZ Bluetooth Controller ID (for local commissioning). Still works for now. |
 | matter_server_args   | Extra command-line arguments passed to the Matter Server at startup (advanced).                         |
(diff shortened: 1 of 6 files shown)


matter_server/DOCS.md
```diff
@@ -53,6 +53,8 @@ App configuration:
 | log_level            | Logging level of the Matter Server component.                                                           |
 | beta                 | Install the latest `matter-server` from npm on startup instead of the bundled version. On failure a warning is logged and the bundled version is started. |
 | enable_test_net_dcl  | Enable test-net DCL for PAA root certificates, OTA updates and other device information.                |
+| time_sync            | Whether the Matter Server pushes the current time to Matter devices: `auto` (default; on only when the host clock is NTP synchronized), `on` (always; warns if NTP is not synchronized), or `off`. |
+| default_fabric_label | Pin the fabric label (shown on Matter devices, max 32 characters) to this value. Use this if you use the Matter Server from multiple Home Assistant instances, which otherwise each keep pushing their own instance name as the label. When set, changing the label via the WebSocket API is blocked. |
```
agners: Maybe worth mentioning what the default is?

```suggestion
| default_fabric_label | Pin the fabric label (shown on Matter devices, max 32 characters) to this value. By default the Home Assistant home name is used. This is useful if the Matter Server is used from multiple Home Assistant instances, which otherwise each keep pushing their own home name as the label. When set, changing the label via the WebSocket API is blocked. |
```

matter_server/translations/en.yaml
```diff
@@ -15,6 +15,22 @@ configuration:
...
+      clock is not NTP synchronized, since the time sent to devices is only as
+      accurate as the host clock); "off" disables it.
+  default_fabric_label:
+    name: Default Fabric Label
+    description: >-
+      Pin the fabric label (shown on Matter devices for this controller, max
+      32 characters) to this value. Use this if you use the Matter Server from
+      multiple Home Assistant instances, which otherwise each keep pushing
+      their own instance name as the label. When set, changing the label via
+      the WebSocket API is blocked.
```
agners:
```suggestion
      Pin the fabric label (shown on Matter devices for this controller, max
      32 characters) to this value. By default the Home Assistant home name is
      used. This is useful if the Matter Server is used from multiple Home
      Assistant instances, which otherwise each keep pushing their own home
      name as the label. When set, changing the label via the WebSocket API is
      blocked.
```

agners (approved):
Nice, LGTM. Just two minor nits.

matter_server/config.yaml
```diff
@@ -28,10 +28,13 @@ options:
   log_level: info
   beta: false
   enable_test_net_dcl: false
+  time_sync: auto
 schema:
   log_level: list(debug|info|notice|warn|error|fatal|verbose|warning|critical)
   beta: bool?
   enable_test_net_dcl: bool?
+  time_sync: list(auto|on|off)
+  default_fabric_label: str(0,32)?
```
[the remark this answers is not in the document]
agners: I guess if we address this, I'd go the other way and allow empty string and pass it. Not sure if empty string is supported throughout though? :thinking:

agners: Actually, just realized this is not really supported by Supervisor, despite docs claiming so 🙈 . I'll have to update the docs.

We have two options here: If you want to keep the 32-char cap (it's the Matter spec limit for fabric labels), use match with an anchored regex:

```suggestion
  default_fabric_label: match(^.{1,32}$)?
```

  {1,32} also disallows the empty string, which is what bashio::config.has_value already assumes. Simpler alternative if you don't care to enforce length in the schema:

```suggestion
  default_fabric_label: str?
```

  …and let the DOCS note + matter.js handle the limit. Not sure what happens if we pass longer than 32 chars? Maybe simpler alternative?

agners (approved):

matter_server/config.yaml
agners: > Actually, just realized this is not really supported by Supervisor, despite docs claiming so 🙈 . I'll have to update the docs.

FWIW, the docs seem to have the right idea here, it is just the implementation in Supervisor is not correct. https://github.com/home-assistant/supervisor/pull/7044 fixes it, so in the future we'll be able to use `str(1,32)?` here.

PR #4736, opened 2026-07-18
samba: migrate to app-based folder mappings and share names, bump base image

agners:
Rename the deprecated `addons` and `all_addon_configs` folder mappings to the new `local_apps` and `all_app_configs` names introduced in Home Assistant Supervisor (#6992), using the new default mount targets (/local_apps and /app_configs).

Rename the exposed shares to `local_apps` and `app_configs` to match Home Assistant's app terminology. Existing installations are migrated automatically at start by extending the enabled_shares normalization added in 12.8.1: stored `addons`/`addon_configs` values are rewritten to `local_apps`/`app_configs` via the Supervisor options API. Because the options merge strategy overrides lists with the user's stored value, changing the default alone would not migrate existing installs, and the schema must keep accepting the legacy names so the app can start to run the migration (and so auto-update schema validation passes).

Both the new and legacy share names stay exposed and point to the same directories, so existing SMB connections keep working while users move to the new names.

Also update the base image to 3.24-2026.06.1.

## Summary by CodeRabbit

* **New Features**
  * Added `local_apps` and `app_configs` SMB share names for local application content and configuration.
  * Kept legacy `addons`/`addon_configs` shares available for backward compatibility.
* **Bug Fixes**
  * Improved automatic migration of `enabled_shares`, including canonicalization, lowercasing, and deduplication.
  * Added clearer warnings when clients use deprecated share names.
* **Documentation**
  * Updated connection guidance and configuration examples to use `local_apps` and `app_configs`.
* **Chores**
  * Refined clean shutdown behavior and updated build base image/version.

### DIFF
--- a/samba/rootfs/usr/share/tempio/smb.gtpl
+++ b/samba/rootfs/usr/share/tempio/smb.gtpl
@@ -53,11 +53,11 @@
    delete veto files = {{ eq (len .veto_files) 0 | ternary "no" "yes" }}
 {{ end }}
 
-{{ if (has "addons" .enabled_shares) }}
-[addons]
+{{ if or (has "local_apps" .enabled_shares) (has "addons" .enabled_shares) }}
+[local_apps]
    browseable = yes
    writeable = yes
-   path = /addons
+   path = /local_apps
 
    valid users = {{ .username }}
    force user = root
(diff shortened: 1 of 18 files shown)


samba/rootfs/usr/share/tempio/smb.gtpl
```diff
@@ -53,11 +53,37 @@
...
+   delete veto files = {{ eq (len .veto_files) 0 | ternary "no" "yes" }}
+{{ end }}
+
+{{ if or (has "local_apps" .enabled_shares) (has "addons" .enabled_shares) }}
 [addons]
    browseable = yes
    writeable = yes
-   path = /addons
+   path = /local_apps
+
```
mdegat01, line 75:
```suggestion
   preexec = /usr/bin/logger -p local0.warning "smbd: %u connected to deprecated share %S from %m (%I), should switch to local_apps"

```
According to claude:

> preexec runs whenever a client connects to that share (tree-connect), and logger -p local0.warning writes it to syslog at warning severity. %u, %S, %m, %I expand to user, share name, client netbios name, and client IP.

Double sharing during the deprecation period makes sense but some logging to let them know they are using something deprecated and should switch seems welcome.

mdegat01 (changes requested):
This makes sense and looks good. I asked claude how to inject some logging into the config of a samba share and it provided the snippet below. Think we should do something like that to let them know they have a client using a deprecated share and should switch it to the new one.

mdegat01 (approved):
Looks good, I like the warning. Does the sigterm thing fix a known bug we can link in description/changelog? If so that'd be nice. If it was just something claude happened to discover during the other fix then guess its good as is.

agners:
> Does the sigterm thing fix a known bug we can link in description/changelog? If so that'd be nice. If it was just something claude happened to discover during the other fix then guess its good as is.

It is something I noticed during testing (since failed state is much more prominent shown in frontend nowadays).

agners:
I've removed the changes from `nmbd` and `wsdd` finish script since only `smbd` showed that exit code behavior. Verified also in actual source code, the way `smbd` works exit code 15 is expected.

```
s6-rc: info: service wsdd: stopping
s6-rc: info: service nmbd: stopping
terminate: Got SIGTERM: going down...
[18:28:06] INFO: Service nmbd exited with code 0 (by signal 0)
s6-rc: info: service nmbd successfully stopped
s6-rc: info: service smbd: stopping
[18:28:06] INFO: Service smbd exited with code 256 (by signal 15)
s6-rc: info: service smbd successfully stopped
[18:28:07] INFO: Service wsdd exited with code 0 (by signal 0)
s6-rc: info: service wsdd successfully stopped
```

PR #4738, opened 2026-07-19
Samba: Add configuration option to disable WSDD

### LINKED ISSUE #4722: [Samba] v12.7.0 enables WSDD with no way to disable it - Home Assistant now advertises itself on the network

App: Samba share
Version where it started: 12.7.0
Previous behavior: 12.6.x and earlier

What changed:
The 12.7.0 changelog entry "Add Windows automatic discovery support with WSDD" introduced a WSDD daemon that runs unconditionally whenever the app starts. After updating, Home Assistant suddenly appears as a device in the Windows Network view.

Expected behavior:
With compatibility_mode set to false, the app previously did not advertise the host on the network. The share was still fully reachable
via \hostname or \. I expect that turning off legacy/compatibility options keeps Home Assistant from broadcasting itself.

Actual behavior:
Home Assistant now broadcasts via WSDD regardless of configuration. There is no option in the app config to turn it off, the current schema has no wsdd key, and compatibility_mode: false has no effect on it. This is a new, unavoidable behavior introduced by an update, with no opt-out.

Impact:
Users who connect by direct \hostname/IP path don't want the host appearing in network discovery, and there's currently no way to revert to the pre-12.7.0 behavior short of downgrading and pinning the app.

Requested fix:
Add a config option to disable WSDD (e.g. wsdd: false), or gate the WSDD daemon behind compatibility_mode / an explicit "advertise on network" toggle, so the discovery behavior can be turned off.

Home Assistant OS

Samba share

12.7.x

Steps to reproduce:

    1. Run Samba share add-on 12.7.x with compatibility_mode: false.
    2. Open File Explorer > Network on a Windows machine on the same LAN.
    3. Home Assistant appears as a discoverable device, with no config option to prevent it.

### System Health information

N/A

```txt

```

```txt

```

### Additional information

_No response_

clowncracker:
Fixes #4722

   WSDD has run unconditionally since 12.7.0, so the host is advertised in the
   Windows Network view with no way to opt out. `compatibility_mode: false` and
   `netbios: false` have no effect on it, since `enable-check.sh` only gates
   `nmbd`.

   This adds a `wsdd` boolean option following the exact pattern of the
   `netbios` option added in 12.8.0: the flag is checked in `enable-check.sh`
   and the service file is removed from `user/contents.d` when disabled.

   Defaults to `true`, so existing installations see no change in behaviour.
   Disabling it does not affect share availability, only discovery.

   Includes the schema entry, DOCS.md documentation, the `en.yaml` UI strings,
   a version bump to 12.9.0, and a changelog entry.

## Summary by CodeRabbit

* **New Features**
  * Added a `network_discovery` configuration option to control WSDD-based advertising for Windows File Explorer.
  * When enabled (default), the host is discoverable on the local network; when disabled, it won’t be advertised.
  * Share access remains available by hostname or IP address regardless of discovery.

* **Documentation**
  * Updated documentation and example configuration to include `network_discovery`.
  * Added UI text and translation for the new setting.

### DIFF
--- a/samba/translations/en.yaml
+++ b/samba/translations/en.yaml
@@ -41,8 +41,8 @@ configuration:
     description: >-
       When NetBIOS is enabled, try and become a local master browser on a
       subnet.
-  wsdd:
-    name: Enable network discovery (WSDD)
+  network_discovery:
+    name: Enable network discovery
     description: >-
       Advertise the host with Web Services Dynamic Discovery so it appears
       automatically under Network in Windows File Explorer. Disable if you
--- a/samba/translations/en.yaml
+++ b/samba/translations/en.yaml
@@ -41,6 +41,13 @@ configuration:
     description: >-
       When NetBIOS is enabled, try and become a local master browser on a
       subnet.
+  wsdd:
+    name: Enable network discovery (WSDD)
+    description: >-
+      Advertise the host with Web Services Dynamic Discovery so it appears
+      automatically under Network in Windows File Explorer. Disable if you
+      connect by hostname or IP address and do not want the host to announce
+      itself on the network. Shares stay reachable either way.
   server_signing:
     name: Server signing
     description: >-
--- a/samba/CHANGELOG.md
+++ b/samba/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 12.9.0
+
+- Add configuration option to disable WSDD. When disabled, the wsdd service is
+  not started and the host is no longer advertised for automatic discovery.
+  Shares remain reachable by hostname or IP address.
+
 ## 12.8.1
 
 - Normalize stored `enabled_shares` values to lower case at startup. Values
--- a/samba/DOCS.md
+++ b/samba/DOCS.md
@@ -50,6 +50,7 @@ compatibility_mode: false
 apple_compatibility_mode: true
 netbios: true
 local_master: true
+wsdd: true
 server_signing: "default"
 veto_files:
   - ._*
@@ -118,6 +119,17 @@ Defaults to `true`.
 ### Option: `local_master`
 
 When NetBIOS is enabled, try and become a local master browser on a subnet.
+
+Defaults to `true`.
+
+### Option: `wsdd`
+
+Advertise the host on the network using Web Services Dynamic Discovery, so it
+appears automatically under Network in Windows File Explorer. Disable this if
+you connect to the shares by hostname or IP address and do not want the host
+to announce itself on the network.
+
+Disabling this has no effect on share availability; only on discovery.
 
 Defaults to `true`.
 
--- a/samba/DOCS.md
+++ b/samba/DOCS.md
@@ -50,7 +50,7 @@ compatibility_mode: false
 apple_compatibility_mode: true
 netbios: true
 local_master: true
-wsdd: true
+network_discovery: true
 server_signing: "default"
 veto_files:
   - ._*
@@ -122,7 +122,7 @@ When NetBIOS is enabled, try and become a local master browser on a subnet.
 
 Defaults to `true`.
 
-### Option: `wsdd`
+### Option: `network_discovery`
 
 Advertise the host on the network using Web Services Dynamic Discovery, so it
 appears automatically under Network in Windows File Explorer. Disable this if
--- a/samba/config.yaml
+++ b/samba/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 12.8.1
+version: 12.9.0
 slug: samba
 name: Samba share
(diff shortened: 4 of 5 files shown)


agners:
What is the use case here? Is it a problem when it's discoverable on the network?

The amount of options gets a bit out of hand. And I guess once we support mDNS (e.g. via #4742), yet another flag needs to be introduced :cry: .

clowncracker:
The primary use case is preventing the host from advertising itself on networks you don't fully control, where you'd rather it not broadcast itself to every device on the LAN. It's also restoring previous behavior rather than introducing a new feature. Before 12.7.0, the add-on didn't advertise itself this way, so servers that were intentionally hidden suddenly became discoverable with no way to disable it. That's the issue being addressed in #4722.

I also agree that adding a separate boolean for every discovery protocol doesn't scale, NetBIOS today, WSDD now, mDNS next. Would a single option, something like discoverable (defaulting to true), make more sense? It could control WSDD now and the upcoming mDNS work in #4742, rather than introducing a new flag for each protocol. NetBIOS is a bit of an exception since it already shipped as a standalone option in 12.8.0, but newer discovery mechanisms could share a common switch.

I'm happy to rework the PR in that direction if you think that's the better approach. Just let me know your preference.

atx32:
> The primary use case is preventing the host from advertising itself on networks you don't fully control, where you'd rather it not broadcast itself to every device on the LAN. It's also restoring previous behavior rather than introducing a new feature. Before 12.7.0, the add-on didn't advertise itself this way, so servers that were intentionally hidden suddenly became discoverable with no way to disable it. That's the issue being addressed in #4722.
> 
> I also agree that adding a separate boolean for every discovery protocol doesn't scale, NetBIOS today, WSDD now, mDNS next. Would a single option, something like discoverable (defaulting to true), make more sense? It could control WSDD now and the upcoming mDNS work in #4742, rather than introducing a new flag for each protocol. NetBIOS is a bit of an exception since it already shipped as a standalone option in 12.8.0, but newer discovery mechanisms could share a common switch.

I also didn't understand the use case for disabling discovery. The reason I added a flag for mDNS in https://github.com/home-assistant/addons/pull/4742 is I had taken note of https://github.com/home-assistant/addons/issues/4722 and decided to just do the work to give the user a choice. I think NetBIOS needs separate control as (like SMB1) it's now considered insecure and should not be used unless absolutely needed. As far as bundling WSDD and mDNS into a single option I'm neutral either way.

agners:
> I also didn't understand the use case for disabling discovery. The reason I added a flag for mDNS in #4742 is I had taken note of #4722 and decided to just do the work to give the user a choice. I think NetBIOS needs separate control as (like SMB1) it's now considered insecure and should not be used unless absolutely needed. As far as bundling WSDD and mDNS into a single option I'm neutral either way.

NetBIOS has an option since #4705. Ultimately we also want to flip the switch to default off, I intend to make a breaking change to the Samba app, I started to collect them in #4750.

samba/translations/en.yaml
```diff
@@ -41,6 +41,13 @@ configuration:
     description: >-
       When NetBIOS is enabled, try and become a local master browser on a
       subnet.
+  wsdd:
+    name: Enable network discovery (WSDD)
```
agners: Let's make this generic so we can bundle mDNS discovery later on.

```suggestion
  network_discovery:
    name: Enable network discovery
```

Mention only WSDD for now, the mDNS PR can then extend the description accordingly.

agners (changes requested):

samba/translations/en.yaml
agners: Technically NetBIOS does network discovery too, so we could argue the introduction of `netbios` as separate option in #4705 was wrong.

However, NetBIOS is a bit difference since it also offers name resolution. mDNS does that too, but it is not this app which offers it. Also it is legacy mechanism which we want to get rid off, so having the option allows a transition phase.

clowncracker:
> > I also didn't understand the use case for disabling discovery. The reason I added a flag for mDNS in #4742 is I had taken note of #4722 and decided to just do the work to give the user a choice. I think NetBIOS needs separate control as (like SMB1) it's now considered insecure and should not be used unless absolutely needed. As far as bundling WSDD and mDNS into a single option I'm neutral either way.
> 
> NetBIOS has an option since #4705. Ultimately we also want to flip the switch to default off, I intend to make a breaking change to the Samba app, I started to collect them in #4750.

I renamed to `network_discovery` across config, schema, the script, and the docs heading. Kept the description WSDD-specific for now so the mDNS PR can extend it, and left `netbios` as its own option as you suggested.

agners (approved):
LGTM!

PR #4742, opened 2026-07-22
Samba: Add service discovery using mDNS

atx32:
Added the capability to register smb service in mDNS for service discovery. This uses the `RegisterService` call on the `org.freedesktop.resolve1` service so the OS level mDNS announcer `systemd-resolved` is used.

- Use existing `network_discovery` configuration option for control of this feature.
- Add `glib` package which provides `gdbus` command needed.
- Add `host_dbus: true` to `config.yaml` to allow D-Bus access.
- In smbd `run` script, use `gdbus` command to register smb service on port 445 in mDNS.
- In smbd `finish` script, use `gdbus` command to unregister smb service in mDNS.
- Bump version to 12.11.0 and update documentation and changelog.

## Summary by CodeRabbit

* **New Features**
  * Added DNS-SD/mDNS service discovery alongside WS-Discovery for Samba on local networks.
  * Samba service entries are automatically unregistered from mDNS when stopping.

* **Documentation**
  * Updated configuration guidance and English translations to describe both discovery methods.

* **Chores**
  * Updated the release to version **12.11.0**.
  * Improved runtime support for network discovery.

### DIFF
--- a/samba/config.yaml
+++ b/samba/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 12.8.1
+version: 12.9.0
 slug: samba
 name: Samba share
 description: Expose Home Assistant folders with SMB/CIFS
@@ -9,6 +9,7 @@ arch:
   - amd64
 hassio_api: true
 host_network: true
+host_dbus: true
 image: homeassistant/{arch}-addon-samba
 init: false
 map:
@@ -33,6 +34,7 @@ options:
     - ssl
   compatibility_mode: false
   apple_compatibility_mode: true
+  mdns: true
   netbios: true
   local_master: true
   server_signing: "default"
@@ -57,6 +59,7 @@ schema:
     - "match(^(?i:(addons|addon_configs|backup|config|media|share|ssl))$)"
   compatibility_mode: bool
   apple_compatibility_mode: bool
+  mdns: bool
   netbios: bool
   local_master: bool
   server_signing: list(default|auto|mandatory|disabled)
--- a/samba/config.yaml
+++ b/samba/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 12.10.0
+version: 12.11.0
 slug: samba
 name: Samba share
 description: Expose Home Assistant folders with SMB/CIFS
--- a/samba/rootfs/etc/s6-overlay/s6-rc.d/smbd/finish
+++ b/samba/rootfs/etc/s6-overlay/s6-rc.d/smbd/finish
@@ -14,6 +14,16 @@ bashio::log.info \
   "Service ${service} exited with code ${exit_code_service}" \
   "(by signal ${exit_code_signal})"
 
+if bashio::config.true 'mdns'; then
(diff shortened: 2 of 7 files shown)


samba/rootfs/etc/s6-overlay/s6-rc.d/smbd/finish
```diff
@@ -14,6 +14,16 @@ bashio::log.info \
...
 
+if bashio::config.true 'mdns'; then
+bashio::log.info "Unregistering smb service in mDNS"
+app_hostname=$(hostname)
+gdbus call --system \
+  --dest=org.freedesktop.resolve1 \
+  --object-path=/org/freedesktop/resolve1 \
+  --method=org.freedesktop.resolve1.Manager.UnregisterService \
+  "/org/freedesktop/resolve1/dnssd/${app_hostname//-/_2d}"
+fi
```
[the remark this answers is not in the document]
atx32: @agners, on this analysis I don't agree that cleanup would be skipped because (as I understand it) configuration changes only take effect after a restart. So even if the setting is changed from true to false the unregister would still occur on restart.

The reason for not storing the object path is if there is an error due to the object path already existing (which aborts the app startup), the path isn't stored to be able to unregister it during shutdown and you are effectively stuck. I didn't see the need to try to solve for all these conditions when the path is deterministic to begin with. If you have different views I can adjust as needed.

Only question is will the app hostname only have dashes (`-`) for non-alphanumeric characters (e.g., `core-samba`) or do I need to account for others?

samba/config.yaml
```diff
@@ -9,6 +9,7 @@ arch:
   - amd64
 hassio_api: true
 host_network: true
+host_dbus: true
```
agners, line 12: Doing this directly from the app works, but it doesn't seem a good approach. It exposes much more attack surface to the app then necessary.

Ideally we extend the app configuration schema to allow service announcements, and let Supervisor handle the interaction with the host OS. We can take the arguments of [org.freedesktop.resolve1.Manager.RegisterService](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.resolve1.html) as inspiration for the config API. See also my comment in https://github.com/home-assistant/addons/pull/4468#discussion_r2871142249.

agners (commented):
Thanks for your PR! This goes into the right direction.

agners (changes requested):

samba/config.yaml
atx32: I have to admit what you are proposing is beyond my current capabilities. I would argue that both the `ssh` and `matter_server` apps have `host_dbus: true`, so I'm not sure why `samba` should be treated differently. If you're still not in favor of using D-Bus, I'll have to defer to someone more skilled to implement your preference.

atx32:
Other than the debate over `host_dbus`, this should be good to go now. I'm now using the `network_discovery` option to enable/disable this feature. Descriptions of the config option have been updated to refer to both DNS-SD/mDNS and WS-Discovery.

PR #4743, opened 2026-07-22
Upgrade piper to trixie

synesthesiam:
Use trixie Debian base for piper instead of bookworm (see: https://github.com/home-assistant/docker-base/pull/382)

## Summary by CodeRabbit

* **New Features**
  * Updated the Piper add-on to version 2.3.2.
  * Migrated the add-on’s base environment to Debian Trixie.
* **Bug Fixes**
  * Improved Python package isolation and installation reliability.
  * Ensured Piper starts in the correct working directory.
  * Improved startup consistency across supported architectures.

### DIFF
--- a/piper/Dockerfile
+++ b/piper/Dockerfile
@@ -8,18 +8,19 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /usr/src
 ARG BUILD_ARCH
 ARG WYOMING_PIPER_VERSION
-ENV PIP_BREAK_SYSTEM_PACKAGES=1
 RUN \
     apt-get update \
     && apt-get install -y --no-install-recommends \
         netcat-traditional \
         python3 \
         python3-pip \
+        python3-venv \
     \
-    && pip3 install --no-cache-dir -U \
+    && python3 -m venv .venv \
+    && .venv/bin/pip3 install --no-cache-dir -U \
         setuptools \
         wheel \
-    && pip3 install --no-cache-dir \
+    && .venv/bin/pip3 install --no-cache-dir \
         --extra-index-url https://www.piwheels.org/simple \
         --extra-index-url https://download.pytorch.org/whl/cpu \
         "wyoming-piper[zeroconf,zh] @ https://github.com/rhasspy/wyoming-piper/archive/refs/tags/v${WYOMING_PIPER_VERSION}.tar.gz" \
--- a/piper/Dockerfile
+++ b/piper/Dockerfile
@@ -27,6 +27,7 @@ RUN \
     \
     && rm -rf /var/lib/apt/lists/*
 
+ENV PATH="/usr/src/.venv/bin:$PATH"
 WORKDIR /
 COPY rootfs /
 
--- a/piper/config.yaml
+++ b/piper/config.yaml
@@ -1,5 +1,5 @@
 ---
-version: 2.3.1
+version: 2.3.2
 slug: piper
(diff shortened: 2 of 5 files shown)


piper/Dockerfile
```diff
@@ -8,18 +8,19 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
     apt-get update \
     && apt-get install -y --no-install-recommends \
         netcat-traditional \
         python3 \
         python3-pip \
+        python3-venv \
     \
-    && pip3 install --no-cache-dir -U \
+    && python3 -m venv .venv \
+    && .venv/bin/pip3 install --no-cache-dir -U \
```
agners, line 20: Instead of relying on cwd and this relative paths, I'd suggest to simply add it to the `PATH` variable, e.g.

```Dockerfile
ENV PATH="/usr/src/.venv/bin:$PATH"

RUN \
    apt-get update \
    && apt-get install -y --no-install-recommends ... python3-venv \
    && python3 -m venv /usr/src/.venv \
    && pip3 install --no-cache-dir -U setuptools wheel \
...
```

It also gets rid of `cd /usr/src || exit 1` and `exec .venv/bin/python3 -m wyoming_piper \`, simply leave the run script as it was before.

If we ever change to a different Python setup (e.g. `uv` managed), we can use the same approach.

agners (changes requested):

piper/config.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-version: 2.3.1
+version: 2.3.2
```
agners, line 2: Btw, I noticed you only bump the patch release. In other apps we usually bump at least minor. I guess mostly to keep Python library version in sync? If that is of higher priority to you, I am OK only bumping patch for these apps.

piper/Dockerfile
synesthesiam: That seems far less readable to me. The first `python3` is the system Python because `/usr/src/.venv/bin` doesn't exist yet, but then the very next `pip3` refers to the venv. And the run script now refers the venv indirectly, so you have to know what happened in the Dockerfile to understand what's actually running.

piper/config.yaml
synesthesiam: I wasn't sure what to do here since the app version historically has followed the version of the underlying Wyoming library (`wyoming-piper` here). I figured I'd get them in sync on the next version bump.

piper/Dockerfile
agners: Hm, yeah granted, in the particular `RUN` command that is a bit confusing. But I'd still prefer to not sprinkle deployment details all over the place. Then maybe let's just move `PATH` after the `RUN` statement, so the `RUN` statement is explicit? Then everyone else doesn't need to bother with what exact Python binary to use and just trust `PATH` to be properly setup.

synesthesiam: Updated to use PATH

piper/rootfs/etc/s6-overlay/s6-rc.d/piper/run
```diff
@@ -4,6 +4,8 @@
 # ==============================================================================
 # Start Piper service
 # ==============================================================================
+cd /usr/src || exit 1
+
```
agners: Actually, this too should no longer be necessary.

```suggestion
```

agners (changes requested):
LGTM!

agners (approved):
LGTM!

PR #4747, opened 2026-07-24
Remove git_pull directory and merge updates from master

wglemas-commits:


agners:
Why? git_pull is still a supported app.

wglemas-commits:
I'm trying to figure things out. So I wanted a clean install. But not to remove it from Home Assistant 😅

PR #4763, opened 2026-08-05
openthread_border_router: Bump stable to v2026.07.0 & beta to v2026.08.0

agners:
Bump OTBR stable to v2026.07.0 tag and beta to v2026.08.0 tag.

Drop patches which are already applied upstream.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved REST server socket recovery, reducing address-in-use issues after restarts.
  - Improved NAT64 handling for IPv4 packets with optional headers.
  - Added validation to reject malformed IPv4 packets and unsupported source-routing options.

- **Chores**
  - Updated the stable and beta OpenThread Border Router builds.
  - Advanced the add-on version to 3.1.0.

- **Documentation**
  - Added release notes for version 3.1.0.


balloob (approved):

puddly (approved):
No reported issues with this beta as far as I can tell so LGTM!

openthread_border_router/Dockerfile
TheJulianJES, line 69: nit: This seems to no longer be true since https://github.com/openthread/ot-br-posix/pull/3250 was merged (with v2026.06.0 and later, so both for new beta and new stable now).

OTBR is now the single source of truth for the name (regardless of if OTBR or OT publishes mDNS). Since the `OTBR_VENDOR_NAME` + `OTBR_PRODUCT_NAME` flags directly above it are [added together with a space](https://github.com/openthread/ot-br-posix/blob/86431d2d89c030f538ea192c47889f659a5bed26/src/border_agent/border_agent.cpp#L194) for the name, this basically results in the same (just with the `#define` having a now unnecessary trailing space). So that OpenThread-side `#define` is no longer necessary: https://github.com/home-assistant/addons/blob/95657e488570873dad94a3e62c2c84b43368fbe9/openthread_border_router/openthread-core-ha-config-posix.h#L74

(*Side note*: https://github.com/openthread/openthread/pull/12790 is the PR that changed the name to use the last 2 bytes of the address instead of the full one, and it now inserts ` #` before that suffix, which is what the `#define`'s trailing space used to be for. If that were still used now, we'd have a double space in the name.)

TheJulianJES (commented):
Should be fine, but flagging anyways: The advertised name changed between old stable and new stable (because of https://github.com/openthread/openthread/pull/12790):

Old stable on 3.0.2 (full EUI64):

Image 

Beta on 3.0.2 and new stable on 3.1.0 (last 2 bytes of EUI64):

Image

PR #4771, opened 2026-08-10
fix: avoid shutdown delay and exit code 137 stopping zwave_js

AlCalzone:
`S6_SERVICES_GRACETIME` controls how long s6 waits for supervised services to stop, while `S6_KILL_GRACETIME` is an **unconditional** delay before killing any remaining processes. Setting both to 30 seconds caused shutdown to exceed Supervisor’s 30-second timeout, resulting in unnecessarily long shutdown times and exit code 137. The fix uses `S6_SERVICES_GRACETIME=25000` and `S6_KILL_GRACETIME=3000`, keeping the worst-case shutdown within the Supervisor limit.

Fixes: https://github.com/home-assistant/core/issues/175860

## Bug Fixes

* Improved add-on shutdown responsiveness by reducing service grace periods.
* Resolved an issue that could cause the add-on to take 30 seconds to stop or exit with code 137.

## Release

* Updated the Z-Wave JS add-on to version 1.6.1.


agners (approved):
Nice catch, LGTM!

PR #4767, opened 2026-08-07
piper: Add it_IT-serena voices to voice list and fix docs URL

bacca87:
## What this PR changes

- Add the \it_IT-serena-high\ and \it_IT-serena-medium\ voices to the hardcoded \oice\ option list in \piper/config.yaml\ (after \it-riccardo_fasol-x-low\), so they can be selected as the add-on default voice.
- Fix the outdated download URL in \piper/DOCS.md\: the add-on (wyoming-piper >= 1.5.4) downloads voice models from the \main\ branch of \
hasspy/piper-voices\, not the old \1.0.0\ tag (which is also missing the new voices).

## Notes

- Both voices are already published on \
hasspy/piper-voices\ (main): \it/it_IT/serena/{medium,high}\ (PR rhasspy/piper-voices#95) and in the \
hasspy/piper-checkpoints\ catalog (PR #22).
- The voices are already usable via the Wyoming integration / Voice Assistant; this PR only makes them selectable as the add-on's default \oice\ option.

## Summary by CodeRabbit

* **New Features**
  * Added support for 12 new Piper voices across Italian, Bengali, Czech, Hebrew, Armenian, Japanese, Korean, Marathi, and Urdu.
  * Added Italian Serena high- and medium-quality voice options.

* **Documentation**
  * Updated the Piper voice-model download link to use the current repository source.
  * Documented the newly supported voices in the changelog.

PR #4791, opened 2026-08-24
piper: Disable ONNX Runtime telemetry

agners:
## What this PR changes

Sets `ORT_DISABLE_TELEMETRY=1` in the Piper add-on image and bumps the version to 2.3.4.

## Background

As reported in #4767, upgrading the add-on from 2.3.2 to 2.3.3 caused a flood of DNS queries for `mobile.events.data.microsoft.com` on installations that block that domain (e.g. via Pi-hole), stopping again on downgrade.

The 2.3.3 source change was data-only, but the version bump triggered an image rebuild with unpinned Python dependencies, which moved `onnxruntime` from 1.28.0 to 1.29.0 (among others). ONNX Runtime 1.29.0 introduced telemetry on Linux (previously Windows-only ETW): a new `PosixTelemetry` provider built on Microsoft's 1DS SDK that uploads usage events and a **persistent device identifier** to `https://mobile.events.data.microsoft.com/OneCollector/1.0` by default. This is verifiable by comparing the published images: the endpoint string is present in `libonnxruntime.so.1.29.0` in the 2.3.3 image and absent from the 2.3.2 image. The observed "flood" is the 1DS uploader retrying against the blocked domain.

## Fix

`ORT_DISABLE_TELEMETRY=1` is the opt-out documented in ONNX Runtime's [Privacy.md](https://github.com/microsoft/onnxruntime/blob/v1.29.0/docs/Privacy.md): set before ONNX Runtime initializes, it prevents the uploader, events, and persistent device identifier from being created for the process lifetime.

## Follow-ups to consider (not in this PR)

- Pin `onnxruntime` (and other dependencies) so rebuilds stop silently pulling in behavior changes.
- Audit other add-ons/images bundling `onnxruntime` — any of them will gain the same phone-home behavior once rebuilt against ≥ 1.29.0. Setting `ORT_DISABLE_TELEMETRY=1` in the base images may be the more durable answer.

## Summary by CodeRabbit

* **Privacy**
  * Disabled ONNX Runtime telemetry, preventing usage events and persistent device identifiers from being sent externally.
* **Documentation**
  * Added release notes documenting the telemetry change.
* **Chores**
  * Updated the Piper add-on version to 2.3.4.

### DIFF
--- a/piper/Dockerfile
+++ b/piper/Dockerfile
@@ -29,8 +29,6 @@ RUN \
 
 ENV PATH="/usr/src/.venv/bin:$PATH"
 
-# ONNX Runtime 1.29.0 introduced telemetry on Linux, uploading usage events and
-# a persistent device identifier to a Microsoft endpoint by default. Opt out.
 ENV ORT_DISABLE_TELEMETRY=1
 WORKDIR /
 COPY rootfs /
--- a/piper/Dockerfile
+++ b/piper/Dockerfile
@@ -28,6 +28,10 @@ RUN \
     && rm -rf /var/lib/apt/lists/*
 
 ENV PATH="/usr/src/.venv/bin:$PATH"
+
+# ONNX Runtime 1.29.0 introduced telemetry on Linux, uploading usage events and
+# a persistent device identifier to a Microsoft endpoint by default. Opt out.
+ENV ORT_DISABLE_TELEMETRY=1
 WORKDIR /
 COPY rootfs /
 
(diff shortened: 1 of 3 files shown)


piper/Dockerfile
```diff
@@ -28,6 +28,10 @@ RUN \
     && rm -rf /var/lib/apt/lists/*
 
 ENV PATH="/usr/src/.venv/bin:$PATH"
+
+# ONNX Runtime 1.29.0 introduced telemetry on Linux, uploading usage events and
+# a persistent device identifier to a Microsoft endpoint by default. Opt out.
```
balloob: No need to document all our thinking here. It's clear why we would not want that.
```suggestion
```

balloob (approved):

PR #4792, opened 2026-08-24
whisper: Pin inference stack and disable ONNX Runtime telemetry

agners:
## What this PR changes

- Pins the inference stack, updated to the latest releases: `onnxruntime==1.29.0`, `sherpa-onnx==1.13.6`, `faster-whisper==1.2.1`, `ctranslate2==4.8.1` (the latter two were already latest). These were previously unpinned transitive dependencies (via `onnx-asr[cpu]`, `wyoming-faster-whisper`), so every rebuild could silently change them.
- Sets `ORT_DISABLE_TELEMETRY=1`, the documented opt-out for ONNX Runtime telemetry.
- Bumps the app to 3.5.2 with a changelog entry.

## Background

Follow-up to #4791 (Piper). ONNX Runtime 1.29.0 introduced telemetry on Linux: a POSIX 1DS provider that uploads usage events and a persistent device identifier to `mobile.events.data.microsoft.com` by default.

**The currently released 3.5.1 image is *not* affected** — it ships onnxruntime 1.28.0, whose binaries contain no telemetry implementation or endpoint. However, because onnxruntime was unpinned, the next rebuild would have pulled in 1.29+ and silently started phoning home, exactly as happened to Piper in 2.3.3 (#4767). With `ORT_DISABLE_TELEMETRY=1` in place there is no reason to hold onnxruntime back, so this PR takes the upgrade to 1.29.0 deliberately, with telemetry disabled, and pins everything so future changes are explicit.

Notes:
- `transformers==4.52.4` and `onnx-asr==0.11.0` are exact pins required by upstream wyoming-faster-whisper 3.5.0, so they stay as-is.
- `sherpa-onnx` bundles its own statically-built ONNX Runtime copy; verified it contains no telemetry endpoint.

## Summary by CodeRabbit

- **New Features**
  - Released Whisper add-on version 3.5.2.
  - Updated speech-processing components to use compatible, configurable version settings.
  - Continued disabling ONNX Runtime telemetry by default.

- **Documentation**
  - Updated release notes to reference the latest inference-stack versions.
  - Simplified guidance about ONNX Runtime telemetry behavior and settings.

### DIFF
--- a/whisper/build.yaml
+++ b/whisper/build.yaml
@@ -8,6 +8,7 @@ args:
   # Must match the exact pins of wyoming-faster-whisper
   TRANSFORMERS_VERSION: 4.52.4
   ONNX_ASR_VERSION: 0.11.0
+  FUNASR_VERSION: 1.4.3
   ONNXRUNTIME_VERSION: 1.29.0
   FASTER_WHISPER_VERSION: 1.2.1
   CTRANSLATE2_VERSION: 4.8.1
--- a/whisper/build.yaml
+++ b/whisper/build.yaml
@@ -4,3 +4,12 @@ build_from:
   aarch64: ghcr.io/home-assistant/aarch64-base-debian:trixie
 args:
   WYOMING_WHISPER_VERSION: 3.5.0
+  WYOMING_VERSION: 1.9.0
+  # Must match the exact pins of wyoming-faster-whisper
+  TRANSFORMERS_VERSION: 4.52.4
+  ONNX_ASR_VERSION: 0.11.0
+  ONNXRUNTIME_VERSION: 1.29.0
+  FASTER_WHISPER_VERSION: 1.2.1
+  CTRANSLATE2_VERSION: 4.8.1
+  SHERPA_ONNX_VERSION: 1.13.6
+  TORCH_VERSION: 2.7.1
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -2,13 +2,12 @@
 
 ## 3.5.2
 
-- Pin the inference stack (`onnxruntime`, `faster-whisper`, `ctranslate2`,
-  `sherpa-onnx`) to the versions shipped in 3.5.1, so rebuilds no longer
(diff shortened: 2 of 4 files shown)


synesthesiam (approved):

whisper/build.yaml
```diff
@@ -4,3 +4,13 @@ build_from:
   aarch64: ghcr.io/home-assistant/aarch64-base-debian:trixie
 args:
   WYOMING_WHISPER_VERSION: 3.5.0
+  WYOMING_VERSION: 1.9.0
+  # Must match the exact pins of wyoming-faster-whisper
```
sairon, line 8: IMHO this would be a bit meh to maintain. Wouldn't it be easier to use `requirements.txt` for dependencies instead? Then you can just copy a part of the `pip freeze` output there instead of picking one dependency after another. Especially since the pins in [wyoming-faster-whisper `pyproject.yaml`](https://github.com/OHF-Voice/wyoming-faster-whisper/blob/main/pyproject.toml) are ranges and not exact versions.

Ideally, with a dedicated app repo, we could have this further simplified with Dependabot/Renovate.

agners: We pin dependency the same way for the Let's Encrypt app, I did not want to reinvent the wheel here :shrug: . With AI doing the bump, it doesn't seem to be a big deal to me :sweat_smile: .

> Ideally, with a dedicated app repo, we could have this further simplified with Dependabot/Renovate.

Agreed. Maybe revisit then? :thinking:

synesthesiam (approved):

whisper/build.yaml
sairon: But in case of Let's Encrypt it's not used to pin dependencies that would otherwise get installed at version only matching range of an installed package - in other words they don't substitute a lockfile like here.

Anyway, I don't have better idea than using `requirements.yaml` (or `uv` and its lockfile) so let's go with this. The `build.yaml` will eventually go away too.

sairon (approved):

PR #4802, opened 2026-08-25
Update hadolint to v2.15.1

agners:
## What this PR changes

- Bumps the CI hadolint from v1.17.2 (2019) to v2.15.1 (current).
- Extends `.hadolint.yaml` with the style-level rules the new version reports across existing Dockerfiles: `DL3025` (shell-form HEALTHCHECK/CMD — required for the piped `nc | grep` health probes), `DL3029` (`--platform` in FROM — intentional in silabs-multiprotocol's multi-stage build), `DL3059`/`DL3066` (info-level), `SC3010`.

## Why

v1.17.2 predates Docker Engine 25's `HEALTHCHECK --start-interval` flag and fails with a parse error (`invalid flag: --start-interval`); hadolint supports it since v2.14.0 (hadolint/hadolint#978). #4800 and #4801 use `--start-interval` to leave the "Starting" state faster and need this to pass lint.

Verified locally: all 26 Dockerfiles in the repo lint clean with v2.15.1 and the updated config, including the two PR branches using `--start-interval`.

## Summary by CodeRabbit

- **Chores**
  - Updated the Dockerfile linting tool to a newer version.
  - Adjusted linting rules to accommodate additional accepted Dockerfile patterns.


sairon (commented):
Seems that the official `hadolint/hadolint-action` supports glob patterns and the output looks good: https://github.com/hadolint/hadolint-action/issues/3#issuecomment-3699423066

Maybe we could do that instead of spinning up a Docker container for each Dockerfile? The adedd benefit is that Dependabot would keep hadolint up to date then.

agners:
Good idea — implemented in #4804, which uses `hadolint/hadolint-action` with `recursive: true` instead of the per-file Docker loop (and so gets Dependabot updates, real file paths, and inline annotations). This PR is superseded by it.

sairon:
Okay, let's close this in favor of #4804.

PR #4803, opened 2026-08-26
OpenThread Border Router: Fixing Headaches with FW and Routing

tannerdsilva:
## What this is for

I run this add-on on a live commercial deployment: a single-floor office around 35,000 sq ft with 450+ Thread devices, six border routers, and two switching racks on one shared L2. At that size, whatever breaks breaks loudly, and it keeps breaking until you fix it. Neither problem this PR fixes was something I read about. Both happened to me, at scale, and both are why the changes here exist.

### **Routing loop**

On a multi-BR network sharing one broadcast domain, a border router can install the broad `fc00::/7` ULA route. That swallows the whole ULA space. Devices end up stranded on stale prefixes, and the mesh starts routing in circles.

### **mDNS leak loop** 

Current OTBR releases leak the agent's mDNS advertisements out of both interfaces, `wpan0` and the backbone. On a shared fabric with a second border router, that wraps around. An advertisement leaves one BR's `wpan0`, crosses the mesh, shows up at the next BR, and comes back out its backbone. The network ends up forwarding its own discovery traffic in a loop.

## What changed

- Deterministic OMR prefix. `custom_omr_prefix` allows all border routers to operate in the same prefix, which removes the trigger for the catch-all route. Patches `0003` and `0004` make the ULA prefix reproducible and persisted.
- Catch-all route removed. Patch `0005` puts the broad `fc00::/7` route behind a compile-time flag. Patch `0006` tightens the Routing Manager to match exact `/64`s and publish the real OMR prefix instead of `::/0`.
- Firewall rebuilt. The nftables table `ip6 otbr` (seven chains) replaces the minimal `ip6tables` rules: default-deny in both forwarding directions, host input and output protection, TREL isolation, scope-gated multicast, ipset-to-nftables sync, and Docker's `DOCKER-USER` chain. The scope gate is what contains mDNS. Link-local multicast, where mDNS lives (`ff02::fb`), is dropped at the boundary in both directions, so an advertisement can't reach another border router and loop.
- Safety services. `otbr-route-guard`, `otbr-ipset-sync`, and `otbr-wpan-sysctl` cover the edge cases that only show up under load.
- Thread 1.4. DHCPv6 Prefix Delegation, Multi-AIL detection, peer-BR tracking.

## A note on the `firewall` setting

I want to be upfront about the `firewall` toggle in the app's config frontend. I didn't know what to do with it. Even with that enabled setting in place, the old behavior fell short of actually being enough, and it was hard to see what the firewall was even supposed to be blocking or managing. So this PR rebuilds that layer from the ground up rather than trying to patch the toggle. If anyone can point me at how the firewall rules should best be sliced out for the config UI, I'm glad to re-apply it properly on top of this.

## Tested where it lives

The network that exposed these bugs is the same one validating the fix. Both loops reproduced at that scale, and both are gone there. The image also builds cleanly from a fresh host. Validating against 450+ nodes across six border routers is a proof most OTBR deployments never get to run.


Apollon77:
@tannerdsilva For the fc00::/7 route issue, did you try simply disabling OPENTHREAD_POSIX_CONFIG_INSTALL_EXTERNAL_ROUTES_ENABLE ?

... see also https://github.com/openthread/openthread/pull/13562

Also @agners @puddly

Apollon77:
@tannerdsilva and for the firewall topic maybe https://github.com/openthread/ot-br-posix/pull/3325 is the best solution

tannerdsilva:
There's a lot in this. As I'm looking over the diff, I'm seeing things like the "suggested ULA prefix" upstream patch. Things like this are where I started (in speculation), not where I ended (in applied success). Should be removed.

I'll shake that out tonight, while also investigating the above concerns.

In my opinion, most painful part of this was the border router mDNS leak into thread interface. I have found this is super problematic because it causes "inverted L2 uplinks" and entropy on Thread in general. Unfortunately, it's entropy that can't be power cycled away at all - only time heals. My environment also required an elevated leader weight to resolve a complete offline deadlock after the mDNS loop was capped.

My fork started by addressing the fc00::/7 route issue because I knew I had undefined Thread issues, and this seemed like a valid angle to investigate. I then went "scorched earth" on all traffic after seeing no improvement in the routing patches alone. The fc00::/7 patches stayed in because it does seem like an incremental and valid technical improvement to an issue that I didn't discover (therefore couldn't feel entitled to omit it on my own).

tannerdsilva:
IMO: the consistent OMR prefix is important here. I can't say why, other than this:

1. My own trial and error evals make me believe that this enables much stronger meshing with multi-BR networks.

2. I have discovered that Apple border routers share OMR prefix across all BR in a network.

I think the easiest way to bring existing installations forward on this is to hash a private-range IPv6 OMR prefix from some public data field of the thread credentials (maybe hashing the network name). I will update this PR with this logic because I believe this is a "no downside, only upside" improvement for all network configurations of all sizes.

@Apollon77 as described, this static omr prefix with multiple BR is causing my visualization to break as we discussed in DM.

tannerdsilva:
Ok I think that concludes the development work for now. Working to validate these changes in my environment today...

agners:
Thanks for your PR! I still need to dive into the details, but one thing upfront: HAOS/Docker currently still uses the iptables backend. Some of the configs Docker applies the BR even relies on (I learned this the hard way with https://github.com/home-assistant/operating-system/pull/4589 and https://github.com/home-assistant/operating-system/pull/4678).

I am a bit worried about bad/negative nft/iptables interactions. I'd like to move to nft throughout (tracked with https://github.com/home-assistant/operating-system/issues/4588), but it needs to start on OS/Docker level).

Ideally we also submit changes to the upstream POSIX OTBR repo, so it gets a broader review. But I just realized that my last firewall changes in https://github.com/home-assistant/addons/pull/4500 also did not make it upstream yet :see_no_evil: .

PR #4804, opened 2026-08-26
Use the official hadolint action for linting

agners:
Supersedes #4802, implementing @sairon's suggestion there.

## What this PR changes

- Replaces the hand-rolled loop (which pipes each Dockerfile into a `hadolint/hadolint` container over stdin) with **`hadolint/hadolint-action`** in `recursive: true` mode, which lints `**/Dockerfile` in a single run.
- Drops the hand-maintained `HADOLINT_VERSION` env var — the hadolint version now comes bundled with the action.
- Extends `.hadolint.yaml` with the style-level rules the much newer bundled hadolint reports on existing Dockerfiles.

## Why

The pinned hadolint was **v1.17.2 (2019)**, which rejects `HEALTHCHECK --start-interval` (needed by #4800 / #4801) with a parse error. Beyond just bumping the version, using the official action brings three things the loop cannot:

- **Dependabot keeps it current.** The repo already watches `github-actions`, so the action — and therefore hadolint — gets update PRs automatically instead of the version silently aging six years.
- **Real file paths.** The stdin loop makes hadolint report `/dev/stdin:60:71`; the action reports `whisper/Dockerfile:60`.
- **Inline PR annotations** via the action's problem matcher.

The action is pinned by SHA with a version comment, matching the other actions in this workflow.

## Ignore-list additions

Bundled hadolint is v2.15.1 vs the previous v1.17.2, so it knows many more rules. All additions are style/info level:

| Rule | Why ignored |
|---|---|
| `DL3025` | Shell-form `HEALTHCHECK CMD` — required for the piped `nc \| grep` health probes |
| `DL3029` | `--platform` in `FROM` — intentional in silabs-multiprotocol's cross-build |
| `DL3059` | Consecutive `RUN` instructions (info) |
| `DL3066` | Non-numeric user id (info) |
| `SC3010` | `[[ ]]` in a script hadolint treats as POSIX sh |

## Testing

Ran the action's actual entrypoint and bundled image locally against this branch:

- Clean run: **exit 0**, no findings across all 24 Dockerfiles.
- Coverage check (with an emptied ignore list): findings reported for **24 of 24** Dockerfiles, confirming the recursive glob matches the same set as the old loop — not silently linting nothing.
- Failure check: a deliberately violating Dockerfile produced **exit 1** with findings identical to running hadolint directly, and with proper `path/Dockerfile:line` prefixes.
- `yamllint` clean on the modified workflow.

## Summary by CodeRabbit

* **Chores**
  * Updated automated Dockerfile linting to use a pinned Hadolint action with recursive scanning.
  * Added project-specific linting rules to reduce warnings for supported Dockerfile patterns.


sairon (approved):
Nice, LGTM!

Issue #4786, opened 2026-08-19
Tracking issue: migrate use of Debian Bookworm (LTS) to Debian Trixie (Stable)


eshattow:
From home-assistant/docker-base#382
@agners writes:
```
$ grep -r -e debian */build.yaml
assist_microphone/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
assist_microphone/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
deconz/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
deconz/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
letsencrypt/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:trixie-2025.12.2
letsencrypt/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:trixie-2025.12.2
mosquitto/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:trixie
mosquitto/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:trixie
openthread_border_router/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:trixie
openthread_border_router/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:trixie
openwakeword/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
openwakeword/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
piper/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
piper/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
silabs-multiprotocol/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bullseye
silabs-multiprotocol/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bullseye
speech_to_phrase/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
speech_to_phrase/build.yaml:  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
whisper/build.yaml:  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
[...]

PR #4808, opened 2026-08-30
Upgrade deCONZ to trixie

mattreim:
As discussed [here](https://github.com/home-assistant/addons/issues/4786), I created a test version using Trixie and have been testing it without issues for several days now, but a reinstallation of the app will likely be necessary.

## Summary by CodeRabbit

* **New Features**
  * Updated the deCONZ add-on to version 8.8.0.
  * Added support for the latest deCONZ Qt6 release on supported architectures.
  * Updated the underlying system environment to Debian Trixie.

* **Improvements**
  * Improved startup, shutdown, and environment handling.
  * Enhanced TigerVNC reliability and visibility of startup messages.

* **Bug Fixes**
  * Corrected log message timestamps during shutdown.


agners (changes requested):
Thanks for tackling this!

> but a reinstallation of the app will likely be necessary.

If we can we should really avoid this. Hm, I see, in a local test with this applied, the VNC server fails to start:

```
[12:08:50] INFO: Websockify waiting for VNC to start
[12:08:50] INFO: Starting VNC server (local - yes)...
[10:08:51] WARNING: Halt app
```

With the help of Claude I found the root cause:

Debian jumps tigervnc 1.12 (bookworm) → 1.15 (trixie). In 1.13+ the user dir moved from ~/.vnc to ~/.config/tigervnc, and Debian's wrapper auto-migrates it (/usr/share/perl5/TigerVNC/Config.pm:1425-1435):

```
  if (!(-d $opts->{'vncUserDir'})) {           # ~/.config/tigervnc missing
    if (!(-l $vncLegacyDir) && (-d $vncLegacyDir)) {  # ~/.vnc exists
      if (!mv($vncLegacyDir, $vncUserDir)) {
        print STDERR "$PROG: Could not migrate ...\n";
        exit 1;
```

mv can't create the parent. Your `/data` has `.vnc` but no `.config` at all.

Fresh `/data` works fine (exit 0) — which is precisely why mattreim's test install ran for days and why he assumed a reinstall was needed.

---

Add the following to the top of `services.d/deconz/run` should do the trick:

```
# tigervnc >= 1.13 migrates ~/.vnc to ~/.config/tigervnc, which needs ~/.config to exist
mkdir -p /data/.config
```

Btw, maybe also remove `&> /dev/null` from `tigervncserver`, that makes any future issues immeaditly visible.

Unrelated and pre-existing: The `run` scripts use `#!/usr/bin/with-contenv bashio` so they get TZ → local time (12:07 CEST). The `finish` scripts use plain `#!/usr/bin/env bashio` with no `with-contenv`, so no TZ → UTC (10:07). Same instant, different clock. Adding with-contenv to the finish scripts would make the logs consistent.

mattreim:
@agners Thank you very much for the detailed troubleshooting and the suggested solutions.

deconz/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.8.0
+
+- Migrate base image from bookworm to trixie (However, a reinstallation of the app is required)
```
agners: I'd say with this the reinstallation hint can go now:

```suggestion
- Upgrade base image from bookworm to trixie
- Fix timestamps in log messages during shutdown
```

agners (changes requested):
Nice, other then the nit, looks good to me!

agners (approved):
LGTM, thanks!

PR #4819, opened 2026-09-09
Upgrade to wyoming-faster-whisper 3.8.1

synesthesiam:
Upgrades to wyoming-faster-whisper 3.8.1, and stops shipping the PyTorch backends in the image.

**Smaller image.** `transformers` and `funasr` both need PyTorch — ~1.5 GB of a ~2 GB install — for backends most configurations never select. A `packs` s6 oneshot now installs them on startup from a uv cache in `/data`, the same approach as the Piper app, leaving the image about 75% smaller. Nothing to turn on: it asks the server's own `guess_stt_library()` what the configured `model`/`stt_library`/`language` resolve to. Nothing is downloaded for the default configuration, and a failed install falls back to faster-whisper with a warning rather than refusing to start.

**New options**

- `bias_names` — bias transcription toward your exposed entity, area and floor names
- `vad_endpointing` — seconds of silence after which the server ends the command itself
- `hf_token` — exported as `HF_TOKEN`, for gated or private models

**Also**

- Qwen3-ASR added as an `stt_library`/`custom_model_type` choice
- Caches (`HF_HOME`, `XDG_CACHE_HOME`, `MODELSCOPE_CACHE`) now live on `/data` instead of being thrown away on every app update, and are excluded from backups
- Health check does a Describe/Info round trip instead of grepping a socket
- Fix unset options reaching the server as the string `"null"` — a stock install was running with `--initial-prompt null`
- README synced with DOCS; every option is now documented

## Summary by CodeRabbit

* **New Features**
  * Added FunASR and Qwen3-ASR speech-to-text backends, including custom Qwen3-ASR models and bias-name prompting.
  * Added voice activity detection endpointing and Hugging Face token settings.
  * Added on-demand optional backend installation with offline cache support.

* **Improvements**
  * Improved automatic backend selection and fallback to faster-whisper.
  * Added offline model loading and expanded cache handling.
  * Updated Whisper components to newer versions.

* **Bug Fixes**
  * Improved transcription, model selection, health checks, and handling of options containing spaces.

### DIFF
--- a/whisper/CHANGELOG.md
+++ b/whisper/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## 3.7.0
+
+- Add `bias_names` option to bias transcription toward the names in your Home
+  Assistant: the names and aliases of your exposed entities, plus your area and
+  floor names, are added to the initial prompt, so "What's the temperature of
+  the incubi?" comes back as "What's the temperature of the Ecobee?"
+  - Off by default. Names are read over the Home Assistant API through the
+    Supervisor proxy, and refreshed in the background while you are still
+    speaking, so it costs no latency
+  - Only the prompt-capable backends (`faster-whisper`, `transformers`,
+    `qwen3-asr`) use the names; `sherpa`, `onnx-asr` and `funasr` ignore the
+    prompt
+- Add [Qwen3-ASR](https://huggingface.co/Qwen/Qwen3-ASR-0.6B) as an
+  `stt_library`/`custom_model_type` choice, defaulting to
+  `rhasspy/qwen3-asr-0.6b-onnx-int4-merged`
+  - Opt-in only: `auto` never selects it. The model is 785 MB and needs around
+    1.6 GB of RAM, and it is slower than the per-language defaults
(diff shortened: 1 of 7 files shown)


whisper/CHANGELOG.md
```diff
@@ -1,5 +1,73 @@
...
+  of every utterance went undecoded
+  - The startup warm-up was a no-op for the same reason, so the first
+    transcription after a restart also paid for it
+- Fix `initial_prompt` being sent as the literal word "null" when it is not
+  set, which is the default. Every transcription was biased toward that word,
+  it sat in front of the names `bias_names` adds, and it triggered the new
+  warning about prompts with a Distil-Whisper model
+- Fix selecting `model` = "custom" without a `custom_model` starting the app
+  with a model named "null" instead of reporting that the model is not set
+- Upgrade to wyoming-faster-whisper 3.8.1
```
agners: This is a huge changelog, I wonder if we can make it a bit denser and user focused.

Also using the YAML option strings is maybe not that user friendly anymore. Users nowadays see the option name from the translation file by default. In fact, in might be not straight foward to find the right setting when looking for the YAML option string in the app configuration nowadays.

PR #4826, opened 2026-09-16
ssh: update HA CLI to v5.5.0, rename addons/addon_configs to local_apps/app_configs

sairon:
* https://github.com/home-assistant/cli/releases/tag/5.4.0
* https://github.com/home-assistant/cli/releases/tag/5.5.0

## Summary by CodeRabbit

- **Updates**
  - Updated the Home Assistant CLI to version 5.5.0.
  - Released SSH add-on version 10.5.0.
  - Renamed shared directories from `addons` and `addon_configs` to `local_apps` and `app_configs`.
  - Updated available shared directories and home-folder links to use the new names.

- **Documentation**
  - Added changelog information for the 10.5.0 release.


agners (changes requested):
While at it, can you also update the mappings to the non-deprecated ones (addon -> app)

sairon:
> Rename the `addons` and `addon_configs` shares to `local_apps` and `app_configs` to match Home Assistant's app terminology.

Done as simple rename. If you think we should add symlinks for backward compatibility (or bump the major to signify breaking change)., let me know. I think it's not as important for this app as for Samba but I can accept a different view.

ssh/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 10.5.0
+
+- Update Home Assistant CLI to 5.5.0
+- Rename the `addons` and `addon_configs` shares to `local_apps` and `app_configs` to match Home Assistant's app terminology.
```
agners: Shares? I'd rather say bind mounts. Or just locations. Also, let's make this useful for users by naming the old and new path/locations here (the mappings are more an implementation/developer view).

agners (changes requested):
> Done as simple rename. If you think we should add symlinks for backward compatibility (or bump the major to signify breaking change)., let me know. I think it's not as important for this app as for Samba but I can accept a different view.

Yeah I share your view on that one. I guess the only downside could be inconsistency. But we do have the new location names as Samba shares as well IIRC. We still can add symlinks it in a follow-up if it turns out problematic.

One change request wrt `CHANGELOG.md`.

agners (approved):
Nice, LGTM!

